PR #1, opened 2016-08-30
Delete pivotcache when deleting a sheet

igitur:
PR originally logged at https://closedxml.codeplex.com/SourceControl/network/forks/igitur/ClosedXML/contribution/8516

When deleting a worksheet, the pivot cache parts have to be deleted as well, else Excel complains of a corrupt file.

There are 2 cases:
When a worksheet containing a pivot table is deleted, the pivot cache has to be deleted.
When a worksheet that is the source to a pivot table on another sheet is deleted, the pivot cache also has to be deleted.

I used a snippet found at http://blogs.msdn.com/b/vsod/archive/2010/02/05/how-to-delete-a-worksheet-from-excel-using-open-xml-sdk-2-0.aspx for the latter part.

### DIFF
--- a/ClosedXML/ClosedXML/ClosedXML/Excel/XLWorkbook_Save.cs
+++ b/ClosedXML/ClosedXML/ClosedXML/Excel/XLWorkbook_Save.cs
@@ -106,6 +106,86 @@ namespace ClosedXML.Excel
             }
         }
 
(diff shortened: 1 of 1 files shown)


Pyropace:
Looks good.  I will merge it 

Thanks

PR #2, opened 2016-08-30
Preserve tables' column names when table is of type queryTable.

igitur:
PR originally logged at https://closedxml.codeplex.com/SourceControl/network/forks/igitur/ClosedXML/contribution/8942

When opening and saving a file, tables' column names are replaced with Column1, Column2, etc. This PR fixes the issue and outputs the original column names.

### DIFF
--- a/ClosedXML/ClosedXML/ClosedXML/Excel/Ranges/XLRange.cs
+++ b/ClosedXML/ClosedXML/ClosedXML/Excel/Ranges/XLRange.cs
@@ -293,6 +293,11 @@ namespace ClosedXML.Excel
             return new XLTable(this, name, true, setAutofilter);
         }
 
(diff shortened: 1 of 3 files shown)


igitur:
Hang on with this PR. I want to add a test.

Pyropace:
Did you add unit test for this ?

igitur:
Not yet. I see there are conflicts here too.

igitur:
This issue is more convoluted than I thought. For one, the problem appears only when the table is of type `queryTable`. For normal tables, this isn't a problem. And query tables are far from fully implemented. So I think we park or delete this PR. It can't go into the next release as it stands.

Pyropace:
Lets park it

PR #3, opened 2016-08-30
Fix number format id bug

igitur:
PR originally logged at https://closedxml.codeplex.com/SourceControl/network/forks/igitur/ClosedXML/contribution/8514

Custom number formats should start at 165. 0 - 164 are reserved for built-in formats.
Also, it's not necessary to output the default numberformat (NumberFormatId = 0, FormatCode = "").


Pyropace:
Great work, 

Thanks for contribution

PR #5, opened 2016-09-01
Check returned value is null when no worksheet was found.

Yukimune:
from CodePlex https://closedxml.codeplex.com/SourceControl/network/forks/gibsoon/ClosedXML/contribution/8764


igitur:
So, do I understand correctly that this fixes the problem where there is a PrintArea defined, but that PrintArea points to a non-existing sheet?

Pyropace:
This looks good thank you

PR #6, opened 2016-09-04
Allow for custom themes in tables.  Fixes Codeplex 9499

vbjay:
Excel files with custom table theme fail to open because enum.parse can't find that value.


vbjay:
See https://closedxml.codeplex.com/workitem/9499

igitur:
I think a type safe enum pattern might be a better solution rather than changing to a string type.

vbjay:
I don't think you see the problem.  If I do a enum.tryparse solution, it will still resolve to an unknown predefined value.  Using a string allows for any value and I left the enum for the standard ones.

vbjay:
I realize we need to update it to allow for custom table theme creation and such.  This pull request just makes it so it won't throw an exception when loading a spreadsheet like it does now.

vbjay:
image

image

Pyropace:
Great do you have a sample test file, I would like to create a unit test for this before merging 

Thanks

vbjay:
Follow the codeplex link.

igitur:
Check out #7 , which implements a type-safe enumeration pattern and also opens the door for future development for having custom table themes.

Pyropace:
Hi Vbjay, 

Thank you again for the PR, we decide to go with igitur PR #7 . i will close this PR, looking forward to your future PRs

PR #7, opened 2016-09-06
Implement type safe XLTableTheme class 

igitur:
Implement type safe XLTableTheme class that won't choke on custom table theme name. Alternative solution to #6


igitur:
Saving workbooks with custom table themes still doesn't work. The code doesn't crash, but the custom table theme is not saved.

Pyropace:
Any idea why Saving workbooks dont work ?

igitur:
Do you mean the saving of the table theme that I mentioned? If so, that has just never been implemented. We should look at it in the future..

Pyropace:
Lets create a road map, if you can please add a unit test to this we can merge it 

Thanks

igitur:
Added a test. There is now a convenient place to test any file which ClosedXML should be able to load successfully, even though it doesn't implement all the parts at the moment. Refer to `LoadingTests.cs`.

Pyropace:
Merged thank you

PR #11, opened 2016-09-09
Fix workbook properties custom datetime property

igitur:
The custom date is correctly saved to the file, but not correctly interpreted as UCT when read from the file. 

To correctly interpret the date as UTC, the K identifier must be used in the DateTime format string.

See  http://stackoverflow.com/questions/6629575/another-datetime-parseexact-for-utc-datetime-string-question


igitur:
This PR fixes an existing test that failed on my machine due to timezone differences.

devel0:
I found issues storing DateTime object, for a simple repro:

```csharp
var dtnow = DateTime.Now; // localtime
var dtnow_utc = dtnow.ToUniversalTime(); // same dtnow but utc

ws.Cell(1, 1).SetValue(dtnow);
ws.Cell(2, 1).SetValue(dtnow_utc);
```

will results in two different date when open the workbook.

Everything should go fine if I convert my datetimes to localtime and I suggest to try apply .ToLocalTime() to the value (when of DateTime type) in the XLCell.SetValue(object value) function in order to ensure a coherent subsequent transformations.

But, I suppose [ToOADate](https://docs.microsoft.com/it-it/dotnet/api/system.datetime.tooadate?view=netframework-4.7.1#System_DateTime_ToOADate) function refer to the distance measured from a reference implied as a UTC because returned number itself doesn't contains any timezone reference. For this would be better todo a (myTime.ToUniversalTime()) before to process that with ToOADate().

PR #13, opened 2016-09-09
Add .editorconfig file

igitur:
See http://editorconfig.org for its use.


vbjay:
https://visualstudiogallery.msdn.microsoft.com/c8bccfe2-650c-4b42-bc5c-845e21f96328

igitur:
@Pyropace You OK with this one? EditorConfig is a great way to keep consistent code formats.

PR #15, opened 2016-09-10
fix "conditional format doesn't work when range has multiple conditio…

msntts:
I found conditional format doesn't work when a range has multiple conditional formats.

This bug seems to be caused by priority of conditional formats which are set to same value in each range.
When I checked XML data which was generated by Microsoft Excel, priority value was unique in a work sheet.

So I fixed to set unique value in each conditional formats.


msntts:
Sorry, I send wrong version.

PR #16, opened 2016-09-10
fix "conditional format doesn't work when range has multiple conditional format"

msntts:
I found conditional format doesn't work when a range has multiple conditional formats.

This bug seems to be caused by priority of conditional formats which are set to same value in each range.
When I checked XML data which was generated by Microsoft Excel, priority value was unique in a work sheet.

So I fixed to set unique value in each conditional formats.


Pyropace:
Thank you  for this PR, could you please create  a unit test for this PR as well, with sample file to test against ? Please look at Tests and Example Folders to see current unit tests.

igitur:
Merged into `develop` branch in 83679e8aa906bb00db0fb9464afa5c3c32243e1e

PR #24, opened 2016-09-14
Suppress warning CS1591

igitur:
Suppress warning CS1591 (Missing XML comment for publicly visible type or member) in order to reduce AppVeyor log output.


igitur:
Ah, appveyor output is much shorter (and faster) now!

Pyropace:
Good find :)

PR #25, opened 2016-09-14
Fix XLColor equals method

igitur:
Use ToArgb() to compare equality of colors, in order to return TRUE when comparing RGB(FF,FF,FF) to White. 
http://stackoverflow.com/questions/8146453/system-drawing-color-equality-operator-doesnt-treat-0xffffffff-as-color-white

Before you merge, I just want to check that the tests pass on AppVeyor too.


igitur:
Finally the test passes on my machine AND on AppVeyor. The issue was that the font used in the rich text string needs to be installed on the server too.  Would you run the tests on your local machine just to double check, please? If it passes, you can merge this.

Pyropace:
All passed, Thanks , this is a great fix

PR #28, opened 2016-09-15
Calculation properties

igitur:


igitur:
Implements these properties on the Workbook:
CalculationOnSave
ForceFullCalculation
FullCalculationOnLoad
FullPrecision

PR #34, opened 2016-09-16
Clearing AutoFilter shouldn't throw an exception

igitur:
Clearing AutoFilter shouldn't throw an exception, even if no AutoFilter is enabled. It should just silently continue.


skkanagaraj:
worksheet.AutoFilter.Enabled=false;
It throws "object reference not set to an instance of an object".

Issue #17, opened 2016-09-12
SetRichText(object value)


EdonGashi:
I cannot copy a RichText from an existing cell to another. Setting IXLCell.Value to a RichText copies only the first element of the rich text as a string. I suspect ```if (SetEnumerable(value)) return;``` is being called before SetRichText.

Can SetRichText(object value) be made public without negative side effects?

Pyropace:
SetRichText is called from XLCell,  setter  

If setter value is not able to set a Richtext, meaning if the content you are trying to set is not of type richtext, it will just call SetValue, It makes no sense to set SetRichText to public since you are required to set your values using setters and getters. 

If you have a sample document and code that you think should work but it doesn't , please post it here and we will investigate 

Thanks

EdonGashi:
Hello. below I have included a program that demonstrates my issue:

```cs
using ClosedXML.Excel;

class Program
{
    static void Main(string[] args)
    {
        using (var doc = new XLWorkbook())
        {
            var ws = doc.AddWorksheet("sheet");
            var sourceCell = ws.Cell(1, 1);
            sourceCell.Style.Fill.SetBackgroundColor(XLColor.Gray);
            sourceCell.RichText
                .AddText("Black")
                .SetBold()
                .AddText("Red")
                .SetFontColor(XLColor.Red)
                .AddText("Blue")
                .SetFontColor(XLColor.Blue);

            var cell2 = ws.Cell(3, 1);
            // Includes background color, which is not desired
            cell2.Value = sourceCell;

            var cell3 = ws.Cell(5, 1);
            // Enumerates the object
            cell3.Value = sourceCell.RichText;

            doc.SaveAs("richtext.xlsx");
        }
    }
}
```
![richtext](https://cloud.githubusercontent.com/assets/12145268/18448080/cfd6b348-7929-11e6-9446-14d40fbe7bc3.png)

I need to programmatically extract the formatted text from cell A1 into A9 without including any other cell style. Is this possible?

Pyropace:
I will take a look

PR #35, opened 2016-09-18
When setting cell value to RichText, don't use SetEnumerable.

igitur:
Solves issue #17 .


igitur:
This is ready for merge.

Issue #40, opened 2016-09-20
Problem with R1C1 Regex


Anon4819:
Ran into this while making something for work, leaving a note here for anyone who cares:

ClosedXML will mangle formulas in which a cell reference is immediately followed by a minus sign and then a number, when said formulas have been generated by a copy paste operation in Excel.

For example, in cell B1, type "10", then in B2, "=B1-1", then drag B2 down so that you get a series of cells counting down from 10. Save the book, open it in ClosedXML, and try reading the copy pasted cells' FormulaA1 string. The strings will show up as references to the A column and one row above. If you save the book, they will be saved this way as well, thus breaking any such formulas in the book.

I haven't investigated this thoroughly, but my hunch is that Excel probably saves copy pasted formulas internally in R1C1 format. Thus the formula in B2, "=B1-1", becomes "=R[-1]C-1". This should be parsed as "=(R[-1]C) - 1", however there is a bug in the regex used to parse R1C1 formulas which causes the "-1" after the cell reference to be parsed as if it were part of the reference itself, shifting the reference one column to the left and losing the minus 1 in the formula.

The regex in question is as follows, from the XLCell.cs file:

`(?<=\W)([Rr]\[?-?\d{0,7}\]?[Cc]\[?-?\d{0,7}\]?)(?=\W)`

The question marks after the brackets here allow the regex to eat a negative number even when there are no brackets around it, which is the cause of the bug. I have changed the regex to the following in my local copy:

`(?<=\W)([Rr](?:\[-?\d{0,7}\]|\d{0,7})?[Cc](?:\[-?\d{0,7}\]|\d{0,7})?)(?=\W)`

This way it either matches a pair of brackets with either a negative or positive number inside, or just a positive number (for absolute references), but not just a negative number.

The above fix seems to be working for me and has solved my immediate problem.

Hopefully this is enough information for someone with more time/dedication than I to make sure the above fix won't break any existing features and commit it.

Regards.

Pyropace:
HI @Anon4819 , 

Thanks for the report, could you create a PR for this, if you have a branch that is working for you? i will investigate this further tonight

Anon4819:
Alright I think I did that right:
https://github.com/ClosedXML/ClosedXML/pull/41

PR #41, opened 2016-09-21
Fixed R1C1 regex bug

Anon4819:
This should fix the regex bug described in https://github.com/ClosedXML/ClosedXML/issues/40


igitur:
Could you add some unit tests, please?

Pyropace:
Looks great thanks. Please add unit test as per @igitur comment

PR #46, opened 2016-09-24
Fix paths for Unix environments

igitur:
Ensure that files are read and written correctly on Unix systems where paths have / path separator.


igitur:
This one is still failing on Mono, so keep open please.

PR #47, opened 2016-09-24
Parse and convert to string using invariant culture

igitur:
I'm running en-ZA culture on my Linux box at home and currently this causes parsing problems with numbers. Numbers have to be stored and parsed according to the invariant culture.

Don't merge this PR yet. I just want to see whether it passes the tests on AppVeyor (which is a Windows machine).


Pyropace:
it seems to be passing now, are you good with a merge ?

igitur:
Yup, you can merge.

PR #62, opened 2016-10-10
Optionally validate the workbook using OpenXmlValidator when saving it.

igitur:
Optionally validate the workbook using OpenXmlValidator when saving it. When in DEBUG configuration, validation is set to TRUE.

This does make one test fail, which means we're outputting incorrect OpenXML for that test. Will follow up on that.


igitur:
I notice the test (`StyleFont`) passes on AppVeyor, which means the tests are running under the Release configuration.

igitur:
Ok, now the `StyleFont` test fails, as expected. Will look at it in a separate PR.

igitur:
Ok, this now passes.

PR #85, opened 2016-10-31
Fix the pivot table support for multiple pivot tables.

Petelids:
Adding multiple pivot tables to the same workbook was failing due to a number of minor bugs (as per this Stackoverflow question - http://stackoverflow.com/questions/31432347/closedxml-creating-multiple-pivot-tables.

### DIFF
--- a/ClosedXML/Excel/XLWorkbook_Save.cs
+++ b/ClosedXML/Excel/XLWorkbook_Save.cs
@@ -1775,31 +1775,31 @@ namespace ClosedXML.Excel
             tableDefinitionPart.Table = table;
         }
 
-        private static void GeneratePivotTables(WorkbookPart workbookPart, WorksheetPart worksheetPart,
-            XLWorksheet xlWorksheet,
-            SaveContext context)
-        {
-            foreach (var pt in xlWorksheet.PivotTables)
-            {
-                var ptCdp = context.RelIdGenerator.GetNext(RelType.Workbook);
+		private static void GeneratePivotTables(WorkbookPart workbookPart, WorksheetPart worksheetPart,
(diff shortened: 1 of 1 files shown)


igitur:
Thanks for the contribution. Please add some tests.

igitur:
I'm using the code you link to in the StackOverflow issue:
```c#
static void CreateTestPivotTables(string filePath)
{
    var wb = new XLWorkbook();

    var wsData = wb.Worksheets.Add("Data");            
    wsData.Cell("A1").Value = "Category";
    wsData.Cell("A2").Value = "A";
    wsData.Cell("A3").Value = "B";
    wsData.Cell("A4").Value = "B";
    wsData.Cell("B1").Value = "Number";
    wsData.Cell("B2").Value = 100;
    wsData.Cell("B3").Value = 150;
    wsData.Cell("B4").Value = 75;
    var source = wsData.Range("A1:B4");

    for (int i = 1; i <= 2; i++)
    {
        var name = "PT" + i;
        var wsPT = wb.Worksheets.Add(name);
        var pt = wsPT.PivotTables.AddNew(name, wsPT.Cell("A1"), source);
        pt.RowLabels.Add("Category");
        pt.Values.Add("Number")
            .ShowAsPctFrom("Category").And("A")
            .NumberFormat.Format = "0%";
    }

    wb.SaveAs(filePath);
}
```
But even with your patch, the resulting file is still corrupt.

igitur:
Fixed in #87 .

Petelids:
Apologies @igitur, my patch was nonsense. I'd only tested it on multiple pivot tables on the same worksheet. I just created a new patch now but noticed you've already done it in #87. I'll take more care over the next patch...

PR #87, opened 2016-11-01
Support for multiple pivot tables

igitur:
Create correct pivot caches with correct cache id for each pivot table. 
TODO: Avoid duplicate pivot caches if separate pivot tables share same data range.
Supercedes #85 .


akantro:
@igitur - just wanted to confirm if this fix is going to be included in any upcoming release build?  I see the latest release was the first release in a while so just wanted to see if its going to be a long time for another patch to get released?

igitur:
It won't take as long again.

akantro:
Thanks for your reply.  is it recommended to take a cut of the latest version of source or do you advise against that.  The reason i ask is that this fix is critical for a project I am working on so want to figure out the quickest way i can start using this fix.

igitur:
I leave that decision to you. You're welcome to clone a branch and compile
it yourself. It's quite easy.

akantro:
which branch has this fix?  I tried downloading 'Develop' and the fix doesn't seem to be there

PR #99, opened 2016-11-11
Pivot table improvements

### LINKED ISSUE #96: More than one Value to Pivot table not working
base on your example with pastries if I would have another column like price and I would like to have in my pivot table two values like : 

pt.Values.Add("NumberOfOrders");
pt.Values.Add("Price");

if you add two Values to that pivot table it crashes.

igitur:
Lots of adjustments and improvements to pivot tables. Needs to be checked thoroughly.

Should fix #96


igitur:
Also contains a small breaking change in the API. I renamed methods and variables containing to Pct to Percentage.

Pyropace:
@igitur this has conflicts has well, we can merge as soon as they are resolved :)

igitur:
Conflicts resolved.

PR #108, opened 2016-11-23
Fix regex for parsing cell formula.

### LINKED ISSUE #40: Problem with R1C1 Regex

### LINKED ISSUE #107: Copying cell with certain formula results in corrupt cell
Hi 

I found an issue with while trying to copy a cell with a valid FormulaA1 to a different cell. Then copying works just fine but when I try to access formula in the copied cell I get an System.FormatException.
This exception is thrown on the FormulaA1, HasFormula and the Value property of the cell. Intrestingly the FormualR1C1 returns a value.

My guess there might be an issue in manipulating the formula on copy.

Here my code for reproduction:
```c#
FileInfo file = new FileInfo("Testfile.xlsx");
using (FileStream stream = file.OpenRead())
{
    using (XLWorkbook workbook = new XLWorkbook(stream))
    {
        IXLWorksheet workSheet = workbook.Worksheet("Sheet1");

        workSheet.Cell("B44").Value = workSheet.Cell("A44");

        Debug.WriteLine(workSheet.Cell("B44").HasFormula);
    }
}
```

The Excelfile is an empty file with this formula in the cell A44:
`=IF(A9="Tender";0;(A72-(A62+A63+A65+A67+A69))*A17)`

Thanks for your help in advance.

igitur:
Fix regex for parsing cell formula. Fixes #40. Fixes #107. Closes #41.


Pyropace:
@igitur there seems to be a conflict, can you check please

igitur:
Conflict fixed.

Issue #205, opened 2017-02-15
IXLColumn.Delete() sometimes creates corrupt XLSX files.


albinsunnanbo:
Reproducing example: https://github.com/albinsunnanbo/ClosedXmlColumnBug
To reproduce, just run the command line app and look in bin\debug\DoWork.xlsx and bin\debug\DoesntWork.xlsx

I create a table with `IXLCell.InsertTable` from a `List<T>`: https://github.com/albinsunnanbo/ClosedXmlColumnBug/blob/master/TestExcelExport/ExcelExporter.cs#L44

Then I remove a column from the worksheet with `IXLColumn.Delete`: https://github.com/albinsunnanbo/ClosedXmlColumnBug/blob/master/TestExcelExport/ExcelExtensions.cs#L38

If my `T` is DoWorkModel

    public class DoWorkModel
    {
        [Display(Name = "First string")]
        public string String1 { get; set; }

        [Display(Name = "Second string")]
        public string String2 { get; set; }

        [Display(Name = "The Int")]
        public int? Int1 { get; set; }

    }

And I remove the third column (Int1) a valid XLXS document is created.

On the other hand, if my `T` is DoesntWorkModel

    public class DoesntWorkModel
    {
        [Display(Name = "First string")]
        public string String1 { get; set; }

        [Display(Name = "The Int")]
        public int? Int1 { get; set; }

        [Display(Name = "Second string")]
        public string String2 { get; set; }
    }

And I remove the second column (Int1) a corrupt XLXS document is created.
When trying to open the document in Excel the following dialog is shown:
![image](https://cloud.githubusercontent.com/assets/5158396/22974700/dfbfbb68-f383-11e6-8b81-5b355204437c.png)
and then
![image](https://cloud.githubusercontent.com/assets/5158396/22974710/eb02eea0-f383-11e6-9d9c-ad3775fba39f.png)

and the resulting log file

	<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
	<recoveryLog xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
		<logFileName>error324960_01.xml</logFileName>
		<summary>Errors were detected in file 'C:\Git\ClosedXmlColumnBug\TestExcelExport\bin\Debug\DoesntWork.xlsx'</summary>
		<repairedRecords summary="Following is a list of repairs:">
			<repairedRecord>Repaired Records: Table from /xl/tables/table.xml part (Table)</repairedRecord>
[...]

PR #208, opened 2017-02-16
Add attribute to ignore column

igitur:
Related to issue #205

### DIFF
--- a/ClosedXML/Attributes/IgnoreAttribute.cs
+++ b/ClosedXML/Attributes/IgnoreAttribute.cs
@@ -1,0 +1,19 @@
+﻿using System;
+
+namespace ClosedXML.Attributes
+{
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property, AllowMultiple = false, Inherited = false)]
+    public class IgnoreAttribute : Attribute
+    {
+        public IgnoreAttribute()
+            : this(true)
+        { }
+
+        public IgnoreAttribute(bool ignore)
+        {
+            this.Ignore = ignore;
+        }
+
+        public bool Ignore { get; private set; }
+    }
+}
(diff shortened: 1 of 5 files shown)


albinsunnanbo:
While we're at it adding attributes a Name Attribute that takes precedence over `[Display(Name = "Column name")]` would be useful.
If often use the same model for displaying lists in ASP.NET as for exporting to Excel, but sometimes I want to tweak the Excel export a bit. Hiding columns is one tweak, changing the name (e.g. to an abbreviation) is another. If I recall correctly there is a column order attribute

There are lest two possible API styles for this.
One is different Attributes:

    [Ignore]
    [Order = 2]
    [Name = "FooBar"]
    public string String1 { get; set; }

The other is to have an attribute with named parameters, like

    [XLColumn(Ignore = false, Order = 2, Name = "FooBar")]
    public string String1 { get; set; }

Both have their benefits, but I have a personal preference towards the second.
I think it gives better discoverability of available options. If you have found the XLColumn attribute intellisense will give you all possible options you can set as named attributes.

vbjay:
I agree. Ignore is too comon and vague. I like the XL part of the
attribute. Makes it clear that it is Excel related.

albinsunnanbo:
The question is how to handle the `ColumnOrderAttribute`. Refactor and move to a named parameter of a `XLColumn` attribute is a breaking change, but leaving outside makes it a bit odd if we want to have a consistent API.

igitur:
I agree. I think the `XLColumn` is better. We're still sub v1.0, so I don't really care about a consistent API. The attributes were added fairly recently anyway.

albinsunnanbo:
@igitur I suppose you mean you prefer a consistent API over a stable API at this time?

albinsunnanbo:
You could keep the old attribute for a few releases with an `ObsoleteAttribute` informing about the new API.

igitur:
Not merging this anymore.

PR #209, opened 2017-02-17
Implement new XLColumnAttribute

igitur:
Replace the current Display() and ColumnOrder attributes with new XLColumn attribute, which also allows user to ignore properties.

Also improve performance of inserting table. Fewer Reflection calls.


ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -2479,23 +2491,6 @@ private XLCell CellShift(Int32 rowsToShift, Int32 columnsToShift)
             return Worksheet.Cell(Address.RowNumber + rowsToShift, Address.ColumnNumber + columnsToShift);
         }
 
-        private static String GetFieldName(Object[] customAttributes)
-        {
-#if NET4
-            var attribute = customAttributes.FirstOrDefault(a => a is DisplayAttribute);
```
albinsunnanbo: It looks like you dropped the support for the DisplayAttribute.
If think it is useful to keep. It is used by ASP.NET, etc to define display names.
I use DisplayAttribute to share column names between ASP.NET and ClosedXml. I think DisplayAttribute.Name should be used as a fallback if XLColumnAttribute.Name == null.
That way you can override the DisplayAttribute if you want something different, but keep a single column name definition where you want it to be the same everywhere.

igitur: No, I'm keeping only one attribute.

PR #210, opened 2017-02-17
Use FastMember instead of slow Reflection

igitur:


igitur:
Merge #209 before this one.

Pitterling:
I tried this PR with my example coding. I don't know if this is in a range what you expected.

Insert 50.000 rows - v0.86
InsertData Time elapsed: 00:00:01.**15**80346

Insert 50.000 rows - Use FastMember instead of slow Reflection #210
InsertData Time elapsed: 00:00:01.**00**96845

Issue #175, opened 2017-01-25
Allow references to worksheet with a space in its name


andrewkress:
when print area has comma in the definedName.Text, ie:
'C100,C101,102,103,C200'!$A$1:$F$271

the line:
`foreach (string area in definedName.Text.Split(','))`
in XLWorkbook_Load.cs does not parse it correctly resulting in the call to ParseReference to throw the index out of range exception when assigning the sheetArea variable.

This will occur when there is a comma in any Sheet name that has a Print Area defined.  Attached is a file that will replicate this issue.
[Test Workbook.xlsx](https://github.com/ClosedXML/ClosedXML/files/730695/Test.Workbook.xlsx)

andrewkress:
I created a fix for this in my repo:
https://github.com/andrewkress/ClosedXML/tree/issue175-parse-commas-in-print-area

Let me know what is required and I can create a pull request.

andrewkress:
#177 Tagging the pull request

igitur:
I'm considering moving to https://github.com/spreadsheetlab/XLParser for the entire parsing.

andrewkress:
If there is an existing library that this functionality is duplicating, that would make sense.  What aspects could we use it to parse?

andrewkress:
I can use this library to fix the current issue.  I will create a new pull request as that will be cleaner.  If there are other instances that the library can be used to parse information, that can be done in other pull requests.

andrewkress:
Alright, digging in more, this maybe useful...I'll play with it.

igitur:
If you can upgrade, I would recommend this. It's getting difficult to maintain the .NET 3.5 dependency.

PR #211, opened 2017-02-20
Issue175 use xlparser to parse print area

andrewkress:
#175 is the main issue addressed.

Bigger changes include removing the .NET 3.5 project build and updating the minimum .NET version to 4.5.

These changes were required to use the XLParser Library to allow more accurate parsing of formulas and area references.


ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -933,7 +933,7 @@ private static bool IsRowReference(string sheetArea)
         private static void ParseReference(string item, out List<Tuple<string, string>> sheetNameAndArea) 
         {
             sheetNameAndArea = new List<Tuple<string, string>>();
-            ParseTreeNode nodes = ExcelFormulaParser.Parse("Test(" + item + ")");
+            ParseTreeNode nodes = ExcelFormulaParser.Parse(String.Format("DummyFormula({0})", item));
```
andrewkress: For the library to parse the print area, it had to be mocked up in a formula format.  Once parsed, the information is dropped.

igitur:
This is good work, but there are some issues to consider. Dropping of .NET 3.5 is potentially problematic. I don't have metrics of the number of users of the release, but recently someone commented on another issue and I realised he was using .NET 3.5.

Also, .NET Standard support is a often requested feature, and XLParser and its underlying Irony will first have to support .NET Standard. I did ask about it at https://irony.codeplex.com/discussions/661482 , but the project looks pretty dead. The last commit was in 2013 (I didn't realise that earlier) and it makes me hesitant to have it as a dependency if the library isn't actively developed. I've asked them whether they plan to support .NET Standard at https://irony.codeplex.com/discussions/661482

ClosedXML/Excel/XLWorkbook_Load.cs
igitur: Yes, I think you can just wrap it in `(` and `)` too, if I remember correctly.

andrewkress:
Yes, those are both good points and need to be considered if this is to get merged.  Using these libraries becomes a nightmare if you need to still support .NET 3.5.  
At almost 10 years old, are there situations where someone does not have a newer version of .NET installed?
As for support of .NET Standard, we can wait and see.  It appears that the maintainer is still active in the discussion there, but it doesn't appear that much has changed in the codebase in years.

andrewkress:
Closing this, didn't realize it was still open. The issue was resolved with a different pull request.

PR #227, opened 2017-03-04
Implement new XLColumnAttribute

albinsunnanbo:
Based on #209 
But keeping the `DisplayAttribute(Name=...)` as fallback when `XLColumnAttribute(Header=...)` is not defined.
To be able to reuse `DisplayAttribute`s used in ASP.NET models according to https://github.com/ClosedXML/ClosedXML/pull/209#pullrequestreview-22456959


albinsunnanbo:
`XLColumnAttribute` as discussed in #205 
Does not solve #205, but adds a proper API for ignoring columns. thus reducing the risk of running into the problem in #205

albinsunnanbo:
I noted that this pull request was not compatible with develop/HEAD
I rebased to make it possible to merge.

Can I assist in any way to get this pull request merged?

igitur:
Thanks, but I don't want to add unnecessary complexity by supporting 2 attributes. We are still pre-v1, so the API can change without notice. You can still use both `DisplayAttribute` and `XLColumnAttribute` on your properties if you wish. 

Does this PR add anything else except fallback support for `DisplayAttribute`?

albinsunnanbo:
No, it only adds the fallback.
But I insist that the `DisplayAttribute` is a relevant feature. Not for backwards compatibility, but since it makes sharing models with ASP.NET easier and cleaner.

igitur:
Maybe I'm missing something. But the only downside to my suggestion is that you would need 2 attributes in your models instead of 1. Am I right? If so, I think that's a fairly insignificant downside compared to the overhead and possible confusion of maintaining two attribute classes.

vbjay:
@igitur  Frankly this would be a great idea.  Think of it this way.  We have the xlColumn attribute for the cases where we want our excel export to be different from our model.  If we have an xlColumn then it wins.  If not use the same display we expect from Display attribute.

igitur:
@vbjay If the only benefit is to avoid a 2nd attribute on a property, I don't think it's worth the overhead in the code. Are there other benefits?

albinsunnanbo:
Yes. "only".
But if you have lots of models with the `DisplayAttribute` for other purposes (like ASP.NET) it would be nice if ClosedXML would just work out of the box without having to go through all your models and replicating all display attributes. Since `DisplayAttribute` is built into the .NET framework and is an ubiquitous attribute for annotating properties with display name I think is it a nice courtesy to support it in ClosedXML.

Being 13 LOC overhead 

https://github.com/ClosedXML/ClosedXML/pull/227/commits/7861ebaeda5f04da0037769c3b1a05111ef0c185#diff-3b8a562c15b0aca901df66d96df87e10L2481
and
https://github.com/ClosedXML/ClosedXML/pull/227/commits/7861ebaeda5f04da0037769c3b1a05111ef0c185#diff-3b8a562c15b0aca901df66d96df87e10R505
and
https://github.com/ClosedXML/ClosedXML/pull/227/commits/7861ebaeda5f04da0037769c3b1a05111ef0c185#diff-3b8a562c15b0aca901df66d96df87e10R631

the overhead is not very large. I agree that it adds to both API and implementation complexity, but personally I think the benefits are larger.

vbjay:
It is not overhead.  It is gracefully falling back to a default the developer meant.  I don't see the point in this
```
[Display(Name="First Name"]
[xlColumn(WhateverDisplayNameProp="First Name"]
string FirstName{get;set;}
```

I do see the point in this

```
[Display(Name="First Name"]
[xlColumn(WhateverDisplayNameProp="FName"]
string FirstName{get;set;}
```

It is not hard to do a simple GetDisplayName function that checks if xlcolumn exists if so use it for the display
If not check is there a display if so, use it
If not, use the  name as is.

igitur:
There is additional complexity in that the .NET 3.5 build of `ClosedXML` uses `System.ComponentModel.DisplayNameAttribute` and .NET 4.0+ uses `System.ComponentModel.DataAnnotations.DisplayAttribute`.

vbjay:
Frankly, with the other issues that make it clear 3.5 is hard to support, I
say dump it.

igitur:
Sorry, I don't think this PR is necessary.

PR #228, opened 2017-03-04
Set ColumnOrderAttribute Obsolete

albinsunnanbo:
Restore the ColumnOrderAttribute removed in #227, but set it as Obsolete to help developers migrate the the new XLColumnAttribute

Implemented a test as a "hack" to implement a compile time reminder to remove the obsolete attribute before releasing v1.0.

Implemented this addition as a separate pull request from #227 to make it easy to merge the #227 without keeping the ColumnOrderAttribute as obsolete.

Merge #227 before merging this pull request since this includes all commits in #227


igitur:
While I appreciate your input a lot, I don't think this is required. We are still pre-v1.0, and I don't think it's necessary to keep a stable API (within reason, of course). I don't think dropping the old attribute in favour of the new one is a major change and that we should keep an obsolete class.

I'm waiting for @Pyropace for his thoughts on this.

albinsunnanbo:
That's fine with me, I made this a separate pull request to make it possible for you to decide whether or not to include it at your own discretion.

igitur:
Sorry, but I don't think this PR is necessary.

Issue #230, opened 2017-03-06
0.86 Sequence contains more than one matching element


BenKmann:
After i upgrade ClosedXML from 0.85 to 0.86 i can't save my Workbooks anymore.
When i downgrade to 0.85 everything works again. 

Stacktrace:
```
System.InvalidOperationException
Sequence contains more than one matching element.
StackTrace:
   bei System.Linq.Enumerable.Single[TSource](IEnumerable`1 source, Func`2 predicate)
   bei ClosedXML.Excel.XLWorkbook.GenerateWorkbookStylesPartContent(WorkbookStylesPart workbookStylesPart, SaveContext context) in C:\Git\ClosedXML\ClosedXML\Excel\XLWorkbook_Save.cs:Zeile 2610.
   bei ClosedXML.Excel.XLWorkbook.CreateParts(SpreadsheetDocument document) in C:\Git\ClosedXML\ClosedXML\Excel\XLWorkbook_Save.cs:Zeile 252.
   bei ClosedXML.Excel.XLWorkbook.CreatePackage(String filePath, SpreadsheetDocumentType spreadsheetDocumentType, Boolean validate) in C:\Git\ClosedXML\ClosedXML\Excel\XLWorkbook_Save.cs:Zeile 104.
   bei ClosedXML.Excel.XLWorkbook.Save(Boolean validate) in C:\Git\ClosedXML\ClosedXML\Excel\XLWorkbook.cs:Zeile 452.
   bei ClosedXML.Excel.XLWorkbook.Save() in C:\Git\ClosedXML\ClosedXML\Excel\XLWorkbook.cs:Zeile 432.
```

And my Code
```c#
var wb = new XLWorkbook(FILEPATH);
wb.Save();   // Just Opening und Saving is Enough.
```
I hope the Stacktrace is enough. Happens with different Excel Files on my side. 
If you need more Information, just ask.

igitur:
Please attach a workbook that produces the problem.

BenKmann:
[Blanko_clean.xlsx](https://github.com/ClosedXML/ClosedXML/files/821345/Blanko_clean.xlsx)

igitur:
I can't reproduce the problem with the latest source code from the `develop` branch and with the file that you supplied.

BenKmann:
Can you reproduce it with the 0.86 Nuget Package? 
Maybe with this Document? 
[...]

PR #233, opened 2017-03-07
Don't crash when file contains multiple default styles

igitur:
It's possible to have duplicate default cell styles - occurs when file gets saved under different cultures.

We prefer the style that is named 'Normal'


igitur:
Fixes #230

PR #235, opened 2017-03-08
Use ClosedXML specific exceptions instead of generic Exception

aakasaka:
Throwing an instance of Exception is a bad practice.
ref: https://msdn.microsoft.com/en-us/library/ms229014(v=vs.100).aspx


igitur:
Thanks. PR looks good. Would you mind resolving the conflict (after we updated the develop branch). Also check that all exception classes are available in the .NET Standard library.

aakasaka:
Thank you for your reply. I resolved the conflict.
>Also check that all exception classes are available in the .NET Standard library.

I used typical exception classes only, except ExpressionParseException class.
It's an original exception class thrown by CalcEngine.

igitur:
Thanks, so can you confirm that besides `ExpressionParseException` the other exception classes are included in .NET Standard?

igitur:
One more request. Please add a unit test to test that the `ExpressionParseException` is thrown when it should. I don't think it's necessary to unit test the other exceptions.

aakasaka:
>Thanks, so can you confirm that besides ExpressionParseException the other exception classes are included in .NET Standard?

Yes. all the other exceptions are included in .NET Standard.

And I  specified all exception types in test codes.
`ExpressionParseException` is thrown only one place in the tests.
I think it's used just for assertion in the other places.

igitur:
I'd like to include this in v0.89. Can you please fix the conflicts?

aakasaka:
Thanks, I fixed the conflicts.

PR #263, opened 2017-04-07
Parse commas in print area

### LINKED ISSUE #175: Allow references to worksheet with a space in its name

igitur:
Fixes #175  .

Thanks @andrewkress

I also had to update the calculation engine to allow references to sheets with commas in the name.


igitur:
@andrewkress Please check whether you're happy with this PR.

andrewkress:
looks good to me.

PR #269, opened 2017-04-12
Evaluate formulas when saving

### LINKED ISSUE #252: References damaged after saving workbook
This is a simplest table : [1.xlsx](https://github.com/ClosedXML/ClosedXML/files/886217/1.xlsx)

First column is base value. second column is reference to first one. And I do this :

```c#
wb = new XLWorkbook("1.xlsx");
wb.Worksheet("Sheet1").Cell("A2").Value = "1";
wb.Save();
```

Cell value is updated properly, however all references are set to null. Not just this one affected by an updated value. All of them. When I open this file in Excel, on load, references are probably updated and that's why when I close Excel, it asks if this file should be resaved (even if I didn't do anything to it, but references were updated). Bigger problem is that all 3rd party applications that read this XLSX via ADO.NET, can never see those values. On read, they all are showing null values in all reference destination cells in the all worksheets.

I tried wb.ForceFullCalculation = true, I tried wb.CalculationOnSave = true;. I tried all possible combinations of wb.ReferenceStyle and wb.CalculateMode. Nothing makes those references reappear for ADO.NET based XLSX readers. The only way to restore them is to open Excel and resave this file.

Is there any way to restore references before save?

Regards,
Wojtek

igitur:
Fixes #252  .

Adds option in `XLWorkbook.Save()` and `XLWorkbook.SaveAs()` to evaluate formulas when saving workbook.


igitur:
@vojtek11 says this issue doesn't fix the problem yet. So don't merge yet.

PR #277, opened 2017-04-18
C#6 syntax

igitur:
Changed CI to use Visual Studio 2017.

Let's see if the build now supports C#6 syntax.


igitur:
Solution file upgrade was required for CI to detect and use MSBuild 14.

PR #282, opened 2017-04-19
Fix comments getting setting

Knight2601:
Fixes # .

how to recreate:
 -  add comments to cells in excel sheets
 -  load the file and read the comments

Changes proposed in this pull request:
 -  fixes the getting of existing comments which was flakey

How did I test this code:
 - sandboxed multiple excel files from different origins, the problem was mainly detectable from sharepoint excel files.
 
- [ ] I attached a sample spreadsheet.  (You can drag files on to this pull request)


igitur:
Please explain what this is about. 

Also, you shouldn't be changing the Sandbox project. You should be adding a unit test though.

igitur:
@Pyropace I disagree. There are no unit tests and the sandbox project is changed, which shouldn't. I'm also awaiting feedback about what is lacking with the current comments and how this change improves it. "fixes the getting of existing comments which was flakey" doesn't say much about this change.

@Knight2601 Please respond when you have time.

vbjay:
@Knight2601 This is why you create branches for pull requests. You don't want to mix changes from different features getting mixed together.  Close this pull request and create another using a branch that just includes the related commits.

PR #284, opened 2017-04-24
Add image support

### LINKED ISSUE #48: Image support

clinchergt:
Adds basic support for image manipulation using absolute, onecell,
and twocell anchors. It is based on ajwhiteway's fork:
https://closedxml.codeplex.com/SourceControl/network/forks/ajwhiteway/ClosedXMLImageSupport

Fixes #48.

Changes proposed in this pull request:
 - Adding Marker and Picture classes
 - Adding them to the worksheet using different anchors depending on the parameters
 
How did I test this code:

The current tests add images using:
 - `AbsoluteAnchor`
 - `OneCellAnchor`
 - `TwoCellAnchor`
 - Images in different worksheets
 - Images in PNG format
 - Images in JPEG format
 - Class tests for XLPicture
 - Class tests for XLMarker
 
TODO: 

 - [x] Add support for other formats
 - [x] Add tests

This should be ready for review now.


igitur:
Thanks foe the contribution. Please add some unit tests.

igitur:
I notice now you have a todo list that mentions the unit tests.

clinchergt:
@igitur so, I've run into an interesting issue. [The blip r:embed property](http://officeopenxml.com/drwPic-ImageData.php) changes every time you add an image, even if it's the same image. This leads to the compares failing even if the files are actually literally the same and only that part varies.

Thoughts on how this could be handled? I could upload the preliminary test I wrote for you to try out and see what I'm talking about.

clinchergt:
This should be ready for review now

igitur:
Thanks for your help with this. Appreciate it a lot. Would you mind rebasing it on the `ajwhiteway-image-support` branch? That is because I'd like to keep ajwhiteway's commit history and yours separate. I would then be able to review the changes you and ajwhiteway did separately. There will be some minor conflicts that you'll have to resolve.

clinchergt:
@igitur sure, I should've done that from the start actually. I just never noticed that branch.

It is not passing tests according to github tho, will that be an issue? Any idea as to why it's not building/passing tests?

clinchergt:
@igitur Did you port those commits over manually? You added a few extra `ClosedXML` directories in the process. Rebasing on that branch in its current state is kinda useless and the file locations don't match. E.g.:

`ClosedXML/ClosedXML/ClosedXML/Excel/Drawings/IXLMarker.cs`
vs 
`ClosedXML/Excel/Drawings/IXLMarker.cs`

igitur:
OK, will have a look at it next week. Stuck on a Linux machine for a few days and can't run unit tests for ClosedXML on it.

clinchergt:
@igitur Would you be okay with the rebase [here](https://github.com/clinchergt/ClosedXML/tree/ajb-is)?

I added ajwhite's original commits and rebased onto the current develop branch. If it seems fine, I'll push those changes to this PR's branch so you guys can keep reviewing this.

igitur:
Hang on. Let's take this step by step otherwise it's going to become very confusing. I fixed the issues with my previous commit and pushed to the `ajwhiteway-image-support`. I also had to do some minor work to get it working in the .NET3.5 build. 

You can now proceed to rebase your work on the `ajwhiteway-image-support` branch. Let me know if you run into serious issues.

vbjay:
Thought you dropped 3.5 support in #292

igitur:
> Thought you dropped 3.5 support in #292

#292 hasn't been merged yet and I haven't prioritised it yet. So for now, .NET3.5 is still alive.

vbjay:
I see. GitHub says it is merged.

clinchergt:
@igitur I kinda wanted to stray away from your branch because it includes an unrelated commit by ajwhite (which adds filter functionality, which should be outside the scope of this PR), and adds a lot of unnecessary commits.

Personally, I find this PR's branch does a good job at preserving ajwhite's authorship without adding a lot of unnecessary commits. Basically what your branch would look like with aggressive rebasing (this branch also initially had the 3.5 fixes, but I dropped them when I saw the relevant PR was merged).

I can rebase my changes onto your branch, but I wanted to share my perspective on the matter beforehand. Thoughts?

igitur:
@clinchergt Yeah, that's OK. Continue with your approach. Just point out where you deviated from ajwhiteway's commits. I can that `XLMarker.ColumnId` is different, for example (off by one). I assume yours is the correct one, but I want to check all these differences.

igitur:
I suppose I have to add developer guidelines at some stage, but in the meantime, please do the follow cosmetic changes:

- Install EditorConfig for your IDE and reformat the code files that you touched. Leave other files in unchanged. You might have to close and reopen Visual Studio after installing EditorConfig.
- Please sort the `using` statements at the top of the file alphabetically.
- Ensure no tabs for indentation (EditorConfig should do this too).

clinchergt:
Ah, that's fair. Apparently I missed some of my changes when rebasing. I'll fix that and let you know when I'm done.

ClosedXML/Excel/XLWorksheet.cs
```diff
@@ -1491,5 +1492,19 @@ public Object Evaluate(String expression)
         }
 
         public String Author { get; set; }
+
+        public List<Drawings.IXLPicture> Pictures()
+        {
+            return pictures;
+        }
+
+        public void AddPicture(Drawings.XLPicture pic)
```
Pyropace: 🔴 I am thinking if List is a good choice here ? what if i add a same picture more than once, would be nice if we can check for item you are trying to add and add only if it doesnot exisits ,, we have enogh memory issues at a moment

ClosedXML_Tests/Examples/ImageHandlingTests.cs
```diff
@@ -0,0 +1,21 @@
+﻿using ClosedXML_Examples;
+using NUnit.Framework;
+
+namespace ClosedXML_Tests.Examples
```
Pyropace: 🔴 can you add tests to specifically test your public methods of your class ?

ClosedXML/Excel/XLWorksheet.cs
clinchergt: Should duplicates really be of ClosedXML's concern? If the user is worried about memory usage they should be taking precautions to avoid duplicates.

ClosedXML_Tests/Examples/ImageHandlingTests.cs
clinchergt: Done.

Pyropace:
👍 LGTM 

@igitur ?

ClosedXML/Excel/XLWorksheet.cs
Pyropace: so ⚪️  is nice to have, yea it be great if you can implment it, it would be nice, if we are smart when it comes to storing 

keep in mind, this can be tricky , identifying the "same picutre" , also please evaluate the cost of doing it

vbjay:
Instead of a list, how about a dictionary <guid, image>. The user can tag
their images with a guid and decide if they want to create a new guid or
use the same image.

igitur:
I think this is a very good start. I would like to improve the API. I don't think the current API is very user-friendly. But I think we should merge this first and then continue the work for the improvements.

igitur:
Why can't I move the images around when I open a generated Excel file? I can't find a way to 'unlock' them.

clinchergt:
>Why can't I move the images around when I open a generated Excel file? I can't find a way to 'unlock' them.

Ah, this is due to the `PictureLocks` properties I set. I blindly restricted all of them.

I guess they should either depend on the type of anchor used or add them as attributes of the `XLPicture` and set them accordingly.

As for making the API more user-friendly, maybe going in the opposite direction would help. The `NoChangeAspect`, `NoMove` and `NoResize` properties could determine the type of anchor used (assuming the correct amount of markers were added).

Not really sure how else it could be more intuitive/friendly, we're gonna need suggestions from users.

igitur:
@clinchergt Instead of the generic ClosedXML images, let's immortalize you for your contribution. Shall we use your Github profile picture?

clinchergt:
@igitur I actually don't own my profile picture (I'm not even sure where I got it from anymore, bleh) so I'm not sure if it's under CC or anything that would allow us to use it. Can't think of an image I identify with either.

Feel free to change the generic images I used tho!

igitur:
I totally reworked the image API in #323 . Check out the `ImageAnchor` and `ImageFormat` example of how it works now. I think it's much more intuitive.

PiotrPerak:
Can this be scheduled to go with new Milestone please? Lack of picture support is a show stopper for me unfortunately.

Issue #225, opened 2017-03-03
Upgrade OpenXML to v2.7.1


Pitterling:
Hi,

in the meantime OpenXML 2.7.1 has been released.

[Open-XML-SDK @Github ](https://github.com/OfficeDev/Open-XML-SDK)
[Nuget package](https://www.nuget.org/packages/Joelharkes.DocumentFormat.OpenXml/2.7.1)

Any plans to upgrade to 2.7.1?

Thx,
P.

igitur:
What are the benefits of upgrading?
And is .NET 3.5 supported?

Pitterling:
[Open-XML-SDK changes](https://github.com/OfficeDev/Open-XML-SDK#change-log)
mainly fixes and newer .NET versions

it seems that minimum .NET release is now 4.5. Why is .NET3.5 still required?

igitur:
Some people are still on .NET 3.5. I don't have metrics of how many, but recently a commenter on an issue mentioned he is on .NET 3.5.

twsouthwick:
If you're interested in support for .NET 3.5, we can definitely take a look at adding it. We're hoping to consolidate on one package to simplify dependency management, and if .NET 3.5 is needed, we can have a discussion on that.

twsouthwick:
Is .NET 3.5 support a blocker for updating to the latest DocumentFormat.OpenXML? It will bring .NET Standard support so you can take your library to .NET Standard as well. I have a PR to support .NET 3.5, but at the moment, the .NET CLI doesn't understand .NET 3.5 and so we'll have to change our build process. So, the question is, how big of a user base are still using .NET 3.5 and how firm is the need to support it?

igitur:
@twsouthwick Yes, that is a blocker.

Figuring out how many users actually use the .NET 3.5 is exactly what has been keeping me busy recently. I was hoping NuGet would have some metrics on the issue, but, unless I'm mistaken, that level of detail isn't stored. 

The only measure I have currently is based on feedback from users in the comments in the issues here on Github. 

My personal biased view is that there are more benefits in supporting .NET standard than supporting .NET 3.5, so if it has to come down to a choice, I'd go with .NET standard. I also consider `ClosedXML` as pre-v1 and don't feel we have to be as stable as post-v1 libraries, but I don't want to alienate users.

Looking for feedback from anybody...

twsouthwick:
[...]

PR #292, opened 2017-05-01
Upgrade OpenXML to v2.7.1

igitur:
Drop .NET3.5 version of ClosedXML
Upgrade ClosedXML to .NET4.5.2 (required for new OpenXML version)
Upgrade OpenXML to v2.7.1


igitur:
Fixes #225

Issue #300, opened 2017-05-08
Performance issue of creating merged cells


wtetsu:
**Do you want to request a *feature* or report a *bug*?**
- [ ] Bug
- [x] Feature

**What is the current behavior?**

Creating hundreds of merged cells takes long time.

**If the current behavior is a bug, please provide the steps to reproduce and 
if possible a minimal demo of the problem with a sample spreadsheet.**

For example:

- Create a row with merged cell.
- Copy the row several hundreds times(The code will be described later)

It gradually get slower and slower.

**What is the expected behavior or new feature?**

I hope merged cells creation finished faster.

**Did this work in previous versions of our tool?  Which versions?**

- [X] I attached a sample spreadsheet.  (You can drag files on to this issue)

- develop branch(0810d0d80155fb01ace3aa28e2f4c8919281cfa0)

**Code to reproduce problem:**
```c#
using (var wb = new XLWorkbook(@"template.xlsx"))
{
    var ws = wb.Worksheets.Worksheet(1);
    var firstRow = ws.FirstRow();

    var startTicks = DateTime.Now.Ticks;

    using (var range = ws.Range(1, 1, 1, 20))
    {
        for (int i = 0; i <= 500; i++)
        {
            using (var rangeTo = ws.Range(2 + i, 1, 2 + i, 20))
            {
                range.CopyTo(rangeTo);
            }
            if (i % 10 == 0)
            {
                Console.WriteLine((DateTime.Now.Ticks - startTicks) / 10000);
            }
        }
    }

    wb.SaveAs(@"template_new.xlsx");
}
```

[template.zip](https://github.com/ClosedXML/ClosedXML/files/983894/template.zip)

![chart01](https://cloud.githubusercontent.com/assets/515948/25810394/f8b8e376-344a-11e7-9750-e95f6d8db88d.png)

If there are no merged cells it takes less than 50 ms.

igitur:
Thanks for you contribution. Keep them coming.

PR #302, opened 2017-05-08
Performance improvement on creation of merged cells

wtetsu:
Performance improvement on creation of merged cells.
#300

```c#
using (var wb = new XLWorkbook(@"template.xlsx"))
{
    var ws = wb.Worksheets.Worksheet(1);
    var firstRow = ws.FirstRow();

    var startTicks = DateTime.Now.Ticks;

    using (var range = ws.Range(1, 1, 1, 20))
    {
        for (int i = 0; i <= 500; i++)
        {
            using (var rangeTo = ws.Range(2 + i, 1, 2 + i, 20))
            {
                range.CopyTo(rangeTo);
            }
            if (i % 10 == 0)
            {
                Console.WriteLine((DateTime.Now.Ticks - startTicks) / 10000);
            }
        }
    }

    wb.SaveAs(@"template_new.xlsx");
}
```

chart02

The performance is fairly improved, but is there a possibility that this change causes side effects?

### DIFF
--- a/ClosedXML/Excel/Ranges/XLRangeBase.cs
+++ b/ClosedXML/Excel/Ranges/XLRangeBase.cs
@@ -344,11 +344,15 @@ namespace ClosedXML.Excel
         {
             if (checkIntersect)
             {
-                string tAddress = RangeAddress.ToString();
-                foreach (var mergedRange in Worksheet.Internals.MergedRanges)
+                using (IXLRange range = Worksheet.Range(RangeAddress))
                 {
-                    if (mergedRange.Intersects(tAddress))
-                        Worksheet.Internals.MergedRanges.Remove(mergedRange);
+                    foreach (var mergedRange in Worksheet.Internals.MergedRanges)
+                    {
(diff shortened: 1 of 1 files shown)


igitur:
Impressive! I'll check it out in detail a bit later.

Pyropace:
@wtetsu impressive indeed, what did you use for the graph ?

wtetsu:
Thank you for your comments!

@Pyropace  I made it in Excel.

igitur:
Would you mind attaching your `template.xlsx`? I'd like to inspect what the reason for the original slowdown was. I suspect it's somewhere in `XLWorksheet.Range(String rangeAddressStr)`.

igitur:
I see the bottleneck is in `XLWorksheet.RangeShiftedRows` which is of type `XLReentrantEnumerableSet<XLCallbackAction>`. That class suffers from poor performance. But I think there are other reasons too.

igitur:
Fixes #300

wtetsu:
Thank you!

PR #306, opened 2017-05-11
Register hyperlink with 2 params in CalcEngine

TDaphneB:
When trying to parse a hyperlink that also has a tool tip/friendly text (2nd parameter), an exception of "Too many parameters" is thrown. Changing the max params to 2 allows these hyperlinks to be read.

#### Questions:
- Is there a blog post?


igitur:
Please add some unit tests.

TDaphneB:
Will add unit tests.

igitur:
You didn't need to close the PR in order to add unit tests. You can just add commits to it.

TDaphneB:
Ah, thanks, I did not realise this.

TDaphneB:
Looks like I am having an problem with the unit test running on AppVeyor. The test passes locally, any advice?

igitur:
It fails locally for me. The order of the hyperlinks are different in the reference and produced files.

image

igitur:
Column widths differ.

TDaphneB:
Thanks, that is a good place to start looking. I will have a look, I noticed the test uses AdjustToContents on the column, is this something that could produce a different result in a different environment and why it could pass locally for me but not elsewhere?

igitur:
If you run the tests on Linux/Mono, the column widths are ignored, but I would be surprised if that's your case. 

Are the other tests passing? 

Can you try manually comparing the contents of the reference file with the generated file? I use Total Commander for this purpose, as in the screenshot:
image

TDaphneB:
You are right - I am not on Linuz. Windows 10, Office 2016.

Not all of the tests pass. I get this result:
unit tests

Here is a compare of the XLSX files with which it has been passing:
winmerge

And how they appear in Excel:
excel

If it is ok with you perhaps I could test setting the column width to an actual value instead of the adjust to contents just to check if that passes on AppVeyor?

igitur:
I think this might be some issue related to Windows 10 or the DPI that changes how the width of the text is calculated. I'll create a separate issue for that. In the meantime, I'll fix your test file from my machine.

TDaphneB:
Awesome, thanks for your help =)

igitur:
Can you please create an issue for those other tests that are failing? Choose a failing test that is based on one of the examples (in `ClosedXML_Examples`), preferably a simple one. Attach the generated Excel file.

Issue #101, opened 2016-11-16
Multi-targeting: net40, net461, netstandard2.0


ksakz:
is it meant to be?

igitur:
The OpenXML library has to support NET Core first before we can attempt it. ClosedXML is just a wrapper around OpenXML

igitur:
Depends on https://github.com/OfficeDev/Open-XML-SDK/issues/111

rogersachan:
@igitur I thought supporting .NET Standard 1.0 makes it easier to support .NET Core.

Pyropace:
@igitur it looks like 111 is closed, perhabs we can start looking into this 

this was merged in https://github.com/OfficeDev/Open-XML-SDK/pull/121

I will assign this to my self and see if i can get it compiling

Pyropace:
Well things are looking alot better than last time i tried this. 

1- OpenXML now  is dot net standard compatible ( however an official package hasnot been released to nuget  yet,(version 3.0.0)  tried the latest version from their source  and built locally )
2- We still need corefx team to finish their work with System.Drawing library, (https://github.com/dotnet/corefx/issues/2315), however using (https://github.com/zkweb-framework/ZKWeb.System.Drawing) can get you to compile 
3- DatTables, datasets are not supported in dotnet standard, so we would have to change all the codes that use DataTable ( not a bad idea ) 
4- we use OADate alot in our code, that also has been dropped as it is windows specific, we would have to replace codes related to that as well. 

I got a version to compile, but it fails a lot of tests, I might create a branch and push what i have so far, and once OpenXml 3.0.0 is released, we can get it to work 

but we would have to make alot of code change

MKI-Miro:
Hi, anything new ?

moattarwork:
+1

danilobreda:
With the new .net standard 2.0. When will it be ported to .net standard?

christophwille:
I am also going through various projects that are candidates for Core 2.0 porting atm (OSS as well as enterprise) and ClosedXML is one that I really like & rely on heavily. As this issue is marked as Parked atm, any plans to make this happen?

igitur:
You can follow the progress at #315 . Even better, you can help contribute.

igitur:
Dependent on https://github.com/OfficeDev/Open-XML-SDK/issues/226

igitur:
[...]

PR #315, opened 2017-05-15
Multi-targeting: net40, net461, netstandard2.0

igitur:
Work in progress


igitur:
Fixes #101 , #450

igitur:
@Pankraty I'd appreciate a 2nd opinion on this PR. I'll merge after feedback from you.

ClosedXML_Tests/Excel/CalcEngine/DateAndTimeTests.cs
```diff
@@ -12,7 +12,7 @@ namespace ClosedXML_Tests.Excel.DataValidations
     [TestFixture]
     public class DateAndTimeTests
     {
-        [OneTimeSetUp]
+        [SetUp]
```
Pankraty: Maybe extract ```SetCultureInfo``` into GlobalTestSetup with ```SetUpFixture``` attribute?

ClosedXML_Tests/Excel/ImageHandling/PictureTests.cs
```diff
@@ -12,6 +13,7 @@ namespace ClosedXML_Tests
     [TestFixture]
     public class PictureTests
     {
+#if _NETFRAMEWORK_
```
Pankraty: It's a matter of personal preference, but I think having a check of the platform inside the test method will indicate more clearly that some parts of the code are not fully cross-platform. I.e.
```
        [Test]
        public void CanAddPictureFromBitmap()
        {
#if !_NETFRAMEWORK_
            Assert.Ignore("Platform is not supported");
#else
            ...
```

igitur:
@pankraty Valid comments, but I was looking for comments related to this PR. The other comments should be addressed in other PRs so that they don't delay this one.

Pankraty:
I understand, I just need more time. I skimmed through changes not even running the application yet.

ClosedXML/ClosedXML.csproj
```diff
@@ -1,386 +1,53 @@
...
-    <DefineConstants>TRACE</DefineConstants>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
-    <DocumentationFile>bin\Release\ClosedXML.xml</DocumentationFile>
-    <NoWarn>1591</NoWarn>
-    <Prefer32Bit>false</Prefer32Bit>
-    <LangVersion>6</LangVersion>
+
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
+    <DefineConstants>$(DefineConstants);_NETSTANDARD_;_NETSTANDARD2_0_</DefineConstants>
```
Pankraty: I noticed an inconsistency in defined constants. Here we have ```'$(TargetFramework)' == 'netstandard2.0'``` and ```_NETSTANDARD2_0_```, in Examples and Sandbox - ```'$(TargetFramework)' == 'netcoreapp2.0'``` and ```_NETSTANDARD2_0_``` as well, and in Tests - ```'$(TargetFramework)' == 'netcoreapp2.0'``` and ```_NETCOREAPP2_0_```

ClosedXML_Sandbox/ClosedXML_Sandbox.csproj
```diff
@@ -1,92 +1,40 @@
...
+    <Reference Include="System.ComponentModel.DataAnnotations" />
   </ItemGroup>
+
   <ItemGroup>
-    <None Include="..\.editorconfig">
-      <Link>.editorconfig</Link>
-    </None>
-    <None Include="app.config" />
-    <None Include="packages.config" />
-    <None Include="test.xlsx">
```
Pankraty: CopyToOutputDirectory was lost, and the application fails after a clean build.

ClosedXML/ClosedXML.csproj
```diff
@@ -1,386 +1,53 @@
...
+
   <ItemGroup>
-    <None Include="..\.editorconfig">
-      <Link>.editorconfig</Link>
-    </None>
-    <None Include="ClosedXML.snk" />
-    <None Include="packages.config">
-      <SubType>Designer</SubType>
-    </None>
+    <PackageReference Include="DocumentFormat.OpenXml" Version="2.7.2" />
```
Pankraty: Did you consider upgrading to 2.8.1? (https://github.com/OfficeDev/Open-XML-SDK/tree/v2.8.0#change-log)

Pankraty:
I did not encounter any serious issues with this PR, and I like how much lighter .csproj files become. The overall performance got better too (did not test it too carefully but found a new version is about 10% faster). Assuming that netstandard support is desired by many users I think this overweights some breaking changes that have been introduced. Do you want to release 0.92 soon after the merge?

ClosedXML/ClosedXML.csproj
igitur: I try to keep to the lowest common denominator. Users are free to upgrade to a higher version of OpenXML. Is there any functionality in OpenXML 2.8+ that we'd want to use that isn't in 2.7.2?

Pankraty: I thought about bug fixes that possibly could affect us, but you are right, it's better to stick to the lowest suitable version.

ClosedXML/ClosedXML.csproj
igitur: Yes, that's deliberate. The `ClosedXML` project is a library and therefor I could base it on .NET Standard 2.0. But for the other projects, that are _executable_, you need to specify a target that is actually runnable, not just a standard. That's how I understand it. I could be wrong. We definitely want the `ClosedXML` to be `netstandard2.0`. I'll try to see if i can get the other projects to build and execute on `netstandard2.0` instead of `netcoreapp2.0`, but IIRC, I couldn't.

ClosedXML_Sandbox/ClosedXML_Sandbox.csproj
igitur: Fixed

ClosedXML/ClosedXML.csproj
Pankraty: My concerns were about constant names. What directive should one use for code being compiled only in netstandard version? In three projects this would be ```#if _NETSTANDARD2_0_ ...``` while in tests - ```#if _NETCOREAPP2_0_ ...```. I suspected there must be either the same one for all projects or both (for all projects as well).

igitur: Tried building `ClosedXML_Examples` with `netstandard2.0`. It builds fine, but when I try to run it:

```
c:\dev\ClosedXML\ClosedXML\ClosedXML_Examples>dotnet run -f netstandard2.0
Microsoft (R) Build Engine version 15.6.82.30579 for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Restore completed in 39.19 ms for c:\dev\ClosedXML\ClosedXML\ClosedXML\ClosedXML.csproj.
  Restore completed in 39.19 ms for c:\dev\ClosedXML\ClosedXML\ClosedXML_Examples\ClosedXML_Examples.csproj.
Unable to run your project.
Please ensure you have a runnable project type and ensure 'dotnet run' supports this project.
A runnable project should target a runnable TFM (for instance, netcoreapp2.0) and have OutputType 'Exe'.
The current OutputType is 'Exe'.                                  
```

igitur: Oh I see. It's not a requirement for successful compilation, but I agree it would make it simpler for developers to understand. Will align them tomorrow.

igitur: They are now all aligned.
` <DefineConstants>$(DefineConstants);_NETSTANDARD_;_NETSTANDARD2_0_</DefineConstants>`

igitur:
Will merge pending user feedback on the `0.92.0-betaX` releases.

PR #319, opened 2017-05-24
Load and copy images on worksheets.

bsermons:
#### What's this PR do?
Load existing images on worksheets and copy those images when copying worksheets.

Most work is in `ClosedXML/Excel/XLWorkbook_Load.cs`.  Also had to fix up some things while saving to get the excel file to stay consistent when loading/saving.
 
Load up an excel with images and use `Worksheet.CopyTo` and then save and verify the worksheets look similar.

There may be some rounding issues when converting to/from emu for the offsets, etc.  This is why I re-added the excel files for the tests to get the numbers fixed.  

#### Questions:
- Is there a blog post?


igitur:
Thanks for the contribution. The image API will probably change a bit in the future. But in the meantime I'll start reviewing this PR. Will get to it thoroughly next week.

bsermons:
No problem.  I understand about the API changing since it was just added.  I don't need to interact with the api, for my use case, any more than to load the images.

 I added some simple unit tests for sanity checks.  The ImageHandlingTests (ImageAnchors/ImageFormats) that already exists seem to handle the deep excel comparisons that touch all the code in this PR.  Let me know if you think this is good enough or should I add more assertions?

igitur:
Thanks for you contribution. I changed the image API a lot and updated your changes for it in #324 .

Issue #325, opened 2017-05-30
DATE Function bug


DmitriyPotapov:
**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

**What is the current behavior?**
In Excel Worksheet i have a formula:
`=DATE(A2;B2+D2;C2)`
My cell values is: 

- A2 = 2017;
- B2 = 10;
- C2 = 25;
- D2 = 5;

Excel calculate this formula as 25.03.2018. But when I trying to read the value of a cell using a function: **_sheet.Cell("E2").GetString()_** ClosedXML throw an exception:
`Exception Message: Year, Month, and Day parameters describe an un-representable DateTime.
StackTrace:    at System.DateTime.DateToTicks(Int32 year, Int32 month, Int32 day)
   at System.DateTime..ctor(Int32 year, Int32 month, Int32 day)
   at ClosedXML.Excel.CalcEngine.Functions.DateAndTime.Date(List1 p) in C:\Sources\GitLab\WebLeasingMVC\ClosedXML\ClosedXML\Excel\CalcEngine\Functions\DateAndTime.cs:line 94 .........`.
In debug mode in a function **_ClosedXML.Excel.CalcEngine.Functions.DateAndTime.Date(List<Expression> p)_**, I have such runtime values

```c#
        private static object Date(List<Expression> p)
        {
            var year = (int)p[0];     // Runtime value 2017
            var month = (int)p[1]; //Runtime value 15
            var day = (int)p[2];      //Runtime value 25

            //Exception in this line
            return (int)Math.Floor(new DateTime(year, month, day).ToOADate()); 
        }
```
I think this is because the function incorrectly recognizes a part of the formula _**B2+C2**_ in Excel function DATE().

**If the current behavior is a bug, please provide the steps to reproduce and 
if possible a minimal demo of the problem with a sample spreadsheet.**

**What is the expected behavior or new feature?**

**Did this work in previous versions of our tool?  Which versions?**

- [x] I attached a sample spreadsheet.  (You can drag files on to this issue)

**Code to reproduce problem:**
```c#
public void Main()
{
      var workbook = new XLWorkbook(uploadedFile.InputStream);
      IXLWorksheet sheet;
      if (!workbook.TryGetWorksheet("ClosedXmlBug", out sheet))
                        return ;
    
     var bugDate = sheet.Cell("B4").GetString();  //Exception
     Console.WriteLine(bugDate.ToString());
}
```
[...]

PR #326, opened 2017-05-30
Fix Issue #325

DmitriyPotapov:
#### Questions:
- Is there a blog post? - NO
- Does the knowledge base need an update? - NO
- Does this add new (C#) dependencies? - NO

- [x] C# Code Review: @csreviewer
- [x] Test Automation Review: @csreviewer

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Functions/DateAndTime.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/DateAndTime.cs
@@ -90,7 +90,12 @@ namespace ClosedXML.Excel.CalcEngine.Functions
             var year = (int)p[0];
             var month = (int)p[1];
             var day = (int)p[2];
-
+            if (month > 12)
+            {
+                int tempMonth = month - 12;
+                var tempDate = new DateTime(year, 12, day).AddMonths(tempMonth);
+                return (int)Math.Floor(new DateTime(tempDate.Year, tempDate.Month, tempDate.Day).ToOADate());
+            }
             return (int)Math.Floor(new DateTime(year, month, day).ToOADate());
         }
 
--- a/ClosedXML/Excel/CalcEngine/Functions/DateAndTime.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/DateAndTime.cs
@@ -84,14 +84,47 @@ namespace ClosedXML.Excel.CalcEngine.Functions
 
             return businessDays;
         }
-
+        //TODO: fix for excell DATE Formula
         private static object Date(List<Expression> p)
         {
             var year = (int)p[0];
             var month = (int)p[1];
             var day = (int)p[2];
             int tempMonth = 0;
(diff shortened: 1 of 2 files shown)


igitur:
Pleased explain what `TryGetExcelSerialDate` does that was lacking before. Is this related to the issue of months > 12? If not, it should go in a separate PR.  Again, a simple unit test should illustrate the problem.

DmitriyPotapov:
> You have to account for negative months too. I think =DATE(2017,-5,15) works in Excel.

> Also months greater than 24.

In this case, the function (_**private static object Date(List<Expression> p)**_) returns the correct value

```
in Excel Worksheet i have a formula:
=DATE(A2;B2+D2;C2)
My cell values is:

A2 = 2017;
B2 = 10;
C2 = 25;
D2 = -5;
```

In this case, too

``` 
in Excel Worksheet i have a formula:
=DATE(A2;B2+D2;C2)
My cell values is:

A2 = 2017;
B2 = 10;
C2 = 25;
D2 = -5;
```
I think the problem lies in the parser of the excel formulas, not in this functions.


Function **TryGetExcelSerialDate** convert integer value passed from cell formula to DateTime value, i added this function because in my cases, function TryGetValue did not return any value and throw the exception **_Cannot convert cell value to System.DateTime_**.  When there are no formulas in a cell with a date, the value is returned by **_TryGetStringValue_** function....

igitur:
Make D2=-15

DmitriyPotapov:
Yes, in case when D2 = -15 or less, date calculated wrong. I update function for negative month.
```c#
 private static object Date(List<Expression> p)
        {
            var year = (int)p[0];
            var month = (int)p[1];
            var day = (int)p[2];
            int tempMonth = 0;

       
            if (month > 12)
            {
                tempMonth = month - 12;
                var tempDate = new DateTime(year, 12, day).AddMonths(tempMonth);
                return (int)Math.Floor(new DateTime(tempDate.Year, tempDate.Month, tempDate.Day).ToOADate());
            }
            if (month < 0)
            {
                month--;
                tempMonth = month;
                var tempDate = new DateTime(year, 1, day).AddMonths(tempMonth);
                return (int)Math.Floor(new DateTime(tempDate.Year, tempDate.Month, tempDate.Day).ToOADate());
            }
            return (int)Math.Floor(new DateTime(year, month, day).ToOADate());
        }
```

igitur:
and if `D2=50`?

DmitriyPotapov:
> and if `D2=50`?

I tested such values:

1.  `D2 = -50`
2. `D2 = -501`
3. `D2 = 50`

In all cases, the function returned the correct values

DmitriyPotapov:
In case where day or month has sub formula `DATE(A2;B2+F2;C2+D2)` i have same exception. 
In latest commit a added code for handle this

igitur:
#333 is a cleaner fix.

PR #343, opened 2017-06-13
Reworked conditional formatting behavior

b0bi79:
FIXED
~ Shifting Conditional Formats
~ Removing Conditional Formats
~ Copying Conditional Formats

ADDED

Conditional formats compress function
While saving, the conditional formats will be compressed
When cleaning a range, the conditional formatting will be removed


igitur:
Looked at this PR now. It's very difficult to review because it's multiple logical changes in one PR. Sorry, but I have to reject this. Please log one PR for each logical change. For example, one PR to compress conditional formats. One PR for cropping of ranges. One PR for getting a relative range. etc. 

Another request, I think `Consolidate` is a better term rather than `Compress`.

igitur:
I'll assist you to get this merged, but at least just list the different logical changes that you're attempting.

PR #362, opened 2017-06-22
Upgrade OpenXML to v2.7.2

igitur:


peterkeating:
@igitur Any indication of when v0.88 will be released to NuGet?

PR #449, opened 2017-08-17
Insert DataTable without .AsEnumerable()

igitur:
Preparation for .NET Standard


igitur:
Merge this before #315

PR #459, opened 2017-08-30
String concatenation performance improvements

igitur:


Pankraty:
Hi,
I have just tested a few ways of string concatenation, and it turned out that not only String.Concat is 20% slower than StringBuilder.ToString, but also using ```s1+s2+s3+s4``` is even more efficient than a StringBuilder (at least with such short string like cell addresses).
Take a look at this benchmark https://github.com/Pankraty/StringBenchmark.

igitur:
Thanks. I was holding back this PR until I could do my own benchmarks. I really though Concat would be faster. I think I read somewhere that it should be faster.

igitur:
Oh, I think what I read was that Concat is faster than String.Format, but comparable in speed to +s. But you say +s is the best?

Pankraty:
Yeah, I was surprised with this too. I always thought that using StringBuilder would be more efficient and added a benchmark for +s just to see how much worse it would be. 
In theory, if we use +s multiple times, like ```"$"+col+"$"+row.ToString() ```, this causes multiple strings to be allocated and interned: ```"$"```, ```"AB"```, ```"$AB"```, ```"$AB$"```, ```"500"``` and ```"$AB$500"```. StringBuilder is designed to avoid this, and from the point of memory consumption, it can win (although I did not test it yet). From the point of performance, however, if we operate with only a few short strings +s is about 10% faster than StringBuilder.

vbjay:
Stringbuilder isn't a speed optimization. It is an optimization for interning.  Each string created is immutable. So if you are doing this
 ```c#
string x = "The " +"cat " + "is " + "black."
```
Which creates multiple strings
The 
The cat 
The cat is 
The cat is black.
And each individual string. 

With string builder you will have the individual strings but the intermediate strings are not created.   Read up on string interning and strings being immutable.

**Comment by igitur** [comment]
So of the 3 options:
- ` rowPart = fixedRow ? String.Format("R{0}", rowNumber) : String.Format("R[{0}]", rowDiff);`
- `rowPart = fixedRow ? String.Concat('R', rowNumber) : String.Concat("R[", rowDiff, ']');`
- `rowPart = fixedRow ? "R" + rowNumber : "R[" + rowDiff + "]";`

The last one is the best?

**Comment by Pankraty** [comment]
From these 3 options yes, the last one is the best. But bearing in mind that ```rowPart``` will go somewhere else (as it is a 'part') I would consider using StringBuilder for entire expression. It shows more stable and predictable result, only slightly slower for cell addresses but a bit faster for range addresses (and the more profitable the longer expression is).

**Comment by igitur** [comment]
So what is the purpose of `String.Concat()`? When does one use it?

**Comment by igitur** [comment]
> So what is the purpose of String.Concat()? When does one use it?

Answered in https://stackoverflow.com/questions/1787696/why-use-string-concat-in-c

**Comment by Pankraty** [comment]
This answer does not correlate with my benchmark. I wonder which one is wrong :)

**Comment by igitur** [comment]
@Pankraty , Here are my results when I run your benchmark.

```
                   Method |     Mean |     Error |    StdDev | Scaled |
------------------------- |---------:|----------:|----------:|-------:|
           'Create cells' | 175.0 ns | 0.7236 ns | 0.6769 ns |   1.00 |
       String.Format(...) | 590.2 ns | 1.4832 ns | 1.3149 ns |   3.37 |
       String.Concat(...) | 557.8 ns | 1.6373 ns | 1.5315 ns |   3.19 |
 StringBuilder.ToString() | 531.8 ns | 2.0172 ns | 1.6844 ns |   3.04 |
              String.Plus | 550.2 ns | 0.8804 ns | 0.7352 ns |   3.14 |
```

I don't see the big difference between `String.Concat()` and `+` that you see.

**Comment by Pankraty** [comment]
For ranges, the difference is subtle, indeed. For cells, however, it was evident.

**Comment by igitur** [comment]
I don't understand. I just forked your repo and ran it as is. What's the cells vs ranges thing?

**Comment by Pankraty** [comment]
There are two benchmarks, and the results from the first one are somewhere in the middle of the output stream. You may either scroll the result or omit the second benchmark.

**Comment by igitur** [comment]
Oooh, I missed that.

# Cells:
```
                     Method |      Mean |     Error |    StdDev | Scaled | ScaledSD |
--------------------------- |----------:|----------:|----------:|-------:|---------:|
              'Create cell' |  91.26 ns | 1.6443 ns | 1.3731 ns |   1.00 |     0.00 |
         String.Format(...) | 311.19 ns | 1.3937 ns | 1.1638 ns |   3.41 |     0.05 |
         String.Concat(...) | 285.33 ns | 1.0348 ns | 0.8641 ns |   3.13 |     0.04 |
   StringBuilder.ToString() | 264.23 ns | 0.8291 ns | 0.6923 ns |   2.90 |     0.04 |
 "$"+col+"$"+row.ToString() | 236.96 ns | 1.1059 ns | 0.9234 ns |   2.60 |     0.04 |
```

# Ranges:
```


                   Method |     Mean |     Error |    StdDev | Scaled | ScaledSD |
------------------------- |---------:|----------:|----------:|-------:|---------:|
           'Create cells' | 181.7 ns |  1.706 ns |  1.512 ns |   1.00 |     0.00 |
       String.Format(...) | 633.3 ns | 15.092 ns | 18.534 ns |   3.49 |     0.10 |
       String.Concat(...) | 586.1 ns |  6.010 ns |  5.327 ns |   3.23 |     0.04 |
 StringBuilder.ToString() | 564.7 ns | 11.214 ns |  9.941 ns |   3.11 |     0.06 |
              String.Plus | 587.3 ns | 19.747 ns | 16.489 ns |   3.23 |     0.09 |
```

Pankraty:
So, the first row "Create cell" stands for the logic that is common for the rest test. That means the actual difference between ```String.Concat``` and ```A+B+C``` is ```(285.33 - 91.26) - (236.96 - 91.26) = 48.37 ns```. It is not huge, of course, but statistically significant.

vbjay:
A.  To make those tests valid, have to run a new process for each test. 
B.  https://coders-corner.net/2014/08/20/concatenate-strings-in-c-operator-vs-string-concat-vs-stringbuilder/

igitur:
Well, the majority of this PR is to replace `String.Format` with `String.Concat` and simplify some logic too. Any objects if I merge this, and then we can debate `Concat` vs `+` later?

Pankraty:
Yeah, you are right, it is not the propper time for _micro_-optimizations, while there are places begging for _mega_-optimizations 8) It was just interesting to experiment a little with such basic things as string concatenation. Sorry, if I delayed you.

igitur:
@Pankraty No problem. This PR has been lying idle for many months.

PR #460, opened 2017-08-31
use slightly less memory

emilyytlin:
Remove _trimmedAddress and _columnLetter members from XLAddress
https://closedxml.codeplex.com/discussions/641663


igitur:
This would use slightly less memory, but would probably have a performance impact as the `TrimmedAddress` and `ColumnLetter` would need to be recalculated every time the method is called. Do you have any quantification of the memory saved vs performance impact?

igitur:
Also, the biggest cause of the memory usage is that each cell has its own instance of the `XLStyle` class, with all the child objects in the graph.

PR #477, opened 2017-09-09
Set sheet state to null if it's visible.

### LINKED ISSUE #476: Unhide worksheets not working

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.88.0 (but was already present long before)

Hiding a sheet works fine - that is OK
UNHIDING a sheet that was hidden and saved doesn't work. Nothing happens
I assume XLWorkbook_Save.cs needs to do this command always (not only when hiding):
sheet.State = xlSheet.Visibility.ToOpenXml();

Unhiding should show the sheet

Not that I'm aware of. It was already failing in V0.60.0.0

## Reproducibility

```c#
public void Main()
{
            // create a file with a visible sheet and a hidden sheet
            string path = @"c:\temp\testonline.xlsx";
            var wb = new XLWorkbook();
            wb.Worksheets.Add("Visiblesheet");
            wb.Worksheets.Add("hiddensheet").Hide();
            wb.SaveAs(path);

            // unhide the hidden sheet
            XLWorkbook w = new XLWorkbook(path);
            IXLWorksheet s = w.Worksheets.Skip(1).First();
            if (s.Visibility == XLWorksheetVisibility.Hidden)
            {
                s.Unhide();
                // OR (neither work and Unhide is only a shortcut to change the visibility)
                //s.Visibility = XLWorksheetVisibility.Visible;
            }
            w.Save();

            // validate 
            w = new XLWorkbook(path);
            s = w.Worksheets.Skip(1).First();
            if (s.Visibility == XLWorksheetVisibility.Hidden)
            {
                Console.WriteLine("bug reproduced");
            }

}
```

igitur:
Fixes #476

### DIFF
--- a/ClosedXML/Excel/XLWorkbook_Save.cs
+++ b/ClosedXML/Excel/XLWorkbook_Save.cs
@@ -672,6 +672,8 @@ namespace ClosedXML.Excel
                     var xlSheet = Worksheet(sheet.Name);
                     if (xlSheet.Visibility != XLWorksheetVisibility.Visible)
                         sheet.State = xlSheet.Visibility.ToOpenXml();
+                    else
+                        sheet.State = null;
 
                     if (foundVisible) continue;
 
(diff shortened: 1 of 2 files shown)


peetn:
Seems to do the trick! Thanks

PR #479, opened 2017-09-12
Adds a reference point for number formats available.

### LINKED ISSUE #478: Enum or static number/date format ids and/or codes

- [ ] Bug
- [x] Feature

I need to go to https://github.com/closedxml/closedxml/wiki/NumberFormatId-Lookup-Table to check which id corresponds to what format code. I was just wondering if you could you make this enum or some static instances with proper XML summary docs? Then I'd simply navigate to declarations and remain within an IDE. Moreover, your list is not complete as https://msdn.microsoft.com/en-us/library/documentformat.openxml.spreadsheet.numberingformat.aspx provides more options.

```csharp
range.Style.NumberFormat.NumberFormatId = NumberFormat._1; // or
range.Style.NumberFormat.NumberFormatId = NumberFormatIds._1 // or 
range.Style.NumberFormat.NumberFormatId = NumberFormats.ZeroZeroDotZero // etc.
```

wachulski:
#### What's this PR do?
Adds a reference point for number formats available.

No need to lookup ids/codes on the Internet. Closes #478.

ClosedXML/Excel/Style/XLFormatIds.cs directly

n/a

n/a

#### Questions:
- Is there a blog post? NO
- Does the knowledge base need an update? YES, in particular this https://github.com/closedxml/closedxml/wiki/NumberFormatId-Lookup-Table
- Does this add new (C#) dependencies? NO

### DIFF
--- a/ClosedXML/ClosedXML.csproj
+++ b/ClosedXML/ClosedXML.csproj
@@ -83,6 +83,7 @@
     <Compile Include="Excel\Misc\XLCallbackAction.cs" />
     <Compile Include="Excel\Misc\XLReentrantEnumerableSet.cs" />
     <Compile Include="Excel\SaveOptions.cs" />
+    <Compile Include="Excel\Style\XLPredefinedFormat.cs" />
     <Compile Include="Excel\Tables\XLTableTheme.cs" />
     <Compile Include="Excel\XLConstants.cs" />
     <Compile Include="Excel\XLWorkbook_ImageHandling.cs" />
--- a/ClosedXML/Excel/Style/XLPredefinedFormat.cs
+++ b/ClosedXML/Excel/Style/XLPredefinedFormat.cs
@@ -1,0 +1,165 @@
+﻿namespace ClosedXML.Excel
+{
+    /// <summary>
+    /// Reference point of date/number formats available.
+    /// See more at: https://msdn.microsoft.com/en-us/library/documentformat.openxml.spreadsheet.numberingformat.aspx
+    /// </summary>
+    public static class XLPredefinedFormat
+    {
+        public enum Number
+        {
+            /// <summary>
+            /// General
+            /// </summary>
+            General = 0,
+
+            /// <summary>
+            /// 0
+            /// </summary>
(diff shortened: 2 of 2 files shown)


igitur:
Isn't it better to use an `enum` for this?

wachulski:
Well. With enum you need to cast to Int32 explicitly. But yes, this has more compile time benefits provided we used the enum as a NumberFormatId type instead of Int32. You to decide. I will still have to correct this as the class itself should be made static. So... ?

igitur:
Oh, I see now you mentioned an `enum` approach in #478 . Yeah, I think an `enum` is better here. But let's consider a better naming convention rather than `F01`, `F02`, ... etc.

wachulski:
Ok, I proposed more semantic approach. Categorized into number and datetime spaces. IMO should be more readable and user friendly to justify the change.

ClosedXML/Excel/Style/XLFormatIds.cs
```diff
@@ -0,0 +1,167 @@
+﻿using System;
+
+namespace ClosedXML.Excel
+{
+    /// <summary>
+    /// Reference point of date/number formats available.
+    /// See more at: https://msdn.microsoft.com/en-us/library/documentformat.openxml.spreadsheet.numberingformat.aspx
+    /// </summary>
+    public static class XLFormat
```
igitur: I think `XLPredefinedFormat` is clearer.

ClosedXML/ClosedXML.csproj
```diff
@@ -83,6 +83,7 @@
     <Compile Include="Excel\Misc\XLCallbackAction.cs" />
     <Compile Include="Excel\Misc\XLReentrantEnumerableSet.cs" />
     <Compile Include="Excel\SaveOptions.cs" />
+    <Compile Include="Excel\Style\XLFormatIds.cs" />
```
igitur: Change to `XLPredefinedFormat.cs` in line with comment below.

wachulski:
Ok, here you go.

PR #485, opened 2017-09-16
Added Crop and Relative methods for XLRangeBase

b0bi79:
Added Crop and Relative methods for XLRangeBase

### DIFF
--- a/ClosedXML/Excel/Cells/IXLCell.cs
+++ b/ClosedXML/Excel/Cells/IXLCell.cs
@@ -32,7 +32,7 @@ namespace ClosedXML.Excel
 
         /// <summary>Gets this cell's address, relative to the worksheet.</summary>
         /// <value>The cell's address.</value>
-        IXLAddress Address { get;  }
+        IXLAddress Address { get; }
 
         /// <summary>
         /// Gets or sets the type of this cell's data.
@@ -117,8 +117,6 @@ namespace ClosedXML.Excel
 
         Boolean TryGetValue<T>(out T value);
 
-
-
         Boolean HasHyperlink { get; }
 
         /// <summary>
@@ -154,7 +152,6 @@ namespace ClosedXML.Excel
         /// </summary>
         IXLRange AsRange();
 
-
         /// <summary>
         /// Gets or sets the cell's style.
         /// </summary>
@@ -181,6 +178,13 @@ namespace ClosedXML.Excel
         /// <param name="tranpose">if set to <c>true</c> the data will be transposed before inserting.</param>
         /// <returns></returns>
         IXLRange InsertData(IEnumerable data, Boolean tranpose);
+
+        /// <summary>
+        /// Inserts the data of a data table.
+        /// </summary>
+        /// <param name="dataTable">The data table.</param>
+        /// <returns>The range occupied by the inserted data</returns>
+        IXLRange InsertData(DataTable dataTable);
 
         /// <summary>
         /// Inserts the IEnumerable data elements as a table and returns it.
@@ -217,7 +221,6 @@ namespace ClosedXML.Excel
         /// <para>if set to <c>false</c> the table will be created in memory.</para>
         /// </param>
         IXLTable InsertTable<T>(IEnumerable<T> data, String tableName, Boolean createTable);
-
 
         /// <summary>
         /// Inserts the DataTable data elements as a table and returns it.
@@ -262,12 +265,15 @@ namespace ClosedXML.Excel
 
         IXLDataValidation DataValidation { get; }
         IXLDataValidation NewDataValidation { get; }
+
         IXLDataValidation SetDataValidation();
 
-
         IXLCells InsertCellsAbove(int numberOfRows);
+
         IXLCells InsertCellsBelow(int numberOfRows);
+
         IXLCells InsertCellsAfter(int numberOfColumns);
+
         IXLCells InsertCellsBefore(int numberOfColumns);
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


ClosedXML/Excel/Ranges/XLRangeBase.cs
```diff
@@ -2067,5 +2067,38 @@ public void Select()
         {
             Worksheet.SelectedRanges.Add(AsRange());
         }
+
+        /// <summary>
+        /// Get cropped range.
+        /// </summary>
+        public IXLRangeBase Crop(IXLRangeBase crop)
+        {
```
igitur: What happens if the 2 ranges don't overlap? Do we need error handlers for that case?

ClosedXML_Examples/app.config
```diff
@@ -1,6 +1,6 @@
-<?xml version="1.0"?>
-<configuration>
-  <startup>
-    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5.2"/>
-  </startup>
-</configuration>
+<?xml version="1.0"?>
```
igitur: Looks like whitespace changes in this file. Please undo this.

ClosedXML/Excel/Ranges/XLRangeBase.cs
```diff
@@ -2067,5 +2067,38 @@ public void Select()
...
+            }
+        }
+
+        /// <summary>
+        /// Get range relative to another range.
+        /// </summary>
+        /// <param name="baseRange">Coordinate system. Coordinates are calculated relative to this range.</param>
+        /// <param name="targetBase"></param>
+        public IXLRange Relative(IXLRangeBase baseRange, IXLRangeBase targetBase)
+        {
```
igitur: Does `baseRange` and `targetRange` have to be the same size? If so, please add error handlers. If not, what happens if it leads to negative numbers below?

igitur: Also, I don't quite the use of this function. It looks like something that will be required in really rare cases. Why would this need to be included in `ClosedXML` instead of leaving it to the user to calculate? Please explain.

ClosedXML_Tests/Excel/Ranges/XLRangeBaseTests.cs
```diff
@@ -170,6 +170,33 @@ public void WsNamedRangesOneString()
             Assert.AreEqual("$A$3:$A$3", namedRanges.Last().RangeAddress.ToStringFixed());
         }
 
+        [Test]
+        public void RelativeTest()
+        {
+            var wb = new XLWorkbook();
```
igitur: Change the `using (var wb = new XLWorkbook())` style.

ClosedXML_Tests/Excel/Ranges/XLRangeBaseTests.cs
```diff
@@ -170,6 +170,33 @@ public void WsNamedRangesOneString()
...
+
+            var assert = range.Relative(baseRange, targetRange);
+            Assert.AreEqual("D4", assert.RangeAddress.FirstAddress.ToStringRelative());
+            Assert.AreEqual("E5", assert.RangeAddress.LastAddress.ToStringRelative());
+        }
+
+        [Test]
+        public void CropTest()
+        {
+            var wb = new XLWorkbook();
```
igitur: Change the `using (var wb = new XLWorkbook())` style.

ClosedXML_Tests/Excel/Ranges/XLRangeBaseTests.cs
```diff
@@ -170,6 +170,33 @@ public void WsNamedRangesOneString()
...
+            IXLWorksheet ws = wb.Worksheets.Add("Sheet");
+            var range = ws.Range("B3:D8");
+            var crop = ws.Range("A5:E6");
+            var assert = range.Crop(crop);
+
+            Assert.AreEqual("B5", assert.RangeAddress.FirstAddress.ToStringRelative());
+            Assert.AreEqual("D6", assert.RangeAddress.LastAddress.ToStringRelative());
+        }
+
         //[Test]
```
igitur: Also add tests for the error handlers after you've added them.

ClosedXML/Excel/Ranges/IXLRangeBase.cs
```diff
@@ -267,5 +267,17 @@ public interface IXLRangeBase : IDisposable
         IXLConditionalFormat AddConditionalFormat();
 
         void Select();
+
+        /// <summary>
+        /// Get cropped range.
+        /// </summary>
+        IXLRangeBase Crop(IXLRangeBase crop);
```
igitur: I think `IXLRangeBase Intersection(IXLRangeBase range);` is a better name for this method, or do I not understand its intent?

ClosedXML/Excel/Ranges/XLRangeBase.cs
b0bi79: I often use this function in my library, so I thought that it might be necessary for someone else. If you do not think so, then I will delete it.

igitur:
Why closing? I think these are good additions.

igitur:
@b0bi79  ? Was it because I requested changes?

b0bi79:
I do not know how to roll back the commit to the git, so it's easier for me to create a new PR. In the new PR, I removed the Relative method.

vbjay:
You can roll back by using revert. https://git-scm.com/docs/git-revert

vbjay:
https://gitextensions.github.io/ is a git ui.

http://www.vogella.com/tutorials/Git/article.html. Good git tutorial

Issue #217, opened 2017-02-28
Conditional formatting to stop when true


igitur:
http://stackoverflow.com/questions/42008831/closedxml-c-sharp-how-to-get-conditional-formating-for-excel-to-stop-when-true

NickNack2020:
@igitur I am currently taking a stab at this.

igitur:
Waiting for @NickNack2020  to clean up the conflicts in his PR.

igitur:
https://www.nuget.org/packages/ClosedXML/0.90.0-beta2 is released and should include the fix for this issue.

igitur:
I'm relooking at this. Am I right that there's no way the user can *get* the `StopIfTrue` value? The property is internal.

PR #501, opened 2017-09-21
Targets issue #217 - Add Support for Conditional Format's StopIfTrue

NickNack2020:
Targets Issue #217 - Add support for StopIfTrue for conditional Formats.


ClosedXML/Excel/ConditionalFormats/Save/XLCFContainsConverter.cs
```diff
@@ -11,8 +8,11 @@ internal class XLCFContainsConverter : IXLCFConverter
         public ConditionalFormattingRule Convert(IXLConditionalFormat cf, int priority, XLWorkbook.SaveContext context)
         {
             String val = cf.Values[1].Value;
-            var conditionalFormattingRule = new ConditionalFormattingRule { FormatId = (UInt32)context.DifferentialFormats[cf.Style], Operator = ConditionalFormattingOperatorValues.ContainsText, Text = val, Type = cf.ConditionalFormatType.ToOpenXml(), Priority = priority };
-
+            var conditionalFormattingRule = XLCFBaseConverter.Convert(cf, priority);
```
igitur: In many places, you changed the style of initialisation, like here. Why did you do that? I don't think it's relevant to this change.

ClosedXML/Excel/ConditionalFormats/Save/XLCFConverters.cs
```diff
@@ -11,23 +11,25 @@ internal class XLCFConverters
...
-            Converters.Add(XLConditionalFormatType.ContainsText, new XLCFContainsConverter());
-            Converters.Add(XLConditionalFormatType.NotContainsText, new XLCFNotContainsConverter());
-            Converters.Add(XLConditionalFormatType.CellIs, new XLCFCellIsConverter());
-            Converters.Add(XLConditionalFormatType.IsUnique, new XLCFUniqueConverter());
-            Converters.Add(XLConditionalFormatType.IsDuplicate, new XLCFUniqueConverter());
-            Converters.Add(XLConditionalFormatType.Expression, new XLCFCellIsConverter());
-            Converters.Add(XLConditionalFormatType.Top10, new XLCFTopConverter());
-            Converters.Add(XLConditionalFormatType.DataBar, new XLCFDataBarConverter());
-            Converters.Add(XLConditionalFormatType.IconSet, new XLCFIconSetConverter());
+            Converters = new Dictionary<XLConditionalFormatType, IXLCFConverter>
```
igitur: Please keep the existing style to ensure the diff of the PR is minimal.

ClosedXML/Excel/ConditionalFormats/XLConditionalFormat.cs
```diff
@@ -39,6 +39,9 @@ public XLConditionalFormat(XLConditionalFormat conditionalFormat)
             ReverseIconOrder = conditionalFormat.ReverseIconOrder;
             ShowIconOnly = conditionalFormat.ShowIconOnly;
             ShowBarOnly = conditionalFormat.ShowBarOnly;
+            StopIfTrueInternal = OpenXmlHelper.GetBooleanValueAsBool(conditionalFormat.StopIfTrueInternal, true);
```
igitur: This doesn't make sense. OpenXML types, e.g. `BooleanValue` should not be used here.

ClosedXML/Excel/ConditionalFormats/XLConditionalFormat.cs
```diff
@@ -110,6 +121,7 @@ public void CopyFrom(IXLConditionalFormat other)
             ReverseIconOrder = other.ReverseIconOrder;
             ShowIconOnly = other.ShowIconOnly;
             ShowBarOnly = other.ShowBarOnly;
+            StopIfTrueInternal = OpenXmlHelper.GetBooleanValueAsBool(((XLConditionalFormat)other).StopIfTrueInternal, true);
```
igitur: This doesn't make sense. OpenXML types, e.g. BooleanValue should not be used here.

ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -1838,6 +1838,9 @@ private void LoadConditionalFormatting(ConditionalFormatting conditionalFormatti
                 foreach (var fr in conditionalFormatting.Elements<ConditionalFormattingRule>())
                 {
                     var conditionalFormat = new XLConditionalFormat(ws.Range(sor.Value));
+
+                    conditionalFormat.StopIfTrueInternal = OpenXmlHelper.GetBooleanValueAsBool(fr.StopIfTrue, false);
```
igitur: This doesn't make sense. OpenXML types, e.g. BooleanValue should not be used here.

ClosedXML/Utils/OpenXmlHelper.cs
```diff
@@ -0,0 +1,17 @@
+﻿using DocumentFormat.OpenXml;
+
+namespace ClosedXML.Utils
+{
+    internal static class OpenXmlHelper
```
igitur: Converting between OpenXML and ClosedXML variables should be limited to loading and saving. I think they should remain private members of the `XLWorkbook` class.

ClosedXML/Excel/ConditionalFormats/IXLConditionalFormat.cs
```diff
@@ -119,6 +119,7 @@ public interface IXLConditionalFormat
         Boolean Bottom { get;  }
         Boolean Percent { get; }
 
+        IXLConditionalFormat StopIfTrue(bool value);
```
igitur: I think you should add another overload method:
```c#
IXLConditionalFormat StopIfTrue();
```
which obviously defaults to `true`

ClosedXML/Excel/ConditionalFormats/Save/XLCFContainsConverter.cs
NickNack2020: I did it because then i only have to set StopIfTrue in one place from the Cf that we read in, which would be in the base constructor. I believe there are other properties that are currently unsupported, and should someone ever have to implement them, they would only have to make the code change in one spot: http://www.datypic.com/sc/ooxml/e-ssml_cfRule-1.html

igitur: Ah ok. I see now. It's fine.

ClosedXML/Excel/ConditionalFormats/XLConditionalFormat.cs
NickNack2020: good catch, sorry about that. Bad copy paste.

ClosedXML/Utils/OpenXmlHelper.cs
NickNack2020: I also need this in the XCLFBaseConverter. I tend to follow DRY principles and see no problem with extracting this method into a utility class. Would you rather a different approach here?

ClosedXML/Excel/XLWorkbook_Load.cs
NickNack2020: Can you expand on this? I think it does make sense. We are dealing with a Conditional Format rule from the OOXML lib. Which is returning a BooleanValue, instead of a bool.

image

igitur: Sorry, this time it was my copy and pasting ;-) This change is fine.

ClosedXML/Utils/OpenXmlHelper.cs
igitur: Yes, I agree now.

igitur:
Fixes #217

igitur:
Thanks for the contribution. Sorry about the confusion. Lesson learnt: don't try to review non-trivial changes in the browser.

PR #506, opened 2017-09-23
Improved performance of some methods.

b0bi79:
Improved performance of some methods.


b0bi79:
Depends on the content of the document. If the document has many merged regions, it can improve 100 times.

igitur:
Do you perhaps have a test case that I can use to reproduce the application hanging? If I can reproduce I can find other problems.

Issue #165, opened 2017-01-21
Dispose hell


b0bi79:
Now the use of the library is very difficult by calls Dispose. This is a real nightmare. Here is a piece of my code:
            using (var grRows = Sheet.Rows(firstRow.RowNumber(), lastRow.RowNumber()))
                foreach (var grRow in grRows)
                    grRow.OutlineLevel++;
            int level;
            using (var row = Sheet.Row(firstRow.RowNumber()))
                level = row.OutlineLevel;
            using (var row = Sheet.Row(summRow.RowNumber()))
                row.OutlineLevel = level - 1;

            if (_pageBreaks)
                Sheet.PageSetup.AddHorizontalPageBreak(summRow.RowNumber());
            firstRow.Dispose();
            lastRow.Dispose();

I think it would be better to make a subscription to Shift events (SubscribeToShiftedRows, SubscribeToShiftedColumns) make on weak references. What are the arguments against?

b0bi79:
Or it may be better to add a simple way to get a non-subscription area. It looks like:
```
sheet.SuppressSubscription();
int level = sheet.Row(firstRow.RowNumber()).OutlineLevel;
sheet.Row(summRow.RowNumber()).OutlineLevel = level - 1;
sheet.ResumeSubscription();
```

PR #507, opened 2017-09-23
Disposing of operating ranges.

b0bi79:
Added Dispose call where it was missed. For large documents, this is critically important. Otherwise, callback lists grow by tens and hundreds of thousands, which leads to the freezing of the application.

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -1277,7 +1277,10 @@ namespace ClosedXML.Excel
             get
             {
                 using (var asRange = AsRange())
-                    return Worksheet.DataValidations.Any(dv => dv.IsDirty() && dv.Ranges.Contains(asRange));
+                    return Worksheet.DataValidations.Any(dv =>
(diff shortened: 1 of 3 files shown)


igitur:
Does this fix #165 ?

igitur:
PS: The recommended way to do changes is to branch on your side first (`git checkout -b newbranchname upstream/develop`), make the changes, push to your local repo (`git push origin newbranchname`) and then creating the PR. 

The way you did it now makes it difficult for you to work on a separate feature until this PR is merged into the `develop` branch.

igitur:
I think there are many more places where the `using` statement has to be used for ranges.

b0bi79:
No, this is not a fix for #165 .
Yes, there are certainly more places where the `using` statement has to be used. If you want, I'll find all these places and correct them. Now I have found the most frequently used places.

Thanks for the help with the git.

igitur:
Please see if #521 improves performance. Also, if you can attach an example of the large spreadsheet and the operations you do on it, it would be nice. Then I can test the performance myself.

PR #521, opened 2017-10-02
Implement WeakEvent pattern to reduce memory usage

igitur:
Implemented the [WeakEvent](http://pragmateek.com/the-net-weak-event-pattern-in-c/) pattern, which should allow objects to be garbage collected even though they still have subscribed events.


igitur:
@b0bi79 Would you mind testing the build at https://ci.appveyor.com/project/Pyropace/closedxml/build/0.8.600/artifacts with your large spreadsheets. 

I have implemented the [WeakEvent](http://pragmateek.com/the-net-weak-event-pattern-in-c/) pattern, which should allow objects to be garbage collected even though they still have subscribed events. 

The disadvantage is that this works only in .NET 4.5+ (not .NET 4.0 or .NET Standard 2.0)

igitur:
@b0bi79 Can you test this? Please. New build at https://ci.appveyor.com/project/Pyropace/closedxml/build/artifacts

igitur:
@ghronkrepps Please test the build at https://ci.appveyor.com/project/Pyropace/closedxml/build/0.8.672/artifacts to see if it solves your memory problems.

@b0bi79 I rely on your feedback too.

b0bi79:
@igitur I tested for performance. With this assembly, it runs 30-40% faster. Also, memory consumption was reduced by 20%.
But this is one particular test. Under different conditions, there will be other results.

DDeas:
I have tested both the builds in this thread so far for memory consumption during large file saving to stream. My project is a website.

I force garbage collection at the following points:
- Just before workbook.SaveAs()
- After workbook.SaveAs() (workbook now disposed)
- After returning file to client

Using the same dataset, to produce a file 17mb (58,439 rows by 38 cols):

- **0.8.672**
-- Peaked at 5.1gb RAM.
-- Held on to 3.8gb

- **0.8.683**
-- Peaked at 5.0gb RAM.
-- RAM dropped to 300mb after file returned

- **Current live release: 0.89.0.0**
-- Peaked at 4.7gb RAM
-- Dropped to 700mb after file returned
-- CPU seemed to average a little lower

Hope this is in someway useful to you guys.

igitur:
Continued in #535

Issue #52, opened 2016-09-28
Implement all Excel functions


igitur:

pichardoSoftware:
Any way I can help you with this?

igitur:
Definitely. We appreciate any help we can get. 

If you want to ease in, you can start with some of the simpler functions. I think the Logical and Information groups of functions in Excel would be the easiest to implement.
![image](https://cloud.githubusercontent.com/assets/145854/19437042/ce9b35e8-9473-11e6-9ffc-d65e01c99599.png)

There is already a `Logical` class, which you can start extending. The `Is` class should probably be renamed to `Information` and extended. You're also welcome to try any of the other functions. Please group them according to the grouping that Excel uses. Let me know if you need additional info.

pichardoSoftware:
Excellent

igitur:
To implement something like `ISERR()`, we'll have to implement ClosedXML-specific exceptions which can be caught in the `ISERR` implementation. I'm almost done with those exceptions. You can have a look at https://github.com/igitur/ClosedXML/tree/improve-calcengine-exceptions if you're interested.

pichardoSoftware:
Ok, I will start extending that class and check the exceptions.

pichardoSoftware:
I've 2 questions:
-----
 - How should I manage errors, like #value, #error, etc?
  - It doesn't seem logical to implement functions like `IsErr()` and `IsError()` on a programmatically generated spreadsheet, Maybe on parsing an existing sheet.
 - There are some functions on the Information section that accept various or one argument, and some that only accept one, so do you think is a good idea to keep using `List<Expression>` or should we use variable parameters `params Expression[]`.

What do you think about this?

pichardoSoftware:
How are you doing with the exceptions?

igitur:
Sorry. Busy week. I'll try to get round to it.

pichardoSoftware:
I understan @igitur, I'll continue with other classes then, Greetings

jahav:
I am closing this issue since it is not reasonable ticket. It's something like "learn English" instead of "Learn conjugation of 5 irregular verbs". 

At the moment, we have [a table](https://closedxml.readthedocs.io/en/latest/features/functions.html ) of all functions and state of implementation and if dev feels like implementing one, the rest will be done per reported issues.

Easy pickings are done, a significant portion of functions are not sufficiently compliant with Excel, need verification (by fuzzer hopefully) against Excel.

PR #523, opened 2017-10-02
#52: implement CSCH excel function...

jongleur1983:
#### What's this PR do?
This PR implements the CSCH function provided by excel that calculates the hyperbolic cosecant of an angle.
See https://support.office.com/en-us/article/CSCH-function-f58f2c22-eb75-4dd6-84f4-a503527f8eeb?ui=en-US&rs=en-US&ad=US for Microsofts documentation.

It's a simple PR of the function and a few tests checking the correct calculation.

To be honest: no idea, I followed PR #74 to get how it works. I there's anything I could do better, please tell me.
#### Questions:
- Is there a blog post?
no.
don't think so: As far as I see there's no documentation about individual implemented functions in the wiki.
no.

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Functions/MathTrig.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/MathTrig.cs
@@ -25,6 +25,7 @@ namespace ClosedXML.Excel.CalcEngine
             ce.RegisterFunction("COMBIN", 2, Combin);
             ce.RegisterFunction("COS", 1, Cos);
             ce.RegisterFunction("COSH", 1, Cosh);
+            ce.RegisterFunction("CSCH", 1, Csch);
             ce.RegisterFunction("DEGREES", 1, Degrees);
             ce.RegisterFunction("EVEN", 1, Even);
             ce.RegisterFunction("EXP", 1, Exp);
@@ -113,6 +114,15 @@ namespace ClosedXML.Excel.CalcEngine
         private static object Cosh(List<Expression> p)
         {
             return Math.Cosh(p[0]);
+        }
+
+        private static object Csch(List<Expression> p)
+        {
+            if (p[0].Evaluate().Equals(0))
+            {
+                return ErrorExpression.ExpressionErrorType.DivisionByZero;
+            }
+            return 1 / Math.Sinh(p[0]);
         }
 
         private static object Exp(List<Expression> p)
--- a/ClosedXML/Excel/CalcEngine/Functions/MathTrig.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/MathTrig.cs
@@ -118,7 +118,7 @@ namespace ClosedXML.Excel.CalcEngine
 
         private static object Csch(List<Expression> p)
         {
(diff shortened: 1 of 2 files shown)


ClosedXML/Excel/CalcEngine/Functions/MathTrig.cs
```diff
@@ -118,7 +118,7 @@ private static object Cosh(List<Expression> p)
 
         private static object Csch(List<Expression> p)
         {
-            if (p[0].Evaluate().Equals(0))
+            if (p[0].Evaluate().Equals(0.0))
```
igitur: For floating point comparison, use `if (Math.Abs(p[0].Evaluate()) < Double.Epsilon)`

ClosedXML_Tests/Excel/CalcEngine/MathTrigTests.cs
```diff
@@ -139,5 +140,37 @@ public void SumProduct()
                 Assert.Throws<NoValueAvailableException>(() => ws.Evaluate("SUMPRODUCT(A1:A10, B1:B5)"));
             }
         }
+
+        [TestCase(1, 0.850918128)]
```
igitur: For the tests, please follow the same pattern as the testing of other functions, where possible, i.e. try to use ` XLWorkbook.EvaluateExpr`

ClosedXML/Excel/CalcEngine/Functions/MathTrig.cs
```diff
@@ -115,6 +116,15 @@ private static object Cosh(List<Expression> p)
             return Math.Cosh(p[0]);
         }
 
+        private static object Csch(List<Expression> p)
+        {
+            if (p[0].Evaluate().Equals(0.0))
+            {
+                return ErrorExpression.ExpressionErrorType.DivisionByZero;
```
igitur: Should be `throw new DivisionByZeroException();`

igitur:
Thanks for the contribution. Please have a look at the requested changes. They're minor. 

Also, if you want to have a stab at the other trigonometry functions, that would be great.

ClosedXML/Excel/CalcEngine/Functions/MathTrig.cs
jongleur1983: I agree in principle, but for 0 this should be save. If you want I'll change it of course.

ClosedXML_Tests/Excel/CalcEngine/MathTrigTests.cs
jongleur1983: yes. will do. Took the wrongest possible example for reference I guess, currently implementing DECIMAL with the other pattern already ;)

ClosedXML/Excel/CalcEngine/Functions/MathTrig.cs
igitur: You can't predict how the floating point rounding will work, e.g. `=CSCH(8-64/8)` . Please change it.

igitur:
Please resolve the conflicts. Your other PR made changes to same file.

PR #531, opened 2017-10-05
implement COT function (co-tangens) and corresponding tests

jongleur1983:
#### What's this PR do?
implements the Excel function COT, which is the co tangens.
Adds tests for it.
as before: PR is quite small, so should be straight forward.
If it feels necessary, compare what Excel does on cell content =COT(x) for any X and what's the result of the function implemented.
Mathmatically there are special cases for any multiple of π as for these values the result is not defined (these are poles with a local limit of +infinity or -infinity, depending on the direction one comes from).
Therefore these are in fact special cases but I have no idea how to cover that by unit tests as PI is impossible to model correct.
Exception: 0 = 0*π is one of these special cases, that case is covered by test.


igitur:
Pull in latest develop branch please.

PR #534, opened 2017-10-07
Fix for Clear cell.

b0bi79:
If the cell had a DateTime value, Clear method does not reset the cell type. This leads to the fact that the cell remains unchanged when the document is saved. The SetCellValue method changes the value of the cells with the DateTime and Number types only if they are not empty. Therefore, if we clear a cell, we must also change the data type.
File XLWorkbook_Save.cs -> method SetCellValue, row 5027:
```
            if (dataType == XLCellValues.DateTime || dataType == XLCellValues.Number)
            {
                if (!String.IsNullOrWhiteSpace(xlCell.InnerText))
                {
                    ....
                    openXmlCell.CellValue = cellValue;
                }
            }
```

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -1008,6 +1008,7 @@ namespace ClosedXML.Excel
                 {
                     Hyperlink = null;
                     _richText = null;
+                    _dataType = XLCellValues.Text;
                     //_comment = null;
                     _cellValue = String.Empty;
                     FormulaA1 = String.Empty;


igitur:
What happens in Excel if you clear a cell with a date? Does it remember the date type? For example if you enter 45000, does it show as a date?

igitur:
I've had a look at this, and I think we need to expand the `XLClearOptions` enum to give the user more control about what he wants to clear. Look at my implementation at https://github.com/igitur/ClosedXML/tree/more-clearing-options

It will be a breaking change, but it changes the options to a flag-based enum. In your case, you will have to clear the cells by passing the `XLClearOptions.DataType` too.

```c#
[Flags]
public enum XLClearOptions
{
    Contents = 1 << 0,
    DataType = 1 << 1,
    NormalFormats = 1 << 2,
    ConditionalFormats = 1 << 3,
    Comments = 1 << 4,
    DataValidation = 1 << 5,

    AllFormats = NormalFormats | ConditionalFormats,
    All = Contents | DataType | NormalFormats | ConditionalFormats | Comments | DataValidation
}
```

Do you agree that we should rather follow this approach?

In fact, I think we should do something similar for all the `Copy` methods.

b0bi79:
I do not really understand why I need to save the type of cell when cleaning? If I give the cell a new value, then the cell must accept the type of this value. And why should I set the date through the number?
But okay, I'm still not very good at this. However, if a person cleans a cell without clearing the type, it will lead to an error when saving. His cell will not be cleared. In this case, we need to fix the code for saving the workbook.

igitur:
I try to replicate the behaviour from Excel or VBA. In Excel, if I enter a date, clear it, and enter a number again, it is again formatted as a date. So in Excel, clearing a cell does not clear its data type. 

In VBA there is something similar. There are different `Range.Clear` methods for clearing different things.

With my changes, if you call `IXLCell.Clear()`, without parameters, by default everything will be cleared, including the data type set to `Text`.

igitur:
>  However, if a person cleans a cell without clearing the type, it will lead to an error when saving. His cell will not be cleared. 

Please post code to reproduce this.

b0bi79:
Here is the test https://github.com/ClosedXML/ClosedXML/compare/develop...b0bi79:fix_cleared_save.

igitur:
I looked at the test that you added. It's definitely a bug, but I think it should be fixed in a different way.

b0bi79:
Here is another version of the solution, it also does not fit?
https://github.com/ClosedXML/ClosedXML/compare/develop...b0bi79:fix_cleared_save

igitur:
Bit busy at the moment. Exams! I'll look at this again in 2 weeks.

igitur:
Blackmail time: Help me with #521 and I will help you with this PR.

b0bi79:
Sorry, just now I saw your message. Still need help? I can help.

igitur:
I have a fix for this, but will finish it only next year.

igitur:
PS: I'm travelling abroad until 1 January, so won't be checking ClosedXML emails in the next 3+ weeks.

PR #535, opened 2017-10-08
WeakEvent pattern

igitur:


igitur:
Latest artifacts: https://ci.appveyor.com/project/Pyropace/closedxml/build/0.8.636/artifacts

pedershk:
What's going on with this pull request? Why is it not being merged and fixed so a new nuget can be released??

igitur:
@pedershk This branch severely degrades performance. See #540

igitur:
@Pankraty I've rebased this to the latest `develop`. This was my attempt to fix leaks from dangling event handlers. However, it severely impacts performance, as discussed in #540 .

igitur:
See https://github.com/ClosedXML/ClosedXML/issues/540#issuecomment-336169070 for an example project that illustrates the performance degradation.

Pankraty:
I will check it in the evening.

igitur:
Ok. AppVeyor is failing, because the new package that I'm using (WeakEvent), isn't signed. But it still works on the dev machine. We can sort this out later if we need to.

igitur:
Also notice that `XLReentrantEnumerableSet` has been dropped in favour of this new pattern.

Pankraty:
Now I see what means "severely"...
I tested Rob's example on my machine and with this branch, I got the following numbers:
```
Debug    68101 ms
Release  43041 ms
```
Profiling results are unambiguous:
image

In my opinion, the problem is not that the subscription and unsubscription work slowly, but that we have far too many of them. Forgive me, but I found the approach of #766 more viable. Yes, there are a few times more changes but the results are fascinating:
```
Debug     8638 ms
Release   6067 ms
```
And that's for **.NET 4.0**, as I haven't rebased it yet. With Core 2.0 it should be even better.

Pankraty:
Now that I rebased, for Core 2.0 version it takes 5.2 sec to run in Release mode.

But I discovered also that removing unsubscription from your solutions makes it almost as fast - 5.7 sec. This happens because unsubscription mechanism in WeakReference is not perfect, it depends linearly on the number of items, and since we unsubscribe from most of the events, overall complexity is O(n²).

Pankraty:
I know you have a hard choice to make, and I will understand if you'll choose your own solution eventually. But let me add a few more considerations.

I ran a well-known Pitterling example and got the following results:
**535 (igitur)**:
image

**766 (pankraty)**
image

Pictures look very similar, they have spikes at the same places. There are some differences, however:
1. Execution time is 64 vs 81 sec. It is not very precise since it was a memory profiling session, running in Debug configuration, but still, there is about 20-25% of the difference.
2. The garbage collector is called far less often in my variant. In fact, it explains a performance difference too. What I would like to emphasize is that 766 branch still contains many occurrences of ```var range = otherRange.AsRange();``` that load GC for nothing and that we can easily get rid of, improving the performance better yet. In the branch 535, we don't have such reserve, I am afraid.

In addition, using my approach we would have less external dependencies.

But again, I will admit any choice you make. Your solution is good.

Issue #545, opened 2017-10-16
Decimal math used to measure cell contents?


kendallb:
## Read and complete the full issue template

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

Current trunk in Git as of today

**What is the current behavior?**

Within the GetWidth() function of FontBaseExtensions the following lines exist:

```c#
            var textSize = GraphicsUtils.MeasureString(text, font);
            double width = (((textSize.Width / (double)7) * 256) - (128 / 7)) / 256;
            width = (double)decimal.Round((decimal)width + 0.2M, 2);
```

The problem is converting a double value to a decimal value and then doing the rounding on that value, and converting back to a double value is way more expensive than simply doing it all in double arithmetic. Decimal's are used primarily for currency calculations as it will avoid odd rounding issues when adding decimal values, but for this kind of math problem it is entirely unnecessary and just going to make it a whole lot slower.

Rather it should be written like so:

```c#
            var textSize = GraphicsUtils.MeasureString(text, font);
            double width = (((textSize.Width / (double)7) * 256) - (128 / 7)) / 256;
            width = Math.Round(width + 0.2, 2);
```

I will be happy to change the code and create a pull request unless I am completely missing something and there was a good reason to use decimal math here?

igitur:
This method is implicitly tested through all tests that call `XLColumns.AdjustToContents()`.

But the current calculation isn't entirely correct. It needs to be reworked. #297 is a good example of where the current logic fails. The best discussion I could find of how to calculate the correct width of a column is at http://polymathprogrammer.com/2010/01/18/calculating-column-widths-in-excel-open-xml/

I agree that the current method is not optimal, but does your fix make a significant difference in performance? If not, I'd prefer delaying this until we can sort out the exact logic.

kendallb:
Significant? In this case probably not because the call to MeasureString is going to take the vast majority of the time. If we were just running only that calculation, it would make a huge difference but something vastly slower is also involved :)

If we wish to time just the two functions themselves, I updated my PR with code in the sandbox to performance profile those two items, and this is what you get:

Running TimeDecimalRound
Action done in 00:00:13.8337287

Running TimeDoubleRound
Action done in 00:00:01.3669589

[...]

PR #547, opened 2017-10-16
Should not be using decimal math here as it is way slower than floating point

kendallb:
I mentioned this in this issue:

https://github.com/ClosedXML/ClosedXML/issues/545

Pretty simple fix but performance will suffer using decimal math here. All unit tests pass but as far as I can tell there is not actual test for this specific function.


vbjay:
It's more accurate.  Floating point has the propensity of having errors.

kendallb:
For currency calculations yes but this is simply converting pixels to twips and then rounding to 2 decimal places. Accuracy is not an issue here so it should be all floating point.

Issue #548, opened 2017-10-16
Regular expressions in formula conditions are not parsed correct


jongleur1983:
## Read and complete the full issue template

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

**Version of ClosedXML**
current develop branch

**What is the current behavior?**
When writing Unit tests for SumIf I tried to cover the examples from the Microsoft Excel documentation.
There the following formulas are given:

```
SUMIF(A2:A7, "Fruits", C2:C7)
SUMIF(A2:A7, "Vegetables", C2:C7)
SUMIF(A2:A7, "*es", C2:C7)
SUMIF(A2:A7, "", C2:C7)
```

The first argument in refers to a cell range containing string values, the second parameter is the condition, the third argument refers to the cell range containing numbers to sum up.

While the first two examples work as expected, the other two fail:

- [ ] `*es` as a condition parameter should be handled as kind of a regex, where * is a placeholder
- [ ] `""` (the empty string) should match all rows where the content is empty.
- [ ] `~*`should match the literal asterisk (`*`)
- [ ] `~?` should match the literal `?`
- [ ] `~~` should match the literal `~`

**What is the expected behavior or new feature?**
According to the Excel documentation at https://support.office.com/en-US/article/SUMIF-function-169B8C99-C05C-4483-A712-1697A653039B there are three special characters:

> You can use the wildcard characters—the question mark (?) and asterisk (*)—as the criteria argument. A question mark matches any single character; an asterisk matches any sequence of characters. If you want to find an actual question mark or asterisk, type a tilde (~) preceding the character.

For the two failing test cases above this means:
- *es should match any string that ends with the two characters `es`.
- the empty string should match any empty cell.

## Reproducibility

**Code to reproduce problem:**
```c#
/// <summary>
/// refers to Example 2 from the Excel documentation,
/// <see cref="https://support.office.com/en-us/article/SUMIF-function-169b8c99-c05c-4483-a712-1697a653039b?ui=en-US&rs=en-US&ad=US"/>
/// </summary>
/// <param name="expectedOutcome"></param>
/// <param name="formula"></param>
[TestCase( 2000, "SUMIF(A2:A7,\"Fruits\", C2:C7)")]
[TestCase(12000, "SUMIF(A2:A7,\"Vegetables\", C2:C7)")]
[TestCase( 4300, "SUMIF(A2:A7, \"*es\", C2:C7)")]
[...]

Issue #53, opened 2016-09-30
Sheet->Right-to-Left does not work


mgolmahi:
It looks ClosedXML does not load sheet options properly.

Symptom: Open an excel file including right-to-left sheet, and save it again by Save() or SaveAs() method, then open the file by excel, now the right-to-left sheet gets left to right.

ClosedXML.Excel.XLWorkbook _excel;
  _excel = new XLWorkbook("full file path");
....
do somting.
...
  _excel.SaveAs("full new file path");

PR #551, opened 2017-10-17
#548 conditions for strings, #53 improving MULTINOMIAL function, implementing SEC

jongleur1983:
#### What's this PR do?
improve wildcard pattern matching for string conditions:
- support ? wildcard
- support escaping wildcards
- support the empty string as a condition
- cover special cases with unit tests

- the MULTINOMIAL now throws the correct Exceptions as Excel does, when negative values or non-numerical values are given.
- implementing SEC function
- COUNTIF is fixed to support the empty string as a condition.

Sorry to mix two different things in this PR, but it built on each other because I stumbled over #548 while implementing functions and unit tests.

I think, the PR should be reviewed in three parts:

1. The improvements tackling #548 are covered in the CalcEngineHelper.cs.
2. SEC as additional function and the improvements of MULTINOMIAL are implemented in MathTrig.cs, while corresponding unit tests are in MathTrigTests.cs
3. The fix for CountIf relies on the changes from (1) (CalcEngineHelper), but is completed in Statistical.cs, corresponding tests are located in StatisticalTests.cs of course.

check with the behaviour of Excel ;)
I stumbled over two more problems I couldn't yet solve, but I don't think, these should be blocker for the PR:

1. Excel can deal with really big numbers. As I know from the past (not sure about the current state) it tended to be unprecise for really big or small numbers. With that in mind it could be that Excel does most calculations in floating point arithmetics, probably in `double` values, rounding afterwards. Currently most functions in ClosedXML are implemented based on interger or long and I found some cases where the Excel crunched much higher numbers than the ClosedXML implementation.
2. For formulas with a range covering a complete column (like A:A) the results differ. I tried a unit test for COUNTIF(A:A, ""), counting all empty values in column G. ClosedXML as well as Excel itself return "very" high numbers, but the exact numbers are different. Guessing again I assume the result is the number of supported lines, which would mean, even Excel versions might not behave equally here.

#### Questions:

### DIFF
--- a/ClosedXML/Excel/CalcEngine/CalcEngineHelpers.cs
+++ b/ClosedXML/Excel/CalcEngine/CalcEngineHelpers.cs
@@ -1,5 +1,7 @@
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Text.RegularExpressions;
 
 namespace ClosedXML.Excel.CalcEngine
@@ -25,8 +27,11 @@ namespace ClosedXML.Excel.CalcEngine
 
             // convert criteria to string
             var cs = criteria as string;
-            if (!string.IsNullOrEmpty(cs))
+            if (cs != null)
             {
+                if (cs == "")
+                    return cs.Equals(value);
+
                 // if criteria is an expression (e.g. ">20"), use calc engine
                 if (cs[0] == '=' || cs[0] == '<' || cs[0] == '>')
                 {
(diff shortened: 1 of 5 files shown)


igitur:
Tests failing. But more importantly, please submit a separate PR for each logical change.

jongleur1983:
okay.
I worked at both parts combined, and the entire PR is still limited on the one hand, so it was worth a try ;)
Going to split into two PRs then, https://github.com/ClosedXML/ClosedXML/pull/555 being the first, tackling the expression parser issue #548 alone.

igitur:
So this PR can be closed?

jongleur1983:
I'm not sure: can I still comment on a closed PR later?
This one is obsolete of course because of your request to split it, but it's not completely split yet, so even when closing it I would like to document that split here.
If I can comment here to link the next PR(s) as well, feel free to close it now, else I'd close it once I have moved the other parts.

igitur:
Ok, keeping this open. Just reference this issue in other PRs / issues and let me know when you feel all issues here are addressed.

PR #552, opened 2017-10-18
Should not be using decimal math here as it is way slower than floating point

kendallb:
I mentioned this in this issue:

https://github.com/ClosedXML/ClosedXML/issues/545

Pretty simple fix but performance will suffer using decimal math here. All unit tests pass but as far as I can tell there is not actual test for this specific function.

### DIFF
--- a/ClosedXML/Extensions.cs
+++ b/ClosedXML/Extensions.cs
@@ -229,7 +229,7 @@ namespace ClosedXML.Excel
             var textSize = GraphicsUtils.MeasureString(text, font);
 
             double width = (((textSize.Width / (double)7) * 256) - (128 / 7)) / 256;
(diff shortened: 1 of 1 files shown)


ClosedXML_Sandbox/PerformanceRunner.cs
```diff
@@ -40,6 +40,23 @@ public static void RunInsertTable()
             EmulateSave(workbook);
         }
 
+        public static void TimeDecimalRound()
```
igitur: Undo the changes in this file, please.

kendallb:
I removed the performance runner changes

igitur:
Thanks. This is scheduled for v0.91.

PR #553, opened 2017-10-18
Added interface file for XLWorkbook

### LINKED ISSUE #550: No interface for XLWorkbook?
- [x] Feature

**Version of ClosedXML**

0.89.0

In order to make it easier to unit tests code that uses ClosedXML, it would be greate if XLWorkbook had an IXLWorkbook interface that defines the public interface. Then you can easily inject it with dependency injection with a wrapper to create the actual class when you need it, which makes it possible to mock any calls to the workbook?

kendallb:
In order to make it easier to unit tests code that uses ClosedXML, I have created an IXLWorkbook interface to the XLWorkbook class. Now you can easily inject it with dependency injection with a wrapper to create the actual class when you need it, and it makes it possible to mock any calls to the workbook. Mentioned in this issue:

Fixes #550

### DIFF
--- a/ClosedXML/Excel/IXLWorkbook.cs
+++ b/ClosedXML/Excel/IXLWorkbook.cs
@@ -12,7 +12,7 @@ namespace ClosedXML.Excel
         ///   Gets an object to manipulate the worksheets.
         /// </summary>
         IXLWorksheets Worksheets { get; }
-        
+
         /// <summary>
         ///   Gets an object to manipulate this workbook's named ranges.
         /// </summary>
@@ -93,7 +93,7 @@ namespace ClosedXML.Excel
         Boolean DefaultShowRowColHeaders { get; }
 
         Boolean DefaultShowRuler { get; }
-        
+
         Boolean DefaultShowWhiteSpace { get; }
 
         Boolean DefaultShowZeros { get; }
--- a/ClosedXML/Excel/IXLWorkbook.cs
+++ b/ClosedXML/Excel/IXLWorkbook.cs
@@ -1,0 +1,220 @@
+using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Globalization;
+using System.IO;
+
+namespace ClosedXML.Excel
+{
+    public interface IXLWorkbook : IDisposable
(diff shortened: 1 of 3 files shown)


ClosedXML/Excel/XLWorkbook.cs
```diff
@@ -41,7 +41,7 @@ public enum XLCellSetValueBehavior
         Simple = 1,
     }
 
-    public partial class XLWorkbook: IDisposable
+    public partial class XLWorkbook: IXLWorkbook, IDisposable
```
igitur: You can remove the `IDisposable` here.

ClosedXML/Excel/IXLWorkbook.cs
```diff
@@ -0,0 +1,220 @@
+using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Globalization;
+using System.IO;
+
+namespace ClosedXML.Excel
+{
+    public interface IXLWorkbook : IDisposable
```
igitur: I haven't added it to the developer guidelines, but would you mind installing the [CodeMaid](http://www.codemaid.net/) extension and doing a "clean-up" of this file. It will sort the methods and properties alphabetically and do some other cleaning up. I generally do this for new files that are added to the project.

kendallb: Ok I did that, but it really did not change much? Mostly just removed a bit of incorrect white space?

igitur: Apologies, I was a bit unclear. It's the "Reorganize Active Document" (Ctrl+M, Z) that I meant. It orders the methods and properties alphabetically and according to type. The "Cleanup Active Document" (Ctrl+M, Space) is also important as I like to keep the whitespace consistent. The reason is that I often have to go way back in the logs and having clean, concise diffs make it much easier. 

I haven't had many contributions from other users, therefore the coding guidelines are poor. I have to expand on them and put them in the documentation.

image

kendallb:
No worries. I will do it later this week as I am traveling till Friday.

PR #555, opened 2017-10-18
#548 enhance string condition capabilities of CalcEngine

jongleur1983:
#### What's this PR do?
improve wildcard pattern matching for string conditions:
- support ? wildcard
- support escaping wildcards
- support the empty string as a condition
- cover special cases with unit tests

Splitting Pull Request #551, first part: string condition parsing.

The improvements themself are covered in the CalcEngineHelper.cs, tested by unit tests in the two test classes.

check with the behaviour of Excel ;)
I stumbled over two more problems I couldn't yet solve, but I don't think, these should be blocker for the PR:

1. Excel can deal with really big numbers. As I know from the past (not sure about the current state) it tended to be unprecise for really big or small numbers. With that in mind it could be that Excel does most calculations in floating point arithmetics, probably in `double` values, rounding afterwards. Currently most functions in ClosedXML are implemented based on interger or long and I found some cases where the Excel crunched much higher numbers than the ClosedXML implementation.
2. For formulas with a range covering a complete column (like A:A) the results differ. I tried a unit test for COUNTIF(A:A, ""), counting all empty values in column G. ClosedXML as well as Excel itself return "very" high numbers, but the exact numbers are different. Guessing again I assume the result is the number of supported lines, which would mean, even Excel versions might not behave equally here.

#### Questions:

### DIFF
--- a/ClosedXML/Excel/CalcEngine/CalcEngineHelpers.cs
+++ b/ClosedXML/Excel/CalcEngine/CalcEngineHelpers.cs
@@ -75,7 +75,7 @@ namespace ClosedXML.Excel.CalcEngine
 
                 // if criteria is a regular expression, use regex
                 if (cs.IndexOfAny(new[] { '*', '?' }) > -1)
-                {                    
+                {
                     var pattern = Regex.Replace(
                         cs,
                         "(" + String.Join(
--- a/ClosedXML/Excel/CalcEngine/CalcEngineHelpers.cs
+++ b/ClosedXML/Excel/CalcEngine/CalcEngineHelpers.cs
@@ -1,5 +1,7 @@
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Text.RegularExpressions;
 
 namespace ClosedXML.Excel.CalcEngine
@@ -25,8 +27,11 @@ namespace ClosedXML.Excel.CalcEngine
 
             // convert criteria to string
             var cs = criteria as string;
-            if (!string.IsNullOrEmpty(cs))
+            if (cs != null)
             {
+                if (cs == "")
+                    return cs.Equals(value);
+
                 // if criteria is an expression (e.g. ">20"), use calc engine
                 if (cs[0] == '=' || cs[0] == '<' || cs[0] == '>')
                 {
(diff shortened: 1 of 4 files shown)


ClosedXML/Excel/CalcEngine/CalcEngineHelpers.cs
```diff
@@ -54,11 +59,26 @@ internal static bool ValueSatisfiesCriteria(object value, object criteria, CalcE
                 }
 
                 // if criteria is a regular expression, use regex
-                if (cs.IndexOf('*') > -1)
+                if (cs.IndexOfAny(new[] { '*', '?' }) > -1)
                 {
-                    var pattern = cs.Replace(@"\", @"\\");
-                    pattern = pattern.Replace(".", @"\");
-                    pattern = pattern.Replace("*", ".*");
+                    var patternReplacements = new Dictionary<string, Tuple<string, string>>();
```
igitur: I think this should rather be a private static field in the class. Populate it in a static constructor.

ClosedXML_Tests/Excel/CalcEngine/MathTrigTests.cs
```diff
@@ -217,6 +217,82 @@ public void Mod()
...
+        /// </summary>
+        /// <param name="expectedOutcome"></param>
+        /// <param name="formula"></param>
+        [TestCase(63000, "SUMIF(A1:A4,\">160000\", B1:B4)")]
+        [TestCase(900000, "SUMIF(A1:A4,\">160000\")")]
+        [TestCase(21000, "SUMIF(A1:A4, 300000, B1:B4)")]
+        [TestCase(28000, "SUMIF(A1:A4, \">\" &C1, B1:B4)")]
+        public void SumIf_ReturnsCorrectValues_ReferenceExample1FromMicrosoft(int expectedOutcome, string formula)
+        {
+            using(var wb = new XLWorkbook())
```
igitur: Do a `Format Document` on all files before committing to ensure correct whitespace.

ClosedXML_Tests/Excel/CalcEngine/MathTrigTests.cs
```diff
@@ -217,6 +217,82 @@ public void Mod()
...
+
+                ws.Cell(1, 3).Value = 300000;
+
+                Assert.AreEqual(expectedOutcome, (double)ws.Evaluate(formula));
+            }
+        }
+
+        /// <summary>
+        /// refers to Example 2 from the Excel documentation,
+        /// <see cref="https://support.office.com/en-us/article/SUMIF-function-169b8c99-c05c-4483-a712-1697a653039b?ui=en-US&rs=en-US&ad=US"/>
```
igitur: Escape &'s with `&amp;` in doctext.

ClosedXML/Excel/CalcEngine/CalcEngineHelpers.cs
jongleur1983: I'm not obsessed with my solution here, so I can change it of course, but I would go for a compromise:
We're talking about a single dictionary, so it's not that much, but keeping a static dictionary in memory is wasting a little bit, but heap space, that's only necessary if there's any string pattern function actually being evaluated.
So I would go for a static, but lazy initialized field.

ClosedXML_Tests/Excel/CalcEngine/MathTrigTests.cs
jongleur1983: ouch, thanks, you're right. Just copied the URL from the browser ;)

jongleur1983:
For your request for COUNTIFS and SUMIFS: I'll think about it, but my last PR was mixing too much in your opinion, so let's see, when that's on my list ;)

ClosedXML_Tests/Excel/CalcEngine/MathTrigTests.cs
jongleur1983: did that now as well.

ClosedXML/Excel/CalcEngine/CalcEngineHelpers.cs
igitur: Yeah, I don't want a dictionary to be populated every time a cell formula is evaluated. I'll sacrifice the slight extra memory for the performance improvement.

igitur:
Thanks. This PR is scheduled for merging in v0.91/

igitur:
Fixes #548

PR #568, opened 2017-10-28
Feature/restrict c sharp version

jongleur1983:
#### What's this PR do?
AppVeyor fails to build with any newer C# features than C#6.
This PR changes the project settings to tell the compiler to restrict to C#6 as a language version, even if it's more up to date and could support newer features.

This avoids to accidentally use features that will fail in AppVeyor later.

Use an IDE and compiler that supports C#7, change the code to use some of the newer features: The build should fail.

### DIFF
--- a/ClosedXML/ClosedXML.csproj
+++ b/ClosedXML/ClosedXML.csproj
@@ -26,6 +26,7 @@
     <WarningLevel>4</WarningLevel>
     <NoWarn>1591</NoWarn>
     <Prefer32Bit>false</Prefer32Bit>
+    <LangVersion>6</LangVersion>
   </PropertyGroup>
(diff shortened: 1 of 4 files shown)


igitur:
We're changing to VS2017 format soon, so not sure whether this is useful for such a short period of time.

jongleur1983:
that's okay. 
I'll keep that change locally in my working copy to avoid building failures then.
Didn't know switching is planned.
Of course including it would be easier for any contributors. 

As far as I know the C# version is not defined anywhere on the project site, so even after changing to VS2017 and C# 7.x (whatever, even VS 2017 is something between 7.0 and 7.2 up to now) keeping the language version unrestricted will bringt that issue up again once there's a newer version.

That's thy I think asdding the langver attribute to the projects will benefit, even when the project upgrades to VS2017 and C#7.2 soon.

PR #570, opened 2017-10-30
OutOfMemoryException. Some range disposing.

b0bi79:
In some cases I received an OutOfMemotyException. As it turned out in the methods ShiftConditionalFormattingRows and ShiftConditionalFormattingColumns there is code
`
ConditionalFormats.Where (cf => cf.Range.Intersects(cellModel.AsRange)).ToList()
`
Here in the cycle, a Range is created, which is not destroyed.


igitur:
Thanks. If I can get the `WeakEvent` pattern to work, then this kind of statements won't be necessary. But I'm struggling to find a fast `WeakEventManager` implementation for `netstandard2.0`.

PR #575, opened 2017-11-01
Fix error when copying sheets with certain image anchors

### LINKED ISSUE #574: Error when copying sheet's with images with MoveAndSize placement.

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

e.g. 0.90.0

When copying sheets with images with certain anchors, the following exception is thrown:
"To set the width, the placement should be FreeFloating or Move"

Copy sheets as expected.

Probably not.

## Reproducibility

```c#
        [Test]
        public void CanCopySheetsWithAllAnchorTypes()
        {
            using (var stream = TestHelper.GetStreamFromResource(TestHelper.GetResourcePath(@"Examples\ImageHandling\ImageAnchors.xlsx")))
            {
                ws.CopyTo("Copy1");

                var ws2 = wb.Worksheets.Skip(1).First();
                ws2.CopyTo("Copy2");

                var ws3 = wb.Worksheets.Skip(2).First();
                ws3.CopyTo("Copy3");

                var ws4 = wb.Worksheets.Skip(3).First();
                ws3.CopyTo("Copy4");
            }
        }

```

bsermons:
#### What's this PR do?
Fixes #574 
Allows a sheet to be copied that contains images with MoveAndSize placement.  It just temporarily sets the image placement to something that allows width/height to be set while copying excel sheets.

This was the easiest thing to do to fix the issue.  If you decide the width/height setters should not have that placement constraint then I'll be glad to change the PR to do that instead.

#### Questions:
- Is there a blog post? no
- Does the knowledge base need an update? no
- Does this add new (C#) dependencies? no

### DIFF
--- a/ClosedXML/Excel/XLWorksheet.cs
+++ b/ClosedXML/Excel/XLWorksheet.cs
@@ -584,8 +584,9 @@ namespace ClosedXML.Excel
             foreach (var picture in Pictures)
             {
                 var newPic = targetSheet.AddPicture(picture.ImageStream, picture.Format, picture.Name)
-                    .WithPlacement(picture.Placement)
-                    .WithSize(picture.Width, picture.Height);
+                    .WithPlacement(XLPicturePlacement.FreeFloating)
+                    .WithSize(picture.Width, picture.Height)
+                    .WithPlacement(picture.Placement);
 
                 switch (picture.Placement)
                 {
--- a/ClosedXML_Tests/Excel/Loading/LoadingTests.cs
+++ b/ClosedXML_Tests/Excel/Loading/LoadingTests.cs
@@ -148,6 +148,26 @@ namespace ClosedXML_Tests.Excel
         }
 
         [Test]
+        public void CanCopySheetsWithAllAnchorTypes()
+        {
+            using (var stream = TestHelper.GetStreamFromResource(TestHelper.GetResourcePath(@"Examples\ImageHandling\ImageAnchors.xlsx")))
+            using (var wb = new XLWorkbook(stream))
+            {
+                var ws = wb.Worksheets.First();
+                ws.CopyTo("Copy1");
+
+                var ws2 = wb.Worksheets.Skip(1).First();
+                ws2.CopyTo("Copy2");
+
+                var ws3 = wb.Worksheets.Skip(2).First();
+                ws3.CopyTo("Copy3");
+
+                var ws4 = wb.Worksheets.Skip(3).First();
+                ws3.CopyTo("Copy4");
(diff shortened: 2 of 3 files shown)


ClosedXML/Excel/XLWorksheet.cs
```diff
@@ -584,8 +584,9 @@ public IXLWorksheet CopyTo(XLWorkbook workbook, String newSheetName, Int32 posit
             foreach (var picture in Pictures)
             {
                 var newPic = targetSheet.AddPicture(picture.ImageStream, picture.Format, picture.Name)
-                    .WithPlacement(picture.Placement)
-                    .WithSize(picture.Width, picture.Height);
+                    .WithPlacement(XLPicturePlacement.FreeFloating)
```
igitur: Isn't it sufficient to have only 2 lines? With Size first, and placement second?

ClosedXML_Tests/Excel/Loading/LoadingTests.cs
```diff
@@ -148,6 +148,26 @@ public void CanLoadFileWithImagesWithCorrectAnchorTypes()
         }
 
         [Test]
+        public void CanCopySheetsWithAllAnchorTypes()
```
igitur: This isn't really a loading test. Please move this test to `XLWorksheetTests`.

ClosedXML/Excel/XLWorksheet.cs
bsermons: Sorry, I got busy and missed your comments.  The problem is that there is an image with the `MoveAndSize` placement type saved in that excel file.  Trying to set the size in `WithSize` throws exceptions for images with this anchor type. See here: https://github.com/ClosedXML/ClosedXML/blob/230236e6ac43ff27c9947c966dd9cd1bc23ee06c/ClosedXML/Excel/Drawings/XLPicture.cs#L223

This PR just sets the placement first to something less restrictive on sizing and then resets it afterwards to the original anchor type.  The other alternative I see would be to remove the lines that throw exceptions in XLPicture.  You might have a better idea of how to fix it that I'm not aware of though.

igitur: Ok, I'm happy with this.

ClosedXML_Tests/Excel/Loading/LoadingTests.cs
igitur: Please do this, then I'll merge.

bsermons:
Made the changes you requested.  Let me know if you need anything else.

PR #576, opened 2017-11-01
Functions and tests math trig starting with C to F

jongleur1983:
#### What's this PR do?
improve functions from MathTrig category for CalcEngine,
where name starts with C to F and add unit tests

improved functions:

- COMBIN (tests were there already)
- FACT (+tests)
- FACTDOUBLE (+ tests)

new unit tests only:

- COS
- COSH
- DEGREES
- EVEN
- EXP

For FLOOR.MATH I refactored the tests so that each test checks one fact only: Splitted the long Test method with several assertions to a theory with several test cases where each executed test checks only one case.

Should work out straight forward through the code, I think.

compare with Execls behaviour.

### DIFF
--- a/ClosedXML_Tests/Excel/CalcEngine/MathTrigTests.cs
+++ b/ClosedXML_Tests/Excel/CalcEngine/MathTrigTests.cs
@@ -822,6 +822,47 @@ namespace ClosedXML_Tests.Excel.CalcEngine
             Assert.Throws<CellValueException>(() => XLWorkbook.EvaluateExpr(string.Format(@"FACT(""x"")")));
         }
 
+        [TestCase(0, 1L)]
+        [TestCase(1, 1L)]
+        [TestCase(2, 2L)]
+        [TestCase(3, 3L)]
+        [TestCase(4, 8L)]
+        [TestCase(5, 15L)]
+        [TestCase(6, 48L)]
+        [TestCase(7, 105L)]
+        [TestCase(8, 384L)]
+        [TestCase(9, 945L)]
+        [TestCase(10, 3840L)]
+        [TestCase(11, 10395L)]
+        [TestCase(12, 46080L)]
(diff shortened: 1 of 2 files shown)


ClosedXML_Tests/Excel/CalcEngine/MathTrigTests.cs
```diff
@@ -447,6 +447,72 @@ public void Base_ReturnsCorrectResultOnInputWithMinimalLength(int input, int the
             Assert.AreEqual(expectedResult, actual);
         }
 
+        [Theory]
```
igitur: Theory? What does that mean?

jongleur1983: A Theory is more than a fact, it's something like a general rule: For some pattern of inputs a rule appllies.
A [Fact] checks a single fact once (the method has no parameters and is executed once during test execution).

A [Theory] checks a rule for a set of input parameter sets.
In NUnit the [TestCase] attribute is in fact some kind of theory itself, but with [TestCase] you have to specify any test case individually.
Here the test data can be generated as any n greater or equal than 0 (in theory ;) ) is a valid input. 

In practise we have two restrictions here:

- the current implementation is restricted to Int64/long  as a result type, while Excel can cope with bigger integer values (I guess Microsoft uses double for any calculation in background), so we can't use too big inputs currently
-  with more test cases generated e.g. by the range attribute test execution time increases, so I'm careful to use a too big range here.

If you're reading through theoretical stuff about testing, using theories and with them more some kind of "general rules" is sometimes also referred to as property based testing (where the rule I referred to above is called a property), from mathmatics background you may know these rules under the term "invariant".

vbjay:
https://xunit.github.io/docs/getting-started-desktop.html#write-first-theory

PR #586, opened 2017-11-09
Use ExcelNumberFormat library to format strings.

### LINKED ISSUE #345: TEXT() incorrect for some dates
- [x] Bug
- [ ] Feature

The TEXT() function doesn't produce a zero for days or hours where the Excel equivalent would do so.

**If the current behavior is a bug, please provide the steps to reproduce and 
if possible a minimal demo of the problem with a sample spreadsheet.**

Sample code is included in the block below. Expected values shown in comments are those observed in MS Excel 2016 32-bit running on Windows 10.0.14393.

Excel appears to produce 0 instead of 12 for the hour component of times less than 1:00 AM, and 0 for the day only when the input number is zero or a positive number less than 1.

Excel also has the [Leap year bug](https://en.wikipedia.org/wiki/Leap_year_bug), which is not mimicked in the .NET implementation. All numbers from 2.0 up to 61.0 (exclusive) produce a day offset in ClosedXML that is 1 less than the Excel version.

My users are using this behavior of the TEXT function for formatting date spans where their dates are less than 30 days apart. (Note: I first attempted to use `ValueCached` on the cells where this was the case, but something about the process used appears not to provide cached values reliably.)

Unknown

```c#
public void Main()
{
    var wb = new XLWorkbook();

    var formula = "TEXT(0, \"d:h:mm\")";
    var result = ws.Evaluate(formula);
    Console.WriteLine("Evaluated \"{0}\": {1}", formula, result); // actual: 30:12:00. expected: 0:0:00

    var formula = "TEXT(1, \"d:h:mm\")";
    var result = ws.Evaluate(formula);
    Console.WriteLine("Evaluated \"{0}\": {1}", formula, result); // actual: 31:12:00. expected: 1:0:00

    var formula = "TEXT(60, \"d:h:mm\")";
    var result = ws.Evaluate(formula);
    Console.WriteLine("Evaluated \"{0}\": {1}", formula, result); // actual: 28:12:00. expected: 29:0:00

    var formula = "TEXT(61, \"d:h:mm\")";
    var result = ws.Evaluate(formula);
    Console.WriteLine("Evaluated \"{0}\": {1}", formula, result); // actual: 1:12:00. expected: 1:0:00

    Console.Write("Press any key to continue...");
    Console.ReadKey();
}
```

[Book1.xlsx](https://github.com/ClosedXML/ClosedXML/files/1072745/Book1.xlsx)

### LINKED ISSUE #566: DateTime not being formatted correctly?

- [x] Bug

**Version of ClosedXML**

0.90.0

I am reading a date from a .xlsx file that is formatted like so:

"10/27/2017 21:00"

using the default Excel formatting for 24 hour times. However when this is loaded via ClosedXML, if I render it to a string using GetFormattedString() I get this as the result:

"0/26/13 9:00;@"

It should be returning the properly formatted date:

"10/27/2017 21:00"

No idea

## Reproducibility

- [x ] I attached a sample spreadsheet.  (You can drag files on to this issue)
[bad-date.xlsx](https://github.com/ClosedXML/ClosedXML/files/1423459/bad-date.xlsx)

igitur:
Fixes #566 
Fixes #345

### DIFF
--- a/ClosedXML/ClosedXML.csproj
+++ b/ClosedXML/ClosedXML.csproj
@@ -50,6 +50,9 @@
       <HintPath>..\packages\DocumentFormat.OpenXml.2.7.2\lib\net40\DocumentFormat.OpenXml.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="ExcelNumberFormat, Version=1.0.2.0, Culture=neutral, PublicKeyToken=23c6f5d73be07eca, processorArchitecture=MSIL">
+      <HintPath>..\packages\ExcelNumberFormat.1.0.2\lib\net20\ExcelNumberFormat.dll</HintPath>
+    </Reference>
     <Reference Include="FastMember.Signed, Version=1.0.0.9, Culture=neutral, PublicKeyToken=9e8f22703bef9a29, processorArchitecture=MSIL">
       <HintPath>..\packages\FastMember.Signed.1.1.0\lib\net40\FastMember.Signed.dll</HintPath>
       <Private>True</Private>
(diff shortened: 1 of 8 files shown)


igitur:
TODO: Add unit tests

igitur:
Depends on https://github.com/andersnm/ExcelNumberFormat/pull/11

PR #602, opened 2017-11-24
Added sorting for pivot fields.

b0bi79:
Added sorting for pivot fields.


ClosedXML/Excel/PivotTables/IXLPivotTable.cs
```diff
@@ -91,6 +91,13 @@ public enum XLPivotTableTheme
         PivotStyleMedium9
     }
 
+    public enum XLPivotSortType
+    {
+        Manual = 0,
```
igitur: I would rather call this `XLPivotSortType.DefaultSort`.

igitur:
Add unit tests, please.

PR #606, opened 2017-11-28
Fix System.InvalidOperationException: Error in implicit conversion. 

b0bi79:
Fix "System.InvalidOperationException: Error in implicit conversion. Cannot convert null object." when pf.MultipleItemSelectionAllowed is null.

### DIFF
--- a/ClosedXML/Excel/XLWorkbook_Load.cs
+++ b/ClosedXML/Excel/XLWorkbook_Load.cs
@@ -683,7 +683,7 @@ namespace ClosedXML.Excel
                                         else
                                             throw new NotImplementedException();
                                     }
-                                    else if (pf.MultipleItemSelectionAllowed != null && BooleanValue.ToBoolean(pf.MultipleItemSelectionAllowed))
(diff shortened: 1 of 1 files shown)


ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -683,7 +683,7 @@ private void LoadSpreadsheetDocument(SpreadsheetDocument dSpreadsheet)
                                         else
                                             throw new NotImplementedException();
                                     }
-                                    else if (BooleanValue.ToBoolean(pf.MultipleItemSelectionAllowed))
+                                    else if (pf.MultipleItemSelectionAllowed != null && BooleanValue.ToBoolean(pf.MultipleItemSelectionAllowed))
```
igitur: I think use `OpenXmlHelper.GetBooleanValueAsBool(pf.MultipleItemSelectionAllowed, false)` rather.

b0bi79: Thanks

PR #615, opened 2017-12-04
Cell value - stack overflow exception

Offline24:
#### What's this PR do?
Prevent throwing _uncatchable_ StackOverflowException when trying get value of cell with circular referenced formula.
This PR proposes counting the depth of recursion and throwing _catchable_ exception when depth limit will be exceeded.
See new unit tests, check changes in IXLCell interface, rest of code.
PS, first commit is only refactor (extract method), second contains logic changes.
Create file with circular referenced cells, try read value of cell, run new unit tests.
I am not sure if the solution proposed by me is a good idea. Maybe an circular reference detection algorithm is a better idea?
#### Questions:
- Is there a blog post? No
- Does the knowledge base need an update? Yes(?). I changed interface, but this is non breaking change (new method). Also 'XLCell.Value' behavior has changed.
- Does this add new (C#) dependencies? No


igitur:
Thanks @Offline24 . We're busy re-evaluating how the whole calculation chain works. I'm parking this issue for now.

igitur:
Thanks for the contribution. #746  and #750 should solve the issues you're experiencing in this regard, so I'm closing this PR. We can reopen if those PRs don't sufficiently address them.

PR #617, opened 2017-12-05
Fix. Clear cells while save

b0bi79:
Replacement #534


igitur:
I think this will also be fixed by #727 . I copied your test over to https://github.com/igitur/ClosedXML/blob/more-clearing-options-with-conditional-formats/ClosedXML_Tests/Excel/Cells/XLCellTests.cs#L430

PR #641, opened 2018-01-06
Fix. Failed to "SaveAs" a file with a ReadOnly attribute

### LINKED ISSUE #640: Failed to "SaveAs" a file with a ReadOnly attribute

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.90.0

If open a file having a read-only attribute and then try to save it as a copy (with a different name) the operation fails (UserNotAuthorizedException). The copy of the original file is created though.

The file must be successfully saved at the specified location. It must have no ReadOnly attribute.

No (as far as I am aware of)

## Reproducibility

A pull request with a test case will be provided soon.

```c#
public void Main()
{
            string id = Guid.NewGuid().ToString();
            string original = string.Format("{0}original{1}.xlsx", _tempFolder, id);
            string copy = string.Format("{0}copy_of_{1}.xlsx", _tempFolder, id);

            {
                var sheet = wb.Worksheets.Add("TestSheet");
                wb.SaveAs(original);
            }
            System.IO.File.SetAttributes(original, FileAttributes.ReadOnly);

            using (var wb = new XLWorkbook(original))
            {
                wb.SaveAs(copy); // This step fails
            }
}
```

Pankraty:
#### Fixes #640

### DIFF
--- a/ClosedXML/Excel/XLWorkbook.cs
+++ b/ClosedXML/Excel/XLWorkbook.cs
@@ -498,7 +498,10 @@ namespace ClosedXML.Excel
             else if (_loadSource == XLLoadSource.File)
             {
                 if (String.Compare(_originalFile.Trim(), file.Trim(), true) != 0)
+                {
                     File.Copy(_originalFile, file, true);
+                    File.SetAttributes(file, FileAttributes.Normal);
+                }
 
                 CreatePackage(file, GetSpreadsheetDocumentType(file), options);
             }
--- a/ClosedXML_Tests/Excel/Saving/SavingTests.cs
+++ b/ClosedXML_Tests/Excel/Saving/SavingTests.cs
@@ -6,6 +6,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 
 namespace ClosedXML_Tests.Excel.Saving
 {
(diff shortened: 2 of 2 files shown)


igitur:
In the case that the file does already exist and be read only, I don't think we should clear the flag.

vbjay:
Exactly why I asked what I asked in the issue. He did say the file to be
written does not exist.

igitur:
In his specific case the file doesn't already exist, but if you look at the patch he submitted, it doesn't make an provision for when the file does exist. And it should.

Pankraty:
>  it doesn't make an provision for when the file does exist. And it should.
With the provided fix, method SaveAs fails on overwriting an existing file with a ReadOnly flag, and silently overwrites one without this flag (since "overwrite" parameter was specified in File.Copy method). 
Does not it seem like an expected behavior?

igitur:
Ah OK, that sounds good then.

Pankraty:
Added couple more tests for these cases.

ClosedXML/Excel/XLWorkbook.cs
```diff
@@ -498,7 +498,10 @@ public void SaveAs(String file, SaveOptions options)
             else if (_loadSource == XLLoadSource.File)
             {
                 if (String.Compare(_originalFile.Trim(), file.Trim(), true) != 0)
+                {
                     File.Copy(_originalFile, file, true);
+                    File.SetAttributes(file, FileAttributes.Normal);
```
igitur: And if `_loadSource` is something else? I think the `File.SetAttributes` should be lower down, outside the `if` statements.

igitur: I'm wrong. It's not needed for the other `_loadSource` values. I do, however, think we need to catch that specific exception and rethrow it with a more meaningful message. What do you think?

Pankraty: That sounds reasonable. On the other hand, File.Copy may fail due to plenty of reasons (invalid path, network failure, disk is full, user has no permissions, file is busy, etc.). Each of them would cause a different exception, more or less specific.
Does this make sense to process some of these cases (like overwriting a read-only file) separately, making the exception more specific than is thrown by the framework? Probably, yes, assuming than user would expect some kind of IO.Exception to be thrown on saving failure, but I am not sure.

vbjay:
I don't see a problem of using InnerException. Throw an exception that
contains the original exception.

igitur:
Merged. We can deal with the exceptions later. Need to think about it, list what we can and decide which we're going to catch.

Thanks for the contribution.

ClosedXML_Tests/Excel/Saving/SavingTests.cs
```diff
@@ -1,15 +1,28 @@
...
 namespace ClosedXML_Tests.Excel.Saving
 {
     [TestFixture]
     public class SavingTests
     {
+        private string _tempFolder;
+        private List<string> _tempFiles;
+
+        [SetUp]
+        public void Setup()
```
igitur: I missed this. Can you check whether you can rather use the `using (var tf = new TemporaryFile(...))` pattern as in `OleDbTests`. Using that pattern would clean up any test files automatically.

PR #643, opened 2018-01-07
Use .HasValue property of nullable instead of using == null && linq usage improvements

Mafii:
Just a minor code quality improvement.

### DIFF
--- a/ClosedXML/Utils/OpenXmlHelper.cs
+++ b/ClosedXML/Utils/OpenXmlHelper.cs
@@ -11,7 +11,7 @@ namespace ClosedXML.Utils
 
         public static bool GetBooleanValueAsBool(BooleanValue value, bool defaultValue)
         {
-            return value.HasValue ? defaultValue : value.Value;
+            return value.HasValue ? value.Value : defaultValue;
         }
     }
 }
--- a/ClosedXML/Utils/OpenXmlHelper.cs
+++ b/ClosedXML/Utils/OpenXmlHelper.cs
@@ -11,7 +11,7 @@ namespace ClosedXML.Utils
 
         public static bool GetBooleanValueAsBool(BooleanValue value, bool defaultValue)
         {
-            return value == null ? defaultValue : value.Value;
+            return value.HasValue ? defaultValue : value.Value;
         }
     }
 }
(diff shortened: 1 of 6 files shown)


igitur:
Unit tests are failing.

ClosedXML/Utils/OpenXmlHelper.cs
```diff
@@ -11,7 +11,7 @@ public static BooleanValue GetBooleanValue(bool value, bool defaultValue)
 
         public static bool GetBooleanValueAsBool(BooleanValue value, bool defaultValue)
         {
-            return value == null ? defaultValue : value.Value;
+            return value.HasValue ? defaultValue : value.Value;
```
Pankraty: Since the ```value``` is not a ```Nullable<bool>``` but is a ```BooleanValue``` instance, this will cause ```NullReferenceException``` if ```value == null```.

igitur: I think it should just be
```c#
 return value.HasValue ?  value.Value : defaultValue;
```

@Mafii, Can you please fix?

Mafii: @igitur I will this weekend. Thanks for the heads up and sorry for the inconvenience!

Mafii: The problem is quite simple: 

```
return value.HasValue ? defaultValue : value.Value;
```

This piece of code will return the default value instead of the value when the value is available, and the other way around. Flipping the two values after the ? will do it.

ClosedXML/Utils/OpenXmlHelper.cs
```diff
@@ -11,7 +11,7 @@ public static BooleanValue GetBooleanValue(bool value, bool defaultValue)
 
         public static bool GetBooleanValueAsBool(BooleanValue value, bool defaultValue)
         {
-            return value == null ? defaultValue : value.Value;
+            return value.HasValue ? value.Value : defaultValue;
```
Pankraty: Again, ```BooleanValue``` is a class, not a Nullable<T>. So the ```value``` variable can be ```null```, and with this ```value.HasValue``` will trigger ```NullReferenceException```. You should check both if ```value != null``` AND ```value.HasValue```.

igitur:
Please run the unit tests before pushing a commit. Something is still failing.

PR #650, opened 2018-01-11
Support TargetFramework with a lower version (4.0)

Pankraty:
**Version of ClosedXML**

0.88 — 0.90

Starting from version 0.88 ClosedXml requires framework version 4.5.2 or higher. Thus the library cannot be installed with the nuget package manager to projects that uses the older versions of the framework.

Support framework versions 4.0 - 4.5.1 as well.

Yes, up till the version 0.87.1

## Reproducibility

I realize that framework 4.0 is something quite ancient, but don't see clear reasons why the target framework version was changed in the 0.88. Currently, we have a few solution with several hundred csproj-es in each of them - all reference target framework 4.5, and a few of them use ClosedXML. Update all these projects only to be able to install a newer version of ClosedXML would be insane, obviously. At the same time, there are only two methods used in entire ClosedXML that are not presented in the framework 4.0:

```c#
PropertyInfo.GetValue(object obj); // used twice in ClosedXML
PropertyInfo.SetValue(object obj, object value); // used in 6 places in ClosedXML_Sandbox
```

Luckily, they both have overloads with an additional argument and exactly the same functionality:
```c#
PropertyInfo.GetValue(object obj, object[] index);
PropertyInfo.SetValue(object obj, object value, object[] index);
```
Adding additional ```null``` in 8 places I was able to build the solution and pass all tests.
So, unless I missed some important considerations I ask you to change TargetFramework back to 4.0.

### DIFF
--- a/ClosedXML/Excel/XLWorkbook.cs
+++ b/ClosedXML/Excel/XLWorkbook.cs
@@ -498,7 +498,10 @@ namespace ClosedXML.Excel
             else if (_loadSource == XLLoadSource.File)
             {
                 if (String.Compare(_originalFile.Trim(), file.Trim(), true) != 0)
+                {
                     File.Copy(_originalFile, file, true);
+                    File.SetAttributes(file, FileAttributes.Normal);
+                }
 
                 CreatePackage(file, GetSpreadsheetDocumentType(file), options);
             }
--- a/ClosedXML_Tests/Excel/Saving/SavingTests.cs
+++ b/ClosedXML_Tests/Excel/Saving/SavingTests.cs
@@ -6,6 +6,7 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 
 namespace ClosedXML_Tests.Excel.Saving
 {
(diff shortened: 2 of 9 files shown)


vbjay:
I think upgrading to at least 4.5.2 would be something you should do. Yes it would be work.  

The process is made simpler by doing this:

1. Create branch upgradeFramework
1. Change versions of projects.
2. Close and save and reload in VS. 
3. Commit
4. In Package manager console run `Update-Package -reinstall`
5. Commit
5. Rebuild solution.  
6. Bughunt
7. Commit bug fixes
8. Squash bug fixes and Update Package commit
9. Merge into dev branch.

Pankraty:
I appreciate your advice, but we are talking about a huge enterprise system with a large legacy part poorly covered with tests. Technically, upgrading procedure is not too hard (except, there are a few repositories that I am not allowed to commit to and a few more that I don't know they exist), however convincing all the code owners, product owners, and project managers to invest their teams time to upgrading (and testing a whole system) with no clear reasons why - would be way harder. 
My opinion is, that infrastructural libraries (such as ClosedXML) should be as less demanding as possible. If there were some reasons why 4.5.2 was chosen (performance issues, or maybe security issues, or stability, whatsoever) - OK, fine, I would accept this. But I suppose this would be covered with a test, so this would be obvious - at 4.5 this works wrong, at 4.5.2 this works right, no questions. Since everything works fine at 4.0 I really see no point in sticking to 4.5.2.

igitur:
We upgraded because OpenXML required .NET 4.5.2. Later, in OpenXML v2.7.2 they started supporting lower framework versions again. I'll consider this, but keep in mind that we're working on a branch to support .NET Standard 2.0. I would like to keep the number of compiler directives to a minimum. Maybe it's not an issue, maybe it will be. 

But really, you should upgrade! It's way overdue.

ClosedXML/Excel/XLWorkbook.cs
```diff
@@ -498,7 +498,10 @@ public void SaveAs(String file, SaveOptions options)
             else if (_loadSource == XLLoadSource.File)
             {
                 if (String.Compare(_originalFile.Trim(), file.Trim(), true) != 0)
+                {
                     File.Copy(_originalFile, file, true);
+                    File.SetAttributes(file, FileAttributes.Normal);
```
igitur: This shouldn't be here. On which branch did you base your PR?

ClosedXML_Tests/Excel/Saving/SavingTests.cs
```diff
@@ -101,5 +114,97 @@ public void CanSaveAndValidateFileInAnotherCulture()
                 }
             }
         }
+
```
igitur: Also shouldn't be in this PR.

ClosedXML/Excel/XLWorkbook.cs
Pankraty: You are right, I made a branch incorrectly. Only commit d5dde24 should be included in PR. I will fix it.

igitur:
Why did you close this?

Pankraty:
I feel stupid, but I had all my commits included in PR no matter what I did. So I removed the branch (which closed the PR automatically), hoping to figure it out when I have time, but still no luck (

vbjay:
So is it just GitHub or do you have local git loss too?

Yes. You never want to create a pull request from your master. Always
create a branch that you control what commits go into it. Your pull request
should say you want to pull your branch into our develop branch.

Pankraty:
> Always create a branch that you control what commits go into it

Yeah, now I got it ) But the reason I guess is that I created my previous PR (#640) from ```develop```, without creating a separate branch for it. Now those 4 commits seem like if they were nailed.

igitur:
@Pankraty Add me on Google Hangouts please. username igitur

vbjay:
So create a branch off of d5dde24 or whichever commit Sha you want from the
list of commits in this pull request. Then create a new pull request and
mention this pull request in it.

Mention by typing

Related pull request #---.

--- being 650

PR #661, opened 2018-01-19
FIX a single quote in a worksheet name issue

### LINKED ISSUE #605: Crash when opening a Worksheet named with a simple quote
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

e.g. 0.90.0

ClosedXml crash when opening a workbook that contains a worksheet whose name contains a simple quote and, if this sheet has a named range.
In the given file, if I remove the apostrophe in the name, it doesn't crash anymore.

## Reproducibility
```c#
            var path = "\Error2.xlsx";
```
[Error2.xlsx](https://github.com/ClosedXML/ClosedXML/files/1506634/Error2.xlsx)

Pankraty:
This PR fixes #605 

When saving a workbook with sheets which names contain a single quote sign (aka apostrophe) Excel escapes it with another quote and puts the sheet name into single quotes too. So the worksheet L'E from the original issue is represented as 'L''E' in the inner XML. Apparently, to avoid unescaping issues Excel prohibits setting worksheet names either starting or ending with a single quote.

ClosedXML, however, does not do the same. It allows using a single quote in a worksheet name, it can save and open books with apostrophes in the middle of worksheet name, but it fails at opening such books if they were saved by Excel, as all single quotes are being washed in the ParseReference method.

This PR 

1. prohibits using a single quote at the beginning or end of a worksheet name,
2. establishes proper escaping and unescaping quotes on loading and saving workbook,
3. adds tests covering these features.

I did not cover with an automatic test whether the references are escaped correctly in a resulted file, although I have checked this manually, and Excel is satisfied with the result too (initially, it complained that file has to be recovered).


ClosedXML_Tests/Excel/Worksheets/XLWorksheetTests.cs
```diff
@@ -234,5 +234,83 @@ public void CanCopySheetsWithAllAnchorTypes()
...
+
+            Assert.DoesNotThrow(saveAndOpenWorkbook);
+            Assert.AreEqual(title, savedTitle);
+        }
+
+        /// <summary>
+        /// Excel escapes symbol ' in worksheet title so we have to process this correctly.
+        /// </summary>
+        [Test]
+        public void CanOpenWorksheetWithEscapedApostrophe()
```
igitur: I think this test belongs with the others in `.\ClosedXML_Tests\Excel\Loading\LoadingTests.cs`. Do you agree?

Pankraty: Yes, I agree. Will move it.

ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -1324,11 +1324,26 @@ private static void ParseReference(string item, out string sheetName, out string
...
         }
 
+        private static string UnescapeName(string worksheetName)
+        {
+            return worksheetName
+                .Trim('\'')
+                .Replace("''", "'");
+        }
+
+        private static string EscapeName(string worksheetName)
```
igitur: I think you should consolidate this function into `StringExtensions.WrapSheetNameInQuotesIfRequired` and rename it to `EscapeSheetName`.

ClosedXML/Excel/XLWorkbook_Save.cs
```diff
@@ -731,7 +731,7 @@ orderby worksheet.Position
                     var definedNameText = worksheet.PageSetup.PrintAreas.Aggregate(String.Empty,
                         (current, printArea) =>
                             current +
-                            ("'" + worksheetName + "'!" +
+                            (EscapeName(worksheetName) + "!" +
```
igitur: The single quotes are necessary when the sheet name has a space too.

igitur: That's why some tests are failing.

Pankraty: I see. And that is why some of the existing tests fail.

ClosedXML/Excel/XLWorkbook_Load.cs
igitur: If you do that, it should fix the failing tests.

igitur:
Also, you seem like you're comfortable with git. When all tests and I approve the changes, would you mind squashing all the commits into one and rebasing? Going forward, I'm attempting to make the commit history cleaner.

Pankraty:
Well, not that I am "comfortable" with Git, but I do my best to learn it (all my previous projects used SVN). I agree with about having a cleaner history.

Pankraty:
With this commit all tests except 2 are green. As I see they both compare the result to the saved files (AutoFilter.xlsx and AdjustToContentsWithAutoFilter.xlsx). Sheet names in those files are escaped in quotes even though they don't have spaces or quotes in. Saving them by Excel removes the escaping characters. Shall I update these two files too?

vbjay:
When you merge, you can tell it to squash.
https://github.com/blog/2141-squash-your-commits

igitur:
@vbjay But I think (not sure) that if I do it, the commit line will be under my name instead of the actual collaborator, which is unfair. Am I wrong?

vbjay:
No.  

image

image

igitur:
What happens if commits from different contributors are squashed? Which one gets the squash commit?

vbjay:
Good question.

igitur:
Please rebase again (since `develop` branch has evolved).

Pankraty:
> Please rebase again (since develop branch has evolved).

Done

Issue #655, opened 2018-01-17
Can not delete picture


ZhXueming:
**Do you want to request a *feature* or report a *bug*?**
- [x] Bug

**Version of ClosedXML**
0.90

**What is the current behavior?**

I load excel file and found the picture and delete it.
then save to file and reopen it.
but the picture even exist.

vbjay:
We have an issue template for a reason.

igitur:
Fixed in #674

PR #674, opened 2018-01-31
FIX Pictures not being deleted

### LINKED ISSUE #673: Pictures not being deleted
- [x] Bug
- [ ] Feature

**Version of ClosedXML**
0.90.0

Calling the Delete() method on picture doesn't remove them from the worksheet.
By that I mean that they're being removed from "Pictures", but when the document is saved, the pictures are stil in the workbook.
This is true whether you use worskeet.Pictures.Delete(pictureName) or simply picture.Delete().

Deleted pictures shouldn't come back when the document is saved, they should be deleted

I haven't used previous versions.

## Reproducibility
```c#
public void Main()
{
           var path = @"C:\path\to\file.xlsx"
            var ws = wb.Worksheet(1);

            while (ws.Pictures.Count > 0)
            {
                var pic = ws.Pictures.First<ClosedXML.Excel.Drawings.IXLPicture>();
                Console.WriteLine(pic.Name);
                ws.Pictures.Delete(pic.Name);
            }
            // The loop ends, there are no more pictures
            wb.SaveAs(path);
            // Open the file, the pictures are still here
}
```
[Example spreadsheet](https://github.com/ClosedXML/ClosedXML/files/1678301/picTest.xlsx)

Pankraty:
This PR fixes #673, #655 

Previously, ClosedXML did not delete images from the file, only added them.
This PR
1. Puts deleted pictures into a separate collection and then removes them from a workbook as well.
2. Removes drawings that have no contents (it is not absolutely necessary, just prevents empty ```drawing``` sections from being stored in the file).
3. Original test expanded to check not only that pictures were deleted from a workbook in memory, but that they were removed from a file too.

### DIFF
--- a/ClosedXML/Excel/Drawings/IXLPictures.cs
+++ b/ClosedXML/Excel/Drawings/IXLPictures.cs
@@ -29,6 +29,8 @@ namespace ClosedXML.Excel.Drawings
 
         void Delete(IXLPicture picture);
 
+        ICollection<String> Deleted { get; }
+
         IXLPicture Picture(String pictureName);
 
         bool TryGetPicture(string pictureName, out IXLPicture picture);
--- a/ClosedXML/Excel/Drawings/IXLPictures.cs
+++ b/ClosedXML/Excel/Drawings/IXLPictures.cs
@@ -29,8 +29,6 @@ namespace ClosedXML.Excel.Drawings
 
         void Delete(IXLPicture picture);
 
-        ICollection<String> Deleted { get; }
-
         IXLPicture Picture(String pictureName);
 
         bool TryGetPicture(string pictureName, out IXLPicture picture);
--- a/ClosedXML/Excel/Drawings/XLPictures.cs
+++ b/ClosedXML/Excel/Drawings/XLPictures.cs
@@ -12,10 +12,12 @@ namespace ClosedXML.Excel.Drawings
     {
         private readonly List<XLPicture> _pictures = new List<XLPicture>();
         private readonly XLWorksheet _worksheet;
+        public ICollection<String> Deleted { get; private set; }
 
         public XLPictures(XLWorksheet worksheet)
         {
             _worksheet = worksheet;
+             Deleted = new HashSet<String>();
         }
 
         public int Count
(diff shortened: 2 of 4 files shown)


igitur:
Thanks, but I'd like a different approach. I've learnt a few lessons from deleting other elements in the past. Rather do it like worksheets are currently deleted. There is a `XLWorksheets.Deleted` property that is used to keep track of deleted worksheets. We should do the same for images. The benefits of this approach is:

- RelId of untouched images don't change on repeated saves (makes testing much easier)
- Any XML elements that are not loaded/saved by ClosedXML is left unchanged.

Pankraty:
Fair point. Will stick to this approach.

Pankraty:
I've updated the PR.

ClosedXML/Excel/Drawings/IXLPictures.cs
```diff
@@ -29,6 +29,8 @@ public interface IXLPictures : IEnumerable<IXLPicture>
 
         void Delete(IXLPicture picture);
 
+        ICollection<String> Deleted { get; }
```
igitur: Doesn't need to be in interface. This is only for internal purposes.

ClosedXML/Excel/Drawings/XLPictures.cs
```diff
@@ -94,7 +96,17 @@ public void Delete(IXLPicture picture)
 
         public void Delete(string pictureName)
         {
-            _pictures.RemoveAll(picture => picture.Name.Equals(pictureName, StringComparison.OrdinalIgnoreCase));
+            var picturesToDelete = _pictures
+                .Where(picture => picture.Name.Equals(pictureName, StringComparison.OrdinalIgnoreCase))
+                .ToList();
```
igitur: Is the `.ToList()` required?

Pankraty:
>Doesn't need to be in interface. This is only for internal purposes

Done

vbjay:
Yes.  The ToList is required to create a solidified current result.  You
want an ienumerable that won't be affected by the removal in the loop.

Pankraty:
Could you provide a sample file, please? It worked fine on a test example.

ZhXueming:
It seems thant worksheetPart.DrawingsPart.WorksheetDrawing not been maintenanced.

igitur:
Weird... I thought I checked that there were no orphaned elements. Anyway, please open a new issue and complete the template. It will require a new PR too.

ZhXueming:
Hello, Pankraty
After delete the DrawingPart, I call the following code, and work fine.
-------
                     OpenXmlElement elm = GetAnchorFromImageId(worksheetPart, removedPicture);
                     elm?.Remove();
-------

ZhXueming:
[test.xlsx](https://github.com/ClosedXML/ClosedXML/files/1706475/test.xlsx)
I uploaded the test excel file. 
In my prg, I found the two pictures after 44 row, and delete them and save as test2.xlsx.
Open test2.xlsx then you will find the problem.

PR #676, opened 2018-01-31
Workaround when parsing html color string fails

### LINKED ISSUE #675: Exception when creating new XLWorkbook object with list separator set to # in regional settings.
https://github.com/ClosedXML/ClosedXML/blob/31206f0258391561bfc5fc1f89819b01339dfdcb/ClosedXML/Excel/Style/Colors/XLColor_Static.cs#L49

- [x] Bug

**Version of ClosedXML**

0.91

I use ClosedXML on machine where all Windows accounts are set with Croatian regional setting, and list separator set to # instead ; 

screenshot_20180131_100101

Exceptions is thrown when creating new XLWorkbook object in file ClosedXML/Excel/Style/Colors/XLColor_Static.cs on line 49.

Exception and Stack trace:

System.Exception occurred
  HResult=0x80131500
  Message= is not a valid value for Int32.
  Source=System

StackTrace:
   at System.ComponentModel.BaseNumberConverter.ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, Object value)
   at System.Drawing.ColorConverter.ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, Object value)
   at System.Drawing.ColorTranslator.FromHtml(String htmlColor)
   at ClosedXML.Excel.XLColor.FromHtml(String htmlColor) in ...\ClosedXML\Excel\Style\Colors\XLColor_Static.cs:line 49
   at ClosedXML.Excel.XLWorkbook.InitializeTheme() in ...\ClosedXML\Excel\XLWorkbook.cs:line 325
   at ClosedXML.Excel.XLWorkbook..ctor(XLEventTracking eventTracking) in ...\ClosedXML\Excel\XLWorkbook.cs:line 744
   at ClosedXML.Excel.XLWorkbook..ctor(String file, XLEventTracking eventTracking) in ...\ClosedXML\Excel\XLWorkbook.cs:line 770
   at ClosedXML.Excel.XLWorkbook..ctor(String file) in ...\ClosedXML\Excel\XLWorkbook.cs:line 765

No exception thrown. Workaround is to create new CultureInfo object and set list separator to ; then use this culture in program.

No

## Reproducibility

Bug is reproducible when creating new XLWorkbook object with above Regional settings.

```VB

Dim workBook As XLWorkbook = New XLWorkbook()

### PR DESCRIPTION
**Pull request by igitur**
Fixes #675

### DIFF
--- a/ClosedXML/Utils/ColorStringParser.cs
+++ b/ClosedXML/Utils/ColorStringParser.cs
@@ -1,0 +1,22 @@
+﻿using System.Drawing;
+using System.Globalization;
+
+namespace ClosedXML.Utils
+{
+    internal static class ColorStringParser
+    {
+        public static Color ParseFromHtml(string htmlColor)
+        {
+            try
+            {
+                return ColorTranslator.FromHtml(htmlColor);
+            }
+            catch
+            {
+                // https://github.com/ClosedXML/ClosedXML/issues/675
+                // When regional settings list separator is # , the standard ColorTranslator.FromHtml fails
+                return Color.FromArgb(int.Parse(htmlColor.Replace("#", ""), NumberStyles.AllowHexSpecifier));
+            }
+        }
+    }
+}
--- a/ClosedXML/ClosedXML.csproj
+++ b/ClosedXML/ClosedXML.csproj
@@ -111,6 +111,7 @@
     <Compile Include="Extensions\GuidExtensions.cs" />
     <Compile Include="Extensions\OpenXmlPartContainerExtensions.cs" />
     <Compile Include="Extensions\StringExtensions.cs" />
+    <Compile Include="Utils\ColorStringParser.cs" />
     <Compile Include="Utils\GraphicsUtils.cs" />
     <Compile Include="Utils\OpenXmlHelper.cs" />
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: ClosedXML/Utils/ColorStringParser.cs [hunk 1]
```diff
@@ -0,0 +1,22 @@
...
+        {
+            try
+            {
+                return ColorTranslator.FromHtml(htmlColor);
+            }
+            catch
+            {
+                // https://github.com/ClosedXML/ClosedXML/issues/675
+                // When regional settings list separator is # , the standard ColorTranslator.FromHtml fails
+                return Color.FromArgb(int.Parse(htmlColor.Replace("#", ""), NumberStyles.AllowHexSpecifier));
```
Pankraty: This would give a wrong result with FF00FF (Alpha will be 0 instead of 255, that is used by ColorTranslator by default)  or with FFF (in HTML it is equal to FFFFFFFF). I doubt this will be an issue in ClosedXML since it gets values stored in a file. Just FYI.

igitur: Yeah, you're right, but I'm not going to spend too much time on such an edge case. Merging this as is.

Issue #667, opened 2018-01-24
Calculation engine result caching


igitur:
cc @Pankraty 

Some issues, e.g. #662 , are about the calculation engine being slow. This could be solved by
- The calculation engine caching a cell's formula's result
- The workbook directly loading and using the cached values that are stored in the file
- The calculation engine implementing the calc chain (although I'm not convinced that the added complexity has any benefit).

My suggestion to fix this is:
- Add optional load parameters to the `XLWorkbook` constructor. Similar to the `SaveOptions` structure that's used by `XLWorkbook.Save()`.
- Add internals to `XLCell` for cached value (or use the existing `ValueCached`?) and a field/property for the date and time the formula was last calculated.
- When cell's result is calculated, store it in the cache field and set date and time property to DateTime.Now.
- When a cell value changes, invalidate all dependent cells - we should decide on whether to invalidate the whole dependency tree onwards.
- Add `IXLWorkbook.RecalculateAllFormulas` and `IXLWorksheet.RecalculateAllFormulas` methods to invalidate and recalculate all cells in applicable scope.
- Add new property in `SaveOptions` to optionally save cached results. Maybe replace the `EvaluateFormulas` property that's already in `SaveOptions`?

More to add?

Pankraty:
Instead of DateTime.Now I would suggest using incremental counter, defined at workbook level. 

- When cells or ranges are modified the counter is increased,
- In XLCell the counter value is stored. When the cell value is obtained, its counter is compared with the workbook state. If they are not equal cell value has to be recalculated. 

With this approach, each cell will be recalculated no more than once after editing, even if the formulas are very complicated.

igitur:
Is that not basically the same approach, except:
- When using a counter, one increments by 1
- When using DateTime, one increments by time passed since last change.

Pankraty:
They are similar indeed, though there is a difference:

- Relying on ```DateTime.Now``` is more fragile since there may be changes performed at the very same moment (multithreading?)
- Comparision of ```int```s (or ```long```s) is performed a few times faster,
- Not mentioning such rare but unpleasant and hard-to-reproduce things like system time changes performed during work (time zone change, time synchronization with NTP server, daylight saving time ending, leap seconds, etc). Time is a complex matter )

Or am I overcomplicating?

igitur:
Valid points, although `ClosedXML` definitely isn't thread-safe yet. Let's assume we continue with a `long` field.

[...]

PR #679, opened 2018-02-04
Implemented support of storing cell calculated values

Pankraty:
This PR implements requirements from #667. 

The idea is the following:

1. ```XLWorkbook``` has a property RecalculationCounter that is increased on every modification (either cell value or cell formula is modified).
2. ```XLCell``` has a few new properties:
- ```ValueCalculated``` - stores the evaluated value of cell formula.
- ```EvaluatedAtVersion``` - stores the value of workbook's ```RecalculationCounte```r that it had at the moment of cell last evaluation. If this value equals the current value of ```RecalculationCounter``` it indicates that ```ValueCalculated``` stores correct value and no re-evaluation has to be performed.
- ```EditedAtVersion``` - stores the value of workbook's ```RecalculationCounter``` that it had at the moment of cell last modification (value or formula). If this value is greater than ```EvaluatedAtVersion``` then cell needs re-evaluation, as well as all dependent cells do.
- ```RecalculationNeeded``` - getter of this property checks if the cell needs to be re-evaluated. It can be true if the cell was modified from the moment of previous evaluation (```EditedAtVersion > EvaluatedAtVersion```) or if any of the cell the current one depends on requires re-evaluation.
3. ```XLCalcEngine``` is extended with methods for enumerating ranges and cells that affect the specified expression directly. To get the desired ranges it simply calls a ```Parse``` method and collects all CellRangeReferences that are included in the expression. To avoid duplication two custom equality comparers (XLRangeAddressComparer and XLAddressComparer) has been introduced.

With these changes, the calculation becomes much more efficient. And the modification of a single cell causes the deferred re-evaluation of dependent cells, leaving the rest untouched.

I covered many scenarios with tests, and ready to give additional comments if needed.

#### Questions:
- Is there a blog post? — No
- Does the knowledge base need an update? — Yes, I suppose so
- Does this add new (C#) dependencies? — No


igitur:
Please rebase on latest `develop`.

igitur:
What's the pros and cons of changing the `RecalculationCounter` from type `long` to type `DateTime`. I know we've briefly chatted about this before. I think I prefer `DateTime`. It would work the same, except that it has the added benefit of actually showing you when the calculations where last redone. A normal `long` counter can't do that. What do you think

vbjay:
Why not use long but make it store ticks?  Convert to datetime if desired?

vbjay:
Heck, we could have a readonly property that is the datetime based off the
RecalculationCounter value.  Could even be nullable to say a recalc wasn't
done yet.

Pankraty:
I don't mind adding ```DateTime``` property on a workbook or a worksheet level, along with the numeric counter. But using it for determining if the cell needs re-evaluation looks like a time bomb for me, it would strike eventually. 
I have not much to add to my previous reasoning (https://github.com/ClosedXML/ClosedXML/issues/667#issuecomment-360144531). To put it shortly, getting current time value is slower, harder to deal with in multithreading computations; also time sometimes goes backward making a reproduction of some bugs extremely hard.

vbjay:
Good points.  A counter solves those problems and if we use interlocked, we
can prevent thread issues.

Pankraty:
Rebased

Pankraty:
For not spreading a discussion among too many PRs I'll proceed here.
I implemented an alternative approach using ```DateTime``` instead of ```long``` counter and I must admit that some of my concerns were groundless. It does not seem to be any slower and using ```DateTime.UtcNow``` protects us from arbitrary timezone changes and switches to daylight saving time.
The other important issue still exists though. I have experimented a bit with the following code:
```
{

    ws.Cell(1, 1).Value = 1;
    ws.Cell(1, 2).FormulaA1 = "A1*10";
    Console.WriteLine("{2} B1 value = {0}, expected {1}", ws.Cell(1, 2).Value, 10, DateTime.UtcNow);
    Console.WriteLine("Pause...");
    Console.ReadKey();

    ws.Cell(1, 1).Value = 2;
    Console.WriteLine("{2} B1 value = {0}, expected {1}", ws.Cell(1, 2).Value, 20, DateTime.UtcNow);
}
```
During the pause I changed system time settings or let the system to syncronize time with NTP server. The result is quite predictable:
```
4/18/2018 5:28:10 AM B1 value = 10, expected 10
Pause...
4/18/2018 5:26:19 AM B1 value = 10, expected 20
```

What is interesting is that 23 tests running fine on a local machine fail at AppVeyor. Do I miss something essential or time flows strangely at the build server?

vbjay:
Read the remarks.
https://msdn.microsoft.com/en-us/library/system.datetime.utcnow(v=vs.110).aspx

ClosedXML/Excel/Cells/IXLCell.cs
```diff
@@ -310,8 +310,31 @@ public interface IXLCell
 
         IXLCell CopyTo(String target);
 
+        /// <summary>
```
igitur: Ok, so for this interface I have some other ideas. 

Firstly, for a user, I think it is going to be confusing to have both `ValueCached` and `ValueCalculated`. I think we should have only 1 and I think ideally it should return `Object`. That will be a breaking change, so maybe we should phase it in with keeping both for now, and marking one as obsolete. We'll need a property name that correctly describes both. Maybe `Object CachedValue {get;}` is OK.

ClosedXML/Excel/Cells/IXLCell.cs
```diff
@@ -310,8 +310,31 @@ public interface IXLCell
...
+        /// Calculated value of cell formula. Is used for decreasing number of computations perfromed.
+        /// May hold invalid value when <see cref="RecalculationNeeded"/> flag is True.
+        /// </summary>
+        Object ValueCalculated { get; }
+
+        /// <summary>
+        /// Flag indicating that previously calculated cell value may be not valid anymore and has to be re-evaluated.
+        /// </summary>
+        Boolean RecalculationNeeded { get; }
```
igitur: Please change to `NeedsRecalculation`. I prefer `VerbObject` format (active voice) for property names, instead of passive voice.

ClosedXML/Excel/Cells/IXLCell.cs
```diff
@@ -310,8 +310,31 @@ public interface IXLCell
...
+
+        /// <summary>
+        /// Perform an evaluation of cell formula. If cell does not contain formula nothing happens, if cell does not need
+        /// recalculation (<see cref="RecalculationNeeded"/> is False) nothing happens either, unless <paramref name="force"/> flag is specified.
+        /// Otherwise recalculation is perfomed, result value is preserved in <see cref="ValueCalculated"/> and returned.
+        /// </summary>
+        /// <param name="force">Flag indicating whether a recalculation must be performed even is cell does not need it.</param>
+        /// <returns>Null if cell does not contain a formula. Calculated value otherwise.</returns>
+        object Evaluate(bool force = false);
```
igitur: I don't think this property is necessary. I would drop this and rather have a `void InvalidateFormula()` method. Then, only when the cell's value is requested, is the value calculated again. `Evaluate` can stay internal. Agree?

ClosedXML/Excel/XLWorkbook.cs
```diff
@@ -119,7 +119,22 @@ public static XLWorkbook OpenFromTemplate(String path)
...
+        /// Counter increasing at workbook data change. Serves to determine if the cell formula
+        /// has to be recalculated.
+        /// </summary>
+        internal long RecalculationCounter { get; private set; }
+
+        /// <summary>
+        /// Notify that workbook data has been changed which means that cached formula values
+        /// need to be re-evaluated.
+        /// </summary>
+        internal void NotifyRecalculationNeeded()
```
igitur: If I understand this correctly, I think we should rename to `InvalidateFormulas`.

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -1199,6 +1251,66 @@ public IXLCell AddToNamed(string rangeName, XLScope scope, string comment)
...
+        private IEnumerable<XLCell> GetAffectingCells()
+        {
+            return Worksheet.CalcEngine.GetAffectingCells(_formulaA1).Cast<XLCell>();
+        }
+
+        /// <summary>
+        /// The value of <see cref="XLWorkbook.RecalculationCounter"/> that workbook had at the moment of cell last modification.
+        /// If this value is greater than <see cref="EvaluatedAtVersion"/> then cell needs re-evaluation, as well as all dependent cells do.
+        /// </summary>
+        internal long EditedAtVersion { get; private set; }
```
igitur: Rather `ModifiedAtVersion` ?

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -362,92 +361,144 @@ public string GetFormattedString()
...
+        /// <summary>
+        /// Perform an evaluation of cell formula. If cell does not contain formula nothing happens, if cell does not need
+        /// recalculation (<see cref="NeedsRecalculation"/> is False) nothing happens either, unless <paramref name="force"/> flag is specified.
+        /// Otherwise recalculation is perfomed, result value is preserved in <see cref="CachedValue"/> and returned.
+        /// </summary>
+        /// <param name="force">Flag indicating whether a recalculation must be performed even is cell does not need it.</param>
+        /// <returns>Null if cell does not contain a formula. Calculated value otherwise.</returns>
+        public object Evaluate(bool force = false)
+        {
+            if (NeedsRecalculation || force)
```
igitur: `if (force || NeedsRecalculation)` be better to avoid `NeedsRecalculation` to be evaluated when `force == true`.

ClosedXML/Excel/CalcEngine/XLCalcEngine.cs
```diff
@@ -23,6 +23,50 @@ public XLCalcEngine(IXLWorksheet ws) : this(ws.Workbook)
...
+                    visitedRanges.Add(range.RangeAddress);
+                    yield return range;
+                }
+            }
+        }
+
+        public IEnumerable<IXLCell> GetAffectingCells(string expression)
+        {
+            if (String.IsNullOrWhiteSpace(expression) && String.IsNullOrEmpty(expression))
+                yield break;
```
igitur: `yield break` isn't required.

ClosedXML/Excel/CalcEngine/XLCalcEngine.cs
```diff
@@ -23,6 +23,50 @@ public XLCalcEngine(IXLWorksheet ws) : this(ws.Workbook)
...
+            {
+                if (!visitedRanges.Contains(range.RangeAddress))
+                {
+                    visitedRanges.Add(range.RangeAddress);
+                    yield return range;
+                }
+            }
+        }
+
+        public IEnumerable<IXLCell> GetAffectingCells(string expression)
```
igitur: Please rename to `GetPrecedentCells`.

By the way, we might want to expose this method publicly on the API. Could be useful. Leave for later.

ClosedXML/Excel/CalcEngine/XLCalcEngine.cs
```diff
@@ -49,66 +93,21 @@ public override object GetExternalObject(string identifier)
...
-                    return cell.Value;
-                else
-                    return new XLCalcEngine(cell.Worksheet).Evaluate(f);
-            }
-            finally
-            {
-                _evaluating = false;
-            }
+            var res = new CellRangeReference(range, this);
+            if (_cellRanges != null)
```
igitur: `_cellRanges?.Add(res.Range);`

ClosedXML/Excel/CalcEngine/XLCalcEngine.cs
```diff
@@ -23,6 +23,50 @@ public XLCalcEngine(IXLWorksheet ws) : this(ws.Workbook)
...
         }
 
+        private IList<IXLRange> _cellRanges;
+
+        /// <summary>
+        /// Get a collection of cell ranges included into the expression. Order is not preserved.
+        /// </summary>
+        /// <param name="expression">Formula to parse.</param>
+        /// <returns>Collection of ranges included into the expression.</returns>
+        public IEnumerable<IXLRange> GetAffectingRanges(string expression)
```
igitur: Rename to `GetPrecedentRanges`. I'm trying to align with Excel:

image

ClosedXML/Excel/CalcEngine/XLCalcEngine.cs
```diff
@@ -23,6 +23,50 @@ public XLCalcEngine(IXLWorksheet ws) : this(ws.Workbook)
...
+                {
+                    visitedRanges.Add(range.RangeAddress);
+                    yield return range;
+                }
+            }
+        }
+
+        public IEnumerable<IXLCell> GetAffectingCells(string expression)
+        {
+            if (String.IsNullOrWhiteSpace(expression) && String.IsNullOrEmpty(expression))
```
Pankraty: This condition is really weird, isn't it?

igitur: Sorry, I don't understand. What I meant was that you could just do this:
```c#
if (!String.IsNullOrWhiteSpace(expression) || !String.IsNullOrEmpty(expression))
{
    var ranges = GetAffectingRanges(expression);
    var visitedCells = new HashSet<IXLAddress>(new XLAddressComparer(true));
    var cells = ranges.SelectMany(range => range.Cells()).Distinct();
    foreach (var cell in cells)
    {
        if (!visitedCells.Contains(cell.Address))
        {
            visitedCells.Add(cell.Address);
            yield return cell;
        }
    }
}
```

igitur: Oh, do you mean we could just do `if (!String.IsNullOrWhiteSpace(expression))` ?

Pankraty: Exactly.

igitur: Ok, so then that method's body could be just:
```c#
if (!String.IsNullOrWhiteSpace(expression))
{
    var ranges = GetAffectingRanges(expression);
    var visitedCells = new HashSet<IXLAddress>(new XLAddressComparer(true));
    var cells = ranges.SelectMany(range => range.Cells()).Distinct();
    foreach (var cell in cells)
    {
        if (!visitedCells.Contains(cell.Address))
        {
            visitedCells.Add(cell.Address);
            yield return cell;
        }
    }
}
```

igitur:
Great work, again! Thanks.

PR #681, opened 2018-02-06
FIX for Clear() for cells sets the value to "", not null #544

### LINKED ISSUE #544: Clear() for cells sets the value to "", not null

- [x] Bug
Version of ClosedXML: 0.89.0

When you set a cell to null, the cell value is set to a blank string when you read it back. 

Excel generally stores empty cells as null values in the Excel files, but ClosedXML is unable to either read the null values as nulls nor write empty cells as nulls on writing.

I am in the process of re-writing our Excel reading/writing library (basically a port of CsvHelper to work with Excel) from the ComponentOne C1.Excel library to ClosedXML. One of the features we supported in our library was the ability to handle empty cells and provide default values when records are loaded, but for this to work it relies on actually reading an empty cell from the Excel file as a null. For our testing I need the ability to write empty cells as nulls so I can read them back that way as well. 

It would appear that an architectural decision was made to not support empty cells as null but rather as simply blank cells, but the problem with this approach is you have data loss because you cannot tell when you load an Excel file if the cell was actually empty, or if someone literally put a blank string in there? 

An empty cell and a blank cell are actually two different things.

I doubt it.

```c#
        [TestMethod]
        public void NullTest()
        {
            using (var stream = new MemoryStream()) {
                // Create some test data to parse
                using (var book = new XLWorkbook()) {
                    var sheet = book.AddWorksheet("Sheet 1");
                    sheet.Cell(1, 1).Value = "test1";
                    // Cell(1,2) is empty
                    sheet.Cell(1, 3).Value = null;
                    book.SaveAs(stream);
                }

                // Now load it back
                stream.Position = 0;
                using (var book = new XLWorkbook(stream)) {
                    var sheet = book.Worksheets.Worksheet(1);
                    var result =  sheet.Cell(1, 1).Value;
                    Assert.AreEqual("test1", result);
                    result = sheet.Cell(1, 2).Value;
                    Assert.AreEqual(null, result);  // Should be null!
                    result = sheet.Cell(1, 3).Value;
                    Assert.AreEqual(null, result);  // Should be null!
                }
            }
        }
```

WilliamsKobie:
_cellValue variable was set to empty string "" rather than null in the XLCell class.

Fixes #544

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -67,7 +67,7 @@ namespace ClosedXML.Excel
 
         private readonly XLWorksheet _worksheet;
 
-        internal string _cellValue = String.Empty;
+        internal string _cellValue = null;
 
         private XLComment _comment;
         internal XLDataType _dataType;
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -67,6 +67,7 @@ namespace ClosedXML.Excel
 
         private readonly XLWorksheet _worksheet;
 
+
         internal string _cellValue = null;
 
         private XLComment _comment;
--- a/ClosedXML_Sandbox/Program.cs
+++ b/ClosedXML_Sandbox/Program.cs
@@ -1,4 +1,6 @@
+using ClosedXML.Excel;
 using System;
+using System.IO;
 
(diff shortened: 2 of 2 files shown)


Pankraty:
Please, choose another branch to merge in (you specified 'master', while it should be  'develop') and make sure your changes (and none else) are included into the PR.

igitur:
Closing this as per discussion in #544

PR #696, opened 2018-02-21
Eliminated significant redundant iterations.

electricessence:
#### What's this PR do?
There were significant redundant uses of Linq where every .Count(), .Average(), .Sum() caused a repeated iteration and filtering of the expressions.  These changes adopt a lazy approach to acquiring the numerical values and avoid repeated .Count() calls etc.

It's one file.  Easy to review.  The key is line 14 `private IReadOnlyList<double> _numericValues;`

Any calls that use Tally.

Note: there are plenty of other places in the ClosedXML code where this type of inefficiency is in place, but this was the one I found and tackled.  Others may want to review any .Count() calls and property call .ToList() where it makes sense.

IMPORTANT: Calling `.Count() > 0` is not the same performance as calling `.Any()` (or `.Count != 0` when a list.).  Although possibly insignificant for small sets, it's simply bad practice and should be changed.

Possibly older code, but I discovered instances of this in:

- XLWorkbook.cs
- XLWorkbook_Load.cs

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Functions/Tally.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/Tally.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 
 namespace ClosedXML.Excel.CalcEngine
 {
@@ -9,6 +10,9 @@ namespace ClosedXML.Excel.CalcEngine
     {
         private readonly List<object> _list = new List<object>();
         private readonly bool NumbersOnly;
+
+        private IReadOnlyList<double> _numericValues;
+
 
         public Tally()
             : this(false)
@@ -32,7 +36,7 @@ namespace ClosedXML.Excel.CalcEngine
                 }
             }
 
-            this.NumbersOnly = numbersOnly;
+            NumbersOnly = numbersOnly;
         }
 
         public void Add(Expression e)
@@ -45,6 +49,7 @@ namespace ClosedXML.Excel.CalcEngine
                 {
                     _list.Add(value);
                 }
+                _numericValues = null;
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


electricessence:
I did compile the project successfully.  The interface change from IEnumerable<double> to IReadOnlyList<double> could be an issue?

ClosedXML/Excel/CalcEngine/Functions/Tally.cs
```diff
@@ -56,85 +61,87 @@ public void Add(Expression e)
...
+                {
+                    if (double.TryParse(value.ToString(), out double tmp))
+                        yield return tmp;
+                }
                 else
                 {
                     foreach (var v in vEnumerable)
                     {
-                        if (Double.TryParse(v.ToString(), out tmp))
+                        if (double.TryParse(v.ToString(), out double tmp))
```
Pankraty: ```out var``` is a feature of C# 7 and ClosedXML uses C# 6. I think that's why the build failed

ClosedXML/Excel/CalcEngine/Functions/Tally.cs
```diff
@@ -56,85 +61,87 @@ public void Add(Expression e)
...
                             yield return tmp;
                         break;
                     }
                 }
             }
         }
 
+        public IReadOnlyList<double> NumericValues()
```
Pankraty: IReadOnlyList<T> was introduced in .NET 4.5, it does not seem to compile with .NET 4.0.

ClosedXML/Excel/CalcEngine/Functions/Tally.cs
```diff
@@ -56,85 +61,87 @@ public void Add(Expression e)
...
-            var retVal = new List<double>();
             foreach (var value in _list)
             {
-                Double tmp;
                 var vEnumerable = value as IEnumerable;
-                if (vEnumerable == null && Double.TryParse(value.ToString(), out tmp))
-                    yield return tmp;
+                if (vEnumerable == null)
+                {
+                    if (double.TryParse(value.ToString(), out double tmp))
```
Pankraty: I had issues with this part of code too. On heavy formulas performing ```double.TryParse(doubleValue.ToString(), out tmp)``` took 50% of entire CPU usage being absolutely unnecessary.  See https://github.com/ClosedXML/ClosedXML/pull/679/files#diff-dcbb29e3ee481081b1cd4cd877f6a3a0R84

ClosedXML/Excel/CalcEngine/Functions/Tally.cs
electricessence: Ok. I'll make an update.

ClosedXML/Excel/CalcEngine/Functions/Tally.cs
electricessence: K.  Will update.

electricessence:
Ok.  Fixes in compatibility made.

ClosedXML/Excel/CalcEngine/Functions/Tally.cs
igitur: I've fixed the project Release configuration to explicitly use C#6. See #697

igitur:
Thanks. Good catch. Do you have any idea of the effect of the performance gain?

electricessence:
@igitur: It simply depends on the usage.  Iterating 5 times over something that only has 5 entries is barely noticeable.  But a 100 entries 5 times will have an impact. The way the code was written (and then never revisited) there was a huge potential for redundant calculations.   There are quite a few micro optimizations that could be applied here to improve it even further.  But the core of the problem was that it was iterating the source collection (discovering doubles and converting/casting) _multiple times_ for all calculations.

igitur:
Ok, thought you encountered it through a real world use case.

electricessence:
@igitur I think someone else here commented that they did encounter a problem with it real world.

Issue #607, opened 2017-11-28
Potential Memory Leak Issues On Second Run


Tongrer:
## Read and complete the full issue template

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

e.g. 0.90.0

**What is the current behavior?**

When opening, editing and saving attached document, the first run and save uses 120mb ram and reduces filesize by 40kb, second run through hits OOM exceptions during save. In addition occasionally when opening excel after ClosedXML save it says it needs to repair and deletes "Drawing Shapes"

**What is the expected behavior or new feature?**

Complete this.

**Did this work in previous versions of our tool?  Which versions?**

Unknown

## Reproducibility

**Code to reproduce problem:**
`XLWorkbook xlwb = new XLWorkbook(attachedFile);
 IXLWorksheet xlChecklist = xlwb.Worksheet("CHECKLIST");
xlwb.Save();
xlwb.Dispose();
GC.Collect();
GC.WaitForPendingFinalizers();
GC.Collect();`
[VB Free.xlsx](https://github.com/ClosedXML/ClosedXML/files/1509658/VB.Free.xlsx)

- [x ] I attached a sample spreadsheet.  (You can drag files on to this issue)

igitur:
You describe opening, editing and saving a file twice, but your code sample does this only once. So obviously I won't be able to reproduce the issue.

Tongrer:
My bad, I have now made it run twice

igitur:
Does this make a difference:

```c#
for(int i = 1; i <=2; i++)
{
    using (var xlwb = new XLWorkbook(attachedFile))
    {
        IXLWorksheet xlChecklist = xlwb.Worksheet("CHECKLIST");
        xlwb.Save();
    }
    GC.Collect();
    GC.WaitForPendingFinalizers();
    GC.Collect();
}
```

Tongrer:
Unfortunately not as per attached picture.

![image](https://user-images.githubusercontent.com/27585041/33320526-78683ccc-d43a-11e7-8144-caf1c16d34e1.png)

igitur:
If you're not planning to insert or delete columns or rows, you can disable event tracking. Maybe that helps. See https://github.com/ClosedXML/ClosedXML/wiki/Turning-off-events

Tongrer:
[...]

PR #703, opened 2018-02-26
Create cell address on demand

Pankraty:
Profiling the application I discovered that XLAddress instances attached to each XLCell instance take a significant amount of memory. Not huge, of course, but still redundant.

If XLCell stores its own rowNumber, columnNumber, etc., it can create a new instance of XLAddress on demand, reducing overall memory consumption.

**Before:**
image

**After:**
image

Both screenshots were taken during PerformanceRunner.RunInsertTable operation, right before EmulateSave is called. As it may be observed, XLAddress instances consumed about 7 MB. 
This optimization does not seem to affect performance.

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -107,21 +107,21 @@ namespace ClosedXML.Excel
 
 
         private int _rowNumber;
-        private int _colNumber;
+        private int _columnNumber;
         private bool _fixedRow;
         private bool _fixedCol;
         public XLAddress Address
         {
             get
             {
-                return new XLAddress(_worksheet, _rowNumber, _colNumber, _fixedRow, _fixedCol);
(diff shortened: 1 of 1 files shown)


Pankraty:
>Here's one (but not sure if it will really be affected by this change):

This particular benchmark does not cause massive XLCell creation, thus is not affected by the PR.

igitur:
Ok, sorry, bad example then. Do you have something proper to test with?

Pankraty:
I've made some changes to eliminate the number of XLAddress creation and tested opening and saving the workbook from the #607 (not in the loop, however). From the point of performance, there was no difference at all - time it takes to run the example is about 7.1-7.2 sec no matter what approach is used. But the memory consumption varies:
image
image

And I cannot deny that the proposed solution breaks the encapsulation principle, so it is completely up to you whether or not reducing memory usage worths it.

PR #709, opened 2018-03-03
Styles handling redesign: support styles reusing to lower the memory consumption

Pankraty:
ClosedXML handles cell styles in such a way that each row, each column, and each cell have its own copy of XLStyle, that holds an XLAlignment instance, XLBorder instance, XLFill instance, XLFont instance, XLNumberFormat instance and XLProtection instance. This causes a huge memory consumption, especially when hundreds of thousands of cells are processed. 

Obviously, there hardly ever be hundreds of thousands unique styles in the workbook, so it is natural to share the same instances of style classes among different cells (rows, columns, etc.). But there are two important problems that have to be solved when implementing caching the styles in some way.

1. Say, we have 4 cells, sharing the same style. Then we modify the background color of one cell and expect that this cell will have s new style, while the rest of them remain unchanged. The difficulty is that we must preserve the existing API, which allows doing the following: ```ws.Cell("A1").Style.Fill.BackgroundColor = XLColor.Red```. If the ```Style``` property would refer to the shared style then all cells having the same style would be colored as well, which is unacceptable.

2. Say, we have 4 cells each having its own style. Then we modify the font color of the entire range's style: ```ws.Range("A1:B2").Style.Font.Color = XLColor.Red``` - and expect that other properties of cell styles (background, alignment, protection, and so on) will remain the same as they were, and only font color would change. Hence, the most obvious way - apply the style of the range to each cell in the range - would produce an incorrect result.

----
Keeping this in mind I propose the following solution
1. Implement a generic thread-safe repository that holds weak references to some entities
2. Implement a concrete repository type for each style element that has to be shared among different styles, and for styles themselves.
3. For each style element (namely, Alignment, Border, Fill, Font, NumberFormat, Protection, Color and Style itself) create a corresponding structure that serves as a key in the repository: XLAlignmentKey, XLBorderKey, etc. Since they are value-types they are immutable, meaning that any change of the existing ```Key``` would produce another copy of it. 
4. Create a read-only version for each style element: XLAlignmentValue, XLBorderValue, etc. These classes are stored in the repository and are shared among different stylized entities. Each entity holds only a reference to the particular value in the repository.
5. Modify existing classes (XLStyle, XLAlignment, XLBorder, ....) in such a way that precisely preserves the existing API, but makes these classes mere wrappers around cached values. The modification of their properties are propagated up to the stylized entity and then down to child entities (i.e. if we change ```Fill.BackgroundColor``` of the worksheet the modification is passed from ```XLFill``` to containing ```XLStyle```, then to ```XLWorksheet```, then to each row's ```Style```, each column's ```Style```, and then to each existing cell's ```Style```).

With everything is done I only had to modify a single existing test which indicates that public API has not been changed, and consumers' code will hardly be affected. The memory consumption became significantly lower.

*Before:*
image

*After:*
image


igitur:
Great work. I want to understand it fully before merging.

Pankraty:
The last commit addresses the styles duplication in CurrentRowColumn.xlsx file. There are only 4 unique styles but the sample file contains 9.

image

igitur:
I notice that this PR removes a lot of the differential formats in the reference files. Can you confirm that this is the intended action?

Example: in `AddingRowToTables.xlsx` : `styles.xml` (left: before your PR, right: after your PR)
image

Pankraty:
As far as understand the idea of differential formats, they are supposed to store only the difference with default values. Thus, sections identical to defaults may be safely omitted. I can't say this was done intentionally from the beginning, but after I discovered this "side-effect" of changing the approach I assumed it is nice to generate slightly smaller files.

igitur:
Ok. Would you please ensure that no empty tags are written to the file, e.g. the `<font />` and `<border />`?

Pankraty:
Yeah, that makes sense. Will do.

ClosedXML/Excel/Style/XLBorder.cs
```diff
@@ -1,80 +1,97 @@
...
         }
 
-        public XLBorder(IXLStylized container, IXLBorder defaultBorder, Boolean useDefaultModify = true)
+        #region Constructors
+        /// <summary>
+        /// Create an instance of XLBorder initializing it with the specified value.
+        /// </summary>
+        /// <param name="style">Style to attach the new instance to.</param>
+        /// <param name="value">Style value to use.</param>
+        public XLBorder(IXLStylized container, XLStyle style, XLBorderValue value)
```
igitur: `warning CS1573: Parameter 'container' has no matching param tag in the XML comment for 'XLBorder.XLBorder(IXLStylized, XLStyle, XLBorderValue)' (but other parameters do)

Pankraty:
I formatted all classes in the Styles namespace in a separate commit so it would be easier to merge (in case of conflicts this commit may be skipped, and auto formatting may be applied once again).

Pankraty:
BTW, this fixes #607.

**Before:** (I gave up before the program finished)
image

**After:**
image

igitur:
Is this ready for merge?

Pankraty:
Yes

ClosedXML/Excel/Style/XLStylizedBase.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                Children.ForEach(child => child.SetStyle(StyleValue, true));
+            }
+        }
+
+        private static ReferenceEqualityComparer<XLStyleValue> _comparer = new ReferenceEqualityComparer<XLStyleValue>();
+
+        public void ModifyStyle(Func<XLStyleKey, XLStyleKey> modification)
+        {
+            var children = GetChildrenRecursively(this);
```
igitur: How about only:

```c#
var children = GetChildrenRecursively(this)
    .GroupBy(child => child.StyleValue, _comparer);
```

ClosedXML/Excel/Style/XLFillKey.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+
+namespace ClosedXML.Excel
+{
+    public struct XLFillKey : IEquatable<XLFillKey>
```
igitur: Add `[DebuggerDisplay(...)]` attributes on the new key structs. Makes it easier for debugging.

ClosedXML/Excel/Style/XLFill.cs
```diff
@@ -5,142 +5,163 @@ namespace ClosedXML.Excel
...
         }
 
-        #region Properties
+        public XLFill(XLStyle style = null, IXLFill d = null) : this(style, GenerateKey(d))
+        {
+        }
+
+        #endregion Constructors
+
+        private void Modify(Func<XLFillKey, XLFillKey> modification)
```
igitur: I notice that the `Modify` methods would recursively call the `modification` method, and it leads to unnecessary, multiple calls of `modification`. Example: `ws.Range("A1:A10").Style.Fill.BackgroundColor = XLColor.Red;` would make `XLFill.cs:125` ( `{ k.PatternType = value; return k; }` ) be called multiple times.

ClosedXML/Excel/Style/XLFill.cs
```diff
@@ -5,142 +5,163 @@ namespace ClosedXML.Excel
...
-            get { return _backgroundColor; }
+            get { return XLColor.FromKey(Key.BackgroundColor); }
             set
             {
-                SetStyleChanged();
-                if (_container != null && !_container.UpdatingStyle)
-                    _container.Styles.ForEach(s => s.Fill.BackgroundColor = value);
-                else
+                if (value == null)
+                    throw new ArgumentNullException("Color cannot be null");
```
igitur: Please change all these to e.g. `throw new ArgumentNullException(nameof(value), "Color cannot be null")` . The overload you're currently using is `ArgumentNullException(string paramName)`, which isn't supposed to include a message, only paramname.

ClosedXML/Excel/Style/IXLStylized.cs
```diff
@@ -6,11 +6,18 @@ namespace ClosedXML.Excel
...
         IXLStyle InnerStyle { get; set; }
+
         IXLRanges RangesUsed { get; }
-        Boolean StyleChanged { get; set; }
-        //Boolean IsDefault { get; set; }
+
+        /// <summary>
+        /// Immutable style
+        /// </summary>
+        XLStyleValue StyleValue { get; }
```
igitur: Can we not keep all the new value and key classes/structs out of the API and internal to ClosedXML? Surely the user should not be concerned wit the internal implementation?

ClosedXML_Sandbox/Program.cs
```diff
@@ -18,6 +18,10 @@ private static void Main(string[] args)
             Console.WriteLine();
 #endif
 
+            Console.WriteLine("Running {0}", nameof(PerformanceRunner.RunInsertTableWithStyles));
```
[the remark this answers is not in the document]
Pankraty: No problem. It may be excluded at all - I used this benchmark to make sure the performance didn't get too worse. There is no need to preserve this test permanently.

ClosedXML/Excel/Style/XLFillKey.cs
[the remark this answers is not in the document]
Pankraty: It would be convenient, indeed, but would encourage a quite inefficient usage. ```XLxxxxKey``` are structs, which means this code

```
var fill = new XLFillKey()
.WithBackgroundColor(a)
.WithPatternType(b)
.WithPatternColor(c);
```

will cause 4 XLFillKey instances created, instead of one. With ```ref struct``` it would be fine, but for now - I'd prefer stick to a 'classic' approach:

```
var fill = new XLFillKey
{
  BackgoundColor = a,
  PatterType = b,
  PatternColor = c
};
```

ClosedXML/Excel/Style/XLStyleValue.cs
```diff
@@ -0,0 +1,64 @@
+﻿using ClosedXML.Excel.Caching;
+
+namespace ClosedXML.Excel
+{
+    public class XLStyleValue
+    {
+        private static readonly XLStyleRepository Repository = new XLStyleRepository(key => new XLStyleValue(key));
```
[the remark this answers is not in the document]
Pankraty: Cells with identical styles will refer to the same instance of ```XLStyleValue``` no matter whether they defined in the same workbook or in different ones. But modification of one cell never changes shared ```XLStyleValue``` instance - instead, the modified cell will refer to another instance of ```XLStyleValue```. So having a global repository a) is absolutely safe, b) simplifies cross-workbook operations (we don't have to bother synchronizing styles), c) reduces memory consumption (we don't store identical instances in different repositories).
There is a flaw, however. After we closed all the workbooks the repository is not being cleared, but normally there should be only a few kilobytes. And I was going to add an automatic clearing of global repositories someday anyway.

ClosedXML/Excel/Caching/XLRepositoryBase.cs
```diff
@@ -0,0 +1,118 @@
...
+            _storage = new ConcurrentDictionary<Tkey, WeakReference>(CONCURRENCY_LEVEL, INITIAL_CAPACITY, comparer);
+            _createNew = createNew;
+        }
+
+        public bool ContainsKey(Tkey key)
+        {
+            WeakReference cachedReference;
+            if (_storage.TryGetValue(key, out cachedReference))
+            {
+                var storedValue = cachedReference.Target as Tvalue;
```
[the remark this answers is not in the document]
Pankraty: Nope. ```cachedReference.Target``` becomes null if the garbage collector has collected the original instance. In this case, we consider the key as non-existing.

ClosedXML/Excel/Style/XLFill.cs
Pankraty: ```Modify``` is called twice, and that's so due to this code:
```
                // 4 ways of determining an "empty" color
                if (new XLFillPatternValues[] { XLFillPatternValues.None, XLFillPatternValues.Solid }.Contains(PatternType)
                    && (BackgroundColor == null
                    || !BackgroundColor.HasValue
                    || BackgroundColor == XLColor.NoColor
                    || BackgroundColor.ColorType == XLColorType.Indexed && BackgroundColor.Indexed == 64))
                {
                    PatternType = value.HasValue ? XLFillPatternValues.Solid : XLFillPatternValues.None;
                }

                Modify(k => { k.BackgroundColor = value.Key; return k; });
```

When we set a solid background, we change the pattern type too. Not a big overhead.
Or do I understand you wrong?

ClosedXML/Excel/Style/IXLStylized.cs
Pankraty: This will require performing a typecast each time we want to get a style value from ```IXLSomething``` (roughly, 50 places in code) but this is surely possible.

ClosedXML/Excel/Caching/XLRepositoryBase.cs
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    internal abstract class XLRepositoryBase<Tkey, Tvalue> : XLRepositoryBase, IXLRepository<Tkey, Tvalue>
+        where Tkey : struct, IEquatable<Tkey>
+        where Tvalue : class
+    {
+        const int CONCURRENCY_LEVEL = 4;
+        const int INITIAL_CAPACITY = 1000;
+
+        private readonly ConcurrentDictionary<Tkey, WeakReference> _storage;
```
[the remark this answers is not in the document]
Pankraty: Not yet, although I thought this can be used for propagating style changes from parent to children in parallel.
Potentially, this may allow working with multiple workbooks in independent threads (as long as they do not interact it is more or less safe, but if they share global style repositories with ordinary dictionaries the result could be unpredictable). Then, with an ordinary dictionary unit tests running in parallel would randomly fail. 
And finally, ConcurrentDictionary may be safely modified when it is enumerated, and this helped me with #766.

ClosedXML/Excel/Style/XLFillKey.cs
Pankraty: I have just tried and evaluation of expressions turned out to be very slow making debugging extremely uncomfortable. So I implemented ```ToString()``` methods instead. It works way faster, and allows using ```XLColorKey.ToString()``` inside the ```XLFillKey.ToString()``` and so on.

ClosedXML/Excel/Style/XLStylizedBase.cs
```diff
@@ -0,0 +1,116 @@
...
+        /// </summary>
+        protected abstract IEnumerable<XLStylizedBase> Children { get; }
+
+        public abstract IXLRanges RangesUsed { get; }
+
+        public abstract IEnumerable<IXLStyle> Styles { get; }
+
+        #endregion Properties
+
+        public XLStylizedBase(XLStyleValue styleValue)
```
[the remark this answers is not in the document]
Pankraty: Done

ClosedXML/Excel/Style/XLStylizedBase.cs
Pankraty: Done

ClosedXML/Excel/Style/XLFill.cs
Pankraty: Done

Pankraty:
Preliminary feedback processed, see the comments. Merge conflicts solved.

ClosedXML/Excel/Style/XLFillKey.cs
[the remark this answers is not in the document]
igitur: Yes, let's park this until we use ref structs.

ClosedXML/Excel/Style/XLStyleValue.cs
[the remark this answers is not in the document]
igitur: Is `Repository` ever cleared? Think about long-running web processes that generate and deliver Excel files. Would a static repository not be a problem? What's the pro's and con's of rather doing it inside the `XLWorkbook`?

ClosedXML/Excel/Caching/XLRepositoryBase.cs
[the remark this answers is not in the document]
igitur: Sorry, I don't understand. If `cachedReference.Target` is null, then `storedValue` will be `null` and `false` will be returned. `return cachedReference.Target is Tvalue` will do the same?

ClosedXML/Excel/Style/IXLStylized.cs
igitur: From the `Sandbox` project, the `XLFillKey` and `XLFillValue` classes are visible, because they are public. I would expect them to be internal and that the API still expose only the `IXLFill` interface. Same for the other style classes. Where would the typecast be needed?

ClosedXML/Excel/Style/XLStyleValue.cs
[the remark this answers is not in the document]
Pankraty: Entries in the repository are collected by GC, so they won't stuck there forever. Keys for dead entries are removed during the enumeration of all values in the repository. For now, such enumeration is not performed, that's true. But as I said, I am going to add auto-removing dead entries. For example, we can call a timer that will collect unused entries once per minute in a separate async task (one more pro for ```ConcurrentDictionary```).
Even without clearing the repository it cannot realistically grow infinitely. 100000 _unique_ style keys would cost approximately 10-20 MB of RAM.

As for defining repositories on a workbook level. Originally I started from this. But then faced with two unpleasant things: 
1. Some of the entities may be constructed outside of the workbook, so they cannot share their styles until they are added to some workbook. And this produces unnecessary complexity as we should always check if the workbook defined, if the current workbook is the same that was used when the style was created, etc.
2. This forces us to pass a workbook into each 'style' class constructor - i.e. when we create ```XLColor``` instance we should specify a workbook to put (or get) the instance in the proper repository.
With this, I finally abandoned the idea of defining repositories on a workbook level.

ClosedXML/Excel/Caching/XLRepositoryBase.cs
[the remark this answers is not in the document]
Pankraty: Ah, you are right.

But it just came to me that I should better define ```public bool ContainsKey(Tkey key, out Tvalue value)``` to guarantee that existing value won't be GCed right after checking for existence...
On the other hand, precautions for this case were made in ```Store``` method.

ClosedXML/Excel/ConditionalFormats/Save/XLCFEndsWithConverter.cs
```diff
@@ -9,9 +9,9 @@ public ConditionalFormattingRule Convert(IXLConditionalFormat cf, int priority,
         {
             String val = cf.Values[1].Value;
             var conditionalFormattingRule = XLCFBaseConverter.Convert(cf, priority);
-
-            if (!cf.Style.Equals(XLWorkbook.DefaultStyle))
-                conditionalFormattingRule.FormatId = (UInt32)context.DifferentialFormats[cf.Style];
+            var cfStyle = cf.Style.Value;
+            if (!cfStyle.Equals(XLWorkbook.DefaultStyle))
```
igitur: Should be ` if (!cfStyle.Equals(XLWorkbook.DefaultStyle.Value))` ?

ClosedXML/Excel/ConditionalFormats/Save/XLCFNotContainsConverter.cs
```diff
@@ -9,9 +9,9 @@ public ConditionalFormattingRule Convert(IXLConditionalFormat cf, int priority,
         {
             String val = cf.Values[1].Value;
             var conditionalFormattingRule = XLCFBaseConverter.Convert(cf, priority);
-
-            if (!cf.Style.Equals(XLWorkbook.DefaultStyle))
-                conditionalFormattingRule.FormatId = (UInt32)context.DifferentialFormats[cf.Style];
+            var cfStyle = cf.Style.Value;
+            if (!cfStyle.Equals(XLWorkbook.DefaultStyle))
```
igitur: Should be ` if (!cfStyle.Equals(XLWorkbook.DefaultStyle.Value))` ?

ClosedXML/Excel/ConditionalFormats/Save/XLCFEndsWithConverter.cs
Pankraty: Good catch, thank you!

ClosedXML/Excel/Style/XLStyleValue.cs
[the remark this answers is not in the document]
kenleese: @Pankraty in a server environment your possible async task might be problematic.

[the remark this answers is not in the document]
Pankraty: @kenleese why?

ClosedXML/Excel/Style/XLFill.cs
Pankraty: I see. That's actually a side effect of preserving an existing API. 
When we modify a pattern color of a ```XLFill``` from inside the code we can't say what exactly is going on:
**option 1**
```
var f = something.Fill;
f.PatternColor = color;
```
**option 2**
```
var s = something.Style;
s.Fill.PatternColor = color;
```
**option 3**
```
var r = something.AsRange();
r.Style.Fill.PatternColor = color;
```

All these cases must work the same way and produce the same result. So, we do the following
1. Modify the ```Key``` of current ```XLFill``` instance - in case it has been captured by some variable the user would expect it will change.
2. Modify the ```PatternColor``` of the fill of the corresponding style. ```XLStyle.Fill``` creates a new instance each time it is accessed, so it is never the same instance as on step 1. But in case Style is captured by some variable the user again will expect it to be modified,
3. Then, modify ```PatternColor``` of all the fills in the ```_container``` if it exists, and all nested containers too. In your example, they all have the same style (range and all cells in it), so this step is executed only once. But if you apply some other styles to some of the cells there will be even more repetitions.

I admit this may not seem very optimal, but I really tried to make it as simple as possible without modifying existing tests ).

ClosedXML/Excel/Style/XLStyleValue.cs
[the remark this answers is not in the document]
igitur: @Pankraty Thanks. I see the removal in the enumeration part now. BTW, your pattern is strikingly similar to https://github.com/thomaslevesque/WeakEvent/blob/master/WeakEvent/WeakEventSource.cs , which I am/was evaluating for a refactor of the events. Maybe there are some more ideas there that we can borrow... just saying.

@kenleese Didn't know you follow this project THIS closely!

ClosedXML/Excel/Caching/XLAlignmentRepository.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace ClosedXML.Excel.Caching
+{
+    internal sealed class XLAlignmentRepository : XLRepositoryBase<XLAlignmentKey, XLAlignmentValue>
+    {
+        #region constructors
+        public XLAlignmentRepository(Func<XLAlignmentKey, XLAlignmentValue> createNew) : base(createNew)
```
igitur: Just a small stylistic request, please. Put the `: base ...` on a new line. For the other classes too. It's easy to miss it if it's so far to the right.

ClosedXML/Excel/Caching/XLRepositoryBase.cs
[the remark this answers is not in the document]
igitur: Ok, let me know when you're finished with this method.

ClosedXML/Excel/Ranges/XLRangeBase.cs
```diff
@@ -841,35 +803,28 @@ public XLCell Cell(XLAddress cellAddressInRange)
             if (cell != null)
                 return cell;
 
-            Int32 styleId = GetStyleId();
-            Int32 worksheetStyleId = Worksheet.GetStyleId();
+            var styleValue = this.StyleValue;
 
-            if (styleId == worksheetStyleId)
+            if (styleValue == Worksheet.StyleValue)
```
igitur: `.Equals()` ?

ClosedXML/Excel/Ranges/XLRangeBase.cs
```diff
@@ -841,35 +803,28 @@ public XLCell Cell(XLAddress cellAddressInRange)
...
             {
                 XLRow row;
                 XLColumn column;
                 if (Worksheet.Internals.RowsCollection.TryGetValue(absRow, out row)
-                    && row.GetStyleId() != worksheetStyleId)
-                    styleId = row.GetStyleId();
+                    && row.StyleValue != Worksheet.StyleValue)
```
igitur: `.Equals()` ?

Pankraty: > Maybe XLStyleValue should override == and != ?

I was pretty sure they are implemented ) Will do right away.

Pankraty: As long as styles are shared it is OK to rely on reference comparison, but anyway, to have explicit operators will be better.

ClosedXML/Excel/Style/XLStyleValue.cs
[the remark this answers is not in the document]
kenleese: @Pankraty I apologize if I commented too soon. Perhaps I should have waited until a specific "async task" implementation might be proposed. However in general our server environment would prefer control over "background tasks" where possible. The current lack of thread safety is already problematic in a multi-threaded server environment.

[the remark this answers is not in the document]
Pankraty: Oh, there is nothing to apologize for. I understand that garbage collecting may be somewhat "advanced" (that why I didn't hurry with the concrete solution) so I appreciate any input on this matter.

igitur:
Thanks, @Pankraty , for a lot of hard work.

PR #710, opened 2018-03-04
range copy fix for conditional formats

Firedragonweb:
#### What's this PR do?
Copying cells between workbooks/Worksheets with conditional formats could never really work. Since the target range of the conditonal format was only set *after* the style, it would always reference a style in the source worksheet. In the best case, this would lead to ClosedXML being unable to save the new workbook. In the worse case, a kind of randomly selected style from the target worksheet would be used for the conditional format.
The PR is a simple change of order. Since the target range is set before, assigning the style for the conditional format correctly creates it in the target worksheet.
Copying a range with conditional formats between worksheets/workbooks.
#### Questions:
- Is there a blog post?
Nope
Nope
Nope

### DIFF
--- a/ClosedXML_Tests/Excel/ConditionalFormats/ConditionalFormatCopyTests.cs
+++ b/ClosedXML_Tests/Excel/ConditionalFormats/ConditionalFormatCopyTests.cs
@@ -11,19 +11,16 @@ namespace ClosedXML_Tests.Excel.ConditionalFormats
         public void StylesAreCreatedDuringCopy()
         {
             var wb = new XLWorkbook();
-            XLWorksheet ws = (XLWorksheet)wb.Worksheets.Add("Sheet");
+            var ws = wb.Worksheets.Add("Sheet");
+            var format = ws.Range("A1:A1").AddConditionalFormat();
+            format.WhenEquals("=" + format.Range.FirstCell().CellRight(4).Address.ToStringRelative()).Fill
+                  .SetBackgroundColor(XLColor.Blue);
 
-            SetFormat1(ws.Range("A1:A1").AddConditionalFormat());
             var wb2 = new XLWorkbook();
-            XLWorksheet ws2 = (XLWorksheet)wb2.Worksheets.Add("Sheet2");
+            var ws2 = wb2.Worksheets.Add("Sheet2");
             ws2.FirstCell().CopyFrom(ws.FirstCell());
             Assert.That(ws2.ConditionalFormats.First().Style.Fill.BackgroundColor, Is.EqualTo(XLColor.Blue)); //Added blue style
 
-        }
-
-        private static void SetFormat1(IXLConditionalFormat format)
-        {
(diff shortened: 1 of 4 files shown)


igitur:
Thanks. Good catch. Could you please include some unit tests? I'll try to find the most appropriate place to include them.

Firedragonweb:
Done :)

Pankraty:
This test does not fail without the proposed changes as well.

igitur:
Just a sidenote: actions between 2 different workbooks is still very much work in progress.

ClosedXML_Tests/Excel/ConditionalFormats/ConditionalFormatCopyTests.cs
```diff
@@ -0,0 +1,29 @@
...
+
+            SetFormat1(ws.Range("A1:A1").AddConditionalFormat());
+            var wb2 = new XLWorkbook();
+            XLWorksheet ws2 = (XLWorksheet)wb2.Worksheets.Add("Sheet2");
+            ws2.FirstCell().CopyFrom(ws.FirstCell());
+            Assert.That(ws2.ConditionalFormats.First().Style.Fill.BackgroundColor, Is.EqualTo(XLColor.Blue)); //Added blue style
+
+        }
+
+        private static void SetFormat1(IXLConditionalFormat format)
```
igitur: Separate method isn't necessary here. Please collapse this code into `StylesAreCreatedDuringCopy`.

ClosedXML_Tests/Excel/ConditionalFormats/ConditionalFormatCopyTests.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace ClosedXML_Tests.Excel.ConditionalFormats
+{
+    [TestFixture]
+    public class ConditionalFormatCopyTests
+    {
+        [Test]
+        public void StylesAreCreatedDuringCopy()
+        {
+            var wb = new XLWorkbook();
+            XLWorksheet ws = (XLWorksheet)wb.Worksheets.Add("Sheet");
```
igitur: No need to cast to `XLWorksheet`. Please keep it simple, e.g. `var ws = wb.Worksheets.Add("Sheet");`

ClosedXML_Tests/Excel/ConditionalFormats/ConditionalFormatCopyTests.cs
```diff
@@ -0,0 +1,29 @@
...
+
+            SetFormat1(ws.Range("A1:A1").AddConditionalFormat());
+            var wb2 = new XLWorkbook();
+            XLWorksheet ws2 = (XLWorksheet)wb2.Worksheets.Add("Sheet2");
```
igitur: No need to cast to `XLWorksheet`. Please keep it simple. Same as above.

Firedragonweb:
sure!

igitur:
Thanks for the contribution.

PR #721, opened 2018-03-07
FIX Shifting conditional format ranges works slow and incorrectly

### LINKED ISSUE #719: Inserting a column on a sheet with conditional formats works slow

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

e.g. 0.92.0 (current develop)

As @Laurakwakwak has reported, ClosedXML causes OutOfMemoryException on attempts to insert a column. The issue is caused by conditional formats since without them adding a column takes 0.1 sec, but when they are presented is takes a minute and 2GB of RAM to add a single column. If columns are added in a loop the OutOfMemory is easily reached.

The new column should be inserted after a specified one in a reasonable time, no matter if there are conditional formats or not.

No information

## Reproducibility
```c#
public void Main()
{
            using (var wb = new XLWorkbook(fileName))
            {
                ws.Column(4).InsertColumnsAfter(1); // <-- hangs for a minute or until OutOfMemory fired
            }
}
```
[sample_719.xlsx](https://github.com/ClosedXML/ClosedXML/files/1788898/sample_715.xlsx)

Pankraty:
This PR fixes #719 

When rows or columns were added conditional formats were shifted in a very specific way: first they were broken into multiple single-cell rules, and then they were applied to newly added cells. Moreover, even though on _row_ inserting the analysis was performed inside used cells only, on _column_ inserting each cell of the entire column (1048576 cells) was instantiated and processed separately. That's why columns were inserted terribly slow consuming gigabytes of RAM.

With this PR logic for conditional format ranges shifting becomes much simpler. Inserting rows and columns on sheets with conditional formats performs instantly and does not produce redundant CF rules.


ClosedXML/Excel/XLWorksheet.cs
```diff
@@ -1305,69 +1311,36 @@ private void ShiftPageBreaksRows(XLRange range, int rowsShifted)
...
+                                     cfAddress.LastAddress.RowNumber + rowsShifted,
+                                     cfAddress.LastAddress.ColumnNumber);
                 }
-            }
-            insertedRange.Dispose();
-            model.Dispose();
-            fr.Dispose();
-        }
-
-        internal void BreakConditionalFormatsIntoCells(List<IXLAddress> addresses)
```
igitur: Are you sure we won't need this for something else?

Pankraty: This method was not used anywhere else, so I assumed there would be no use in dead code, not even covered by tests. Besides, as a user I always hate when Excel splits a single CF into small pieces - say, from A:A it can create A1:A4, A5:A5, A6:A1048456.
I really do not see why it might be needed to split a CF into separate cells.

igitur: Ok, there are some other PR's related to conditional formatting, but it doesn't look like they're dependent on this method.

PR #732, opened 2018-03-14
Fix COUNTBLANK for invalid ranges

Pankraty:
In the test ```CountBlank``` there was a mistake causing a range ```E3:E45``` to be not parsed properly as no worksheet was specified. By accident, the test worked fine, however.

This PR fixes and extends the original test, makes ```COUNTBLANK``` function more strict (it accepts only a single argument which must be either a cell range reference), and, in addition, prevents non-parsed strings from being split into chars during enumeration

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Functions/Statistical.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/Statistical.cs
@@ -2,6 +2,7 @@ using System;
 using System.Net;
 using System.Collections;
 using System.Collections.Generic;
+using ClosedXML.Excel.CalcEngine.Exceptions;
 
 namespace ClosedXML.Excel.CalcEngine
 {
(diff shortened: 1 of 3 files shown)


ClosedXML/Excel/CalcEngine/Functions/Statistical.cs
```diff
@@ -117,23 +118,18 @@ static object CountA(List<Expression> p)
         static object CountBlank(List<Expression> p)
         {
             var cnt = 0.0;
-            foreach (Expression e in p)
+
+            if (p == null || p.Count != 1 ||
```
igitur: This test is already catered for. At the top of the class, this statement implies that only 1 parameter is allowed: `ce.RegisterFunction("COUNTBLANK", 1, CountBlank);`

ClosedXML/Excel/CalcEngine/Functions/Statistical.cs
```diff
@@ -2,6 +2,7 @@
 using System.Net;
```
igitur: Sort usings alphabetically. CodeMaid should do it.

Issue #734, opened 2018-03-15
SUMIFS fails at valid formula


Pankraty:
**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

Current develop

**What is the current behavior?**

The formula that works fine in Excel throws NullReferenceException during evaluation in ClosedXML

**What is the expected behavior or new feature?**

Return the correct result

**Did this work in previous versions of our tool?  Which versions?**

No information

## Reproducibility

**Code to reproduce problem:**
```c#
        // put this into StatisticalTests
        [TestCase("SUMIFS(H:H, G:G, 50, I:I, \">900\")", 19.99d)]
        public void SumIfsMultipleCriteriaTest(string formulaA1, double expectedResult)
        {
            using (var wb = SetupWorkbook())
            {
                var ws = wb.Worksheets.First();

                var actualResult = (double)ws.Evaluate(formulaA1);

                Assert.AreEqual(expectedResult, actualResult, tolerance);
            }
        }
```
- [x] I attached a sample spreadsheet.  (You can drag files on to this issue)
[sample_734.xlsx](https://github.com/ClosedXML/ClosedXML/files/1814209/sample_734.xlsx)

igitur:
`SetupWorkbook()` ?

igitur:
Oh, found it in `LookupTests`.

Pankraty:
I used ```StatisticalTests```, but they seem the same

igitur:
Found the bug. The current `SumIfs_ReturnsCorrectValues_ReferenceExampleFromMicrosoft` test never had an assert!

PR #736, opened 2018-03-15
WIP Compile and cache Regexes for SUMIFS and COUNTIFS

igitur:
Follow on from #734. Possible performance improvement.

### DIFF
--- a/ClosedXML/Excel/CalcEngine/CalcEngineHelpers.cs
+++ b/ClosedXML/Excel/CalcEngine/CalcEngineHelpers.cs
@@ -23,7 +23,7 @@ namespace ClosedXML.Excel.CalcEngine
                 return patternReplacements;
             });
 
-        internal static bool ValueSatisfiesCriteria(object value, object criteria, CalcEngine ce)
+        internal static bool ValueSatisfiesCriteria(object value, object criteria, CalcEngine ce, IDictionary<string, Regex> cachedRegexes)
         {
             // safety...
             if (value == null)
@@ -41,6 +41,11 @@ namespace ClosedXML.Excel.CalcEngine
 
             // convert criteria to string
             var cs = criteria as string;
+
+            Regex regex = null;
+            if (cachedRegexes.ContainsKey(cs))
+                regex = cachedRegexes[cs];
+
             if (cs != null)
             {
                 if (value is string && (value as string).Trim() == "")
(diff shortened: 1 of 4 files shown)


igitur:
Doesn't seem to make that much of a difference. I suspected it would - especially for whole-column lookups.

Pankraty (approved):
This way of caching is not thread-safe, but since the entire calculation engine is not thread-safe too I don't think it's a problem. As for the performance of column-wide computations, I am currently experimenting with the alternative approach that does not require enumeration of empty values. After all, ranges in formulas may cover the whole sheet, and there is not any possibility to enumerate 17 billion values in reasonable time. If I succeed with this optimization the effect of the current PR will become more significant.

igitur:
@Pankraty Do you think we should first quantify the effect on performance before I merge?

Pankraty:
If you have time to measure that would be great. This will give us more information for future optimizations.

Pankraty:
I finally managed to do a little performance testing. Unfortunately, the effect is negligible, if any. Here is the test file I used. 

[SumIfs.xlsx](https://github.com/ClosedXML/ClosedXML/files/3374131/SumIfs.xlsx)

Reading `I1` value would take ~13 seconds, reading `I2` - around 112 seconds, both with this PR and without it. As for `I3` and `I4`, I gave up before the evaluation finished.

Will check out if there are other ways to speed up the computation.

igitur:
Yeah, I was thinking to close this eventually without merging.

jahav:
I am closing this PR, because the regexps are not a good way to parse criteria for SUMIF/COUNTIF/database functions.

What is a better way forward: Use a custom parser. There already is a allocation free fast parser for wildcards (`Wildcard.Search`), as well as proper parser of numbers `ScalarValue.TextToNumber` (kind of awkward place, but it works), as well as `XLErrorParser` and few others.

A custom parser will parse the criteria into a comparison op and a value/bool/error/wildcard using other parsers that already exist. Benefits:
* More in-line with excel criteria (e.g. ">1 1/2" is a valid criteria in Excel that finds only values greater than 1.5)
* No startup cost for compiling
* IIRC wildcard was faster than regex
* Less dangerous from DoS point of view

Current approach has several drawbacks:
* Potential denial of service, there is a reason why regex match function has optional timeout argument. Regex is a basically a program and attacker could supply a very slow program.
* Requires allocations
* Is mismatched with Excel

PR #740, opened 2018-03-15
Optimize tally computations

Pankraty:
Calculation of formulas referring to wide ranges (such as entire rows or columns) required enumeration of all cells included. For a single column, this might take a few seconds (and huge memory allocation); for the entire worksheet, the computation would never end due to the out-of-memory.

This PR optimizes a computation in the following ways:

1. ```IXLRangeBase``` now implements the ```GetCellValues``` method which returns cell values without a need to initialize ```XLCell``` instances. This alone would increase the performance noticeably and would not require modification of math and statistical functions, but still would fail on extremely wide ranges.
2. When obtaining cell values for some range this range is trimmed by current worksheet's used range. Say, if we know that last used row on a worksheet is 5th, there is no need to enumerate each cell below A5 running through the range A:A since they all are obviously empty. This modification affects several functions that may actually depend on how much empty cells are there: ```COUNTBLANK```, ```COUNTIF```, ```SUMIF```, etc. All the rest (like ```MIN```, ```MAX```, ```AVG```) may safely ignore empty cells. 
Again, this optimization alone would bring performance gain but is not enough as it is not efficient in cases when there is a single initialized cell far-far away (i.e. ```ZZ100000```; I recall there was an issue reported with the example file matching this pattern).

I added a test performing wide-ranged computations (including a worksheet-wide), and even with this the total execution time of the test suite became twice lower (about 19 seconds against 40).

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -356,6 +356,11 @@ namespace ClosedXML.Excel
                 return cValue;
         }
 
+        /// <summary>
+        /// Flag showing that the cell is in formula evaluation state.
+        /// </summary>
+        internal bool IsEvaluating { get; private set; }
+
         public object Value
         {
             get
@@ -363,6 +368,9 @@ namespace ClosedXML.Excel
                 var fA1 = FormulaA1;
                 if (!String.IsNullOrWhiteSpace(fA1))
                 {
+                    if (IsEvaluating)
+                        throw new InvalidOperationException("Circular Reference");
+
                     if (fA1[0] == '{')
                         fA1 = fA1.Substring(1, fA1.Length - 2);
 
@@ -394,7 +402,17 @@ namespace ClosedXML.Excel
                             return referenceCell.Value;
                     }
 
-                    var retVal = Worksheet.Evaluate(fA1);
+                    object retVal;
+                    try
+                    {
+                        IsEvaluating = true;
+                        retVal = Worksheet.Evaluate(fA1);
+                    }
+                    finally
+                    {
+                        IsEvaluating = false;
+                    }
+
                     var retValEnumerable = retVal as IEnumerable;
 
                     if (retValEnumerable != null && !(retVal is String))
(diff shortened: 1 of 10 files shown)


igitur:
👍

ClosedXML/Excel/Ranges/IXLRangeBase.cs
```diff
@@ -93,6 +94,11 @@ public interface IXLRangeBase : IDisposable
 
         IXLCells CellsUsed(Boolean includeFormats, Func<IXLCell, Boolean> predicate);
 
+        /// <summary>
```
igitur: Why does this need to be publically exposed?

igitur:
Also check whether there are any major conflicts with #736

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -356,13 +356,21 @@ public string GetFormattedString()
                 return cValue;
         }
 
+        /// <summary>
+        /// Flag showing that the cell is in formula evaluation state.
+        /// </summary>
+        internal bool IsEvaluating { get; private set; }
```
igitur: Would you mind submitting the `XLCell` changes as a separate PR? I'll merge that first.

ClosedXML/Excel/Ranges/IXLRangeBase.cs
Pankraty: Tried to avoid explicit typecasting in ```CellRangeReference``` and thought it may be useful to have a possibility to quickly get all values in a range no matter how big it is. I do not insist on making this public, however.

Pankraty:
> Also check whether there are any major conflicts with #736

There should not be any. After you merge 736 I can rebase current PR and test once again.

ClosedXML/Excel/Ranges/IXLRangeBase.cs
igitur: Yes, I'd prefer to keep the interface unchanged unless we deliberately want to expose them as public.

PR #741, opened 2018-03-16
Treat negative row and column numbers as invalid

Pankraty:
Another funny mistake:
```
         if (row <= 0)
             RangeAddress.IsInvalid = false;
```


igitur:
I've long thought to change this to an `IsValid` property. It makes more sense to me. But to change everything is going to be dangerous. One would have to be VERY careful.

PR #742, opened 2018-03-16
Refactor IsInvalid to IsValid

igitur:
Refactor IsInvalid to IsValid
Remove public setter from IXLRangeAddress.IsValid

### DIFF
--- a/ClosedXML/Excel/Ranges/IXLRangeAddress.cs
+++ b/ClosedXML/Excel/Ranges/IXLRangeAddress.cs
@@ -13,12 +13,12 @@ namespace ClosedXML.Excel
         IXLAddress FirstAddress { get; set; }
 
         /// <summary>
-        /// Gets or sets a value indicating whether this range is invalid.
+        /// Gets or sets a value indicating whether this range is valid.
         /// </summary>
         /// <value>
-        /// 	<c>true</c> if this instance is invalid; otherwise, <c>false</c>.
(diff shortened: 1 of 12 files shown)


ClosedXML/Excel/Ranges/IXLRangeAddress.cs
```diff
@@ -13,12 +13,12 @@ public interface IXLRangeAddress
         IXLAddress FirstAddress { get; set; }
 
         /// <summary>
-        /// Gets or sets a value indicating whether this range is invalid.
+        /// Gets or sets a value indicating whether this range is nvalid.
```
Pankraty: ```nvalid``` - a typo

ClosedXML/Excel/Ranges/XLRangeAddress.cs
```diff
@@ -123,7 +123,7 @@ IXLAddress IXLRangeAddress.LastAddress
             set { LastAddress = value as XLAddress; }
         }
 
-        public bool IsInvalid { get; set; }
+        public bool IsValid { get; set; } = true;
```
Pankraty: Make setter internal?

igitur: `XLRangeAddress` is already an internal class, so the setter is by default internal.

Pankraty: My bad

PR #745, opened 2018-03-19
Make cell addresses immutable

Pankraty:
This is another small step in a refactoring of addresses and ranges handling.

With this PR I propose the following changes:
1. There was an optimization letting column letter of cell address to be initialized only once per instance of ```XLAddress```. After the initialization, it was being stored in the private field. The thing is, that the gain of this optimization would be less than 0.1 microseconds, so it would become noticeable only if the column letter of the **same** ```XLAddress``` instance is determined millions of times (which does not look as an ordinary scenario). And the drawback of such optimization was that **each** instance of ```XLAddress``` held a reference to a string value. And since the number of instances can easily reach a few million, the drawback overweights. So I suggest not to store a column letter in ```XLAddress```, but read it on demand, by accessing an element in a pre-initialized array by its index.
2. If an address is mutable there is always a risk that multiple entities referring to the same instance of ```XLAddress``` will affect each other. To avoid this, the new instance should have been created from the existing one (```FirstAddress = XLAddress.Create(firstAddress);```). When the ```XLAddress``` is immutable, this is not needed anymore; entities are protected from affecting each other "by design". 

Here are results of profiling using the code of Pitterling.
**develop branch**
Objects: 31 671 019, Heap size 2 042 171.63 KB
image

**This PR**
Objects: 25 186 160, Heap size 1 840 127.19 KB
image

The weight of ```XLRangeAddress``` instances grew, but this class is next on the list to be refactored.

Execution time has not been affected.

### DIFF
--- a/ClosedXML/Excel/Coordinates/XLAddress.cs
+++ b/ClosedXML/Excel/Coordinates/XLAddress.cs
@@ -4,7 +4,7 @@ using System.Diagnostics;
 
 namespace ClosedXML.Excel
 {
-    internal class XLAddress : IXLAddress
+    internal struct XLAddress : IXLAddress, IEquatable<XLAddress>
     {
         #region Static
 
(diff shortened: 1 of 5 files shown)


igitur:
Very well explained, thank you.

igitur:
I'm holding this for v0.93. When we merge, after the release of v0.92, I will ask you to rebase.

Pankraty:
Deal

igitur:
Please rebase on `develop`.

Pankraty:
Rebased

PR #746, opened 2018-03-19
IsEvaluating flag added to XLCell

Pankraty:
A small part of #740

### DIFF
--- a/ClosedXML/Excel/CalcEngine/XLCalcEngine.cs
+++ b/ClosedXML/Excel/CalcEngine/XLCalcEngine.cs
@@ -90,9 +90,9 @@ namespace ClosedXML.Excel.CalcEngine
         // ** implementation
         private object GetValue(IXLCell cell)
         {
-            if (_evaluating)
+            if (_evaluating || (cell as XLCell).IsEvaluating)
             {
-                throw new InvalidOperationException("Circular Reference");
+                throw new InvalidOperationException($"Circular Reference occured during evaluation. Cell: {cell.Address.ToString(XLReferenceStyle.Default, true)}");
             }
             try
             {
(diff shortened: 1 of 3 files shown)


igitur:
Sorry, one more thing. Please add unit test to catch circular reference exception.

Pankraty:
You got me! It fails )

Pankraty:
This, BTW, interferes with #679, where circular references were tested and fixed too.

ClosedXML/Excel/CalcEngine/XLCalcEngine.cs
```diff
@@ -90,7 +90,7 @@ public IEnumerator GetEnumerator()
         // ** implementation
         private object GetValue(IXLCell cell)
         {
-            if (_evaluating)
+            if (_evaluating || (cell is XLCell && (cell as XLCell).IsEvaluating))
```
igitur: I think you can safely assume that `cell is XLCell` will always be `true`. If not, basically everything in ClosedXML will break.

ClosedXML/Excel/CalcEngine/XLCalcEngine.cs
```diff
@@ -90,7 +90,7 @@ public IEnumerator GetEnumerator()
...
             {
                 throw new InvalidOperationException("Circular Reference");
```
igitur: Include the cell address (including sheet name) in the exception message. This will allow the user to find the culprit.

igitur: One of the `cell.Address.ToString(XLReferenceStyle.Default, false)` is what you should include.

Pankraty: Sorry, missed this comment

PR #750, opened 2018-03-20
Implement IsEvaluating flag for simple/direct formula evaluation too.

igitur:
Follow-on from #746

### DIFF
--- a/ClosedXML_Tests/Excel/Cells/XLCellTests.cs
+++ b/ClosedXML_Tests/Excel/Cells/XLCellTests.cs
@@ -637,5 +637,15 @@ namespace ClosedXML_Tests
                 });
             }
         }
+
+        public void FormulaWithCircularReferenceFails2()
+        {
+            var cell = new XLWorkbook().Worksheets.Add("Sheet1").FirstCell();
+            cell.FormulaA1 = "A1";
+            Assert.Throws<InvalidOperationException>(() =>
+            {
+                var _ = cell.Value;
+            });
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


ClosedXML_Tests/Excel/Cells/XLCellTests.cs
```diff
@@ -637,5 +637,15 @@ public void FormulaWithCircularReferenceFails()
                 });
             }
         }
+
+        public void FormulaWithCircularReferenceFails2()
+        {
+            Assert.Throws<InvalidOperationException>(() =>
+            {
+                var cell = new XLWorkbook().Worksheets.Add("Sheet1").FirstCell();
```
Pankraty: I'd prefer to have arrangement separated from assertion. For instance, if at some point in future ```cell.FormulaA1 = "A1"``` will start throwing the exception it may remain unnoticed.

igitur: I don't understand.

Pankraty:
```
            var cell = new XLWorkbook().Worksheets.Add("Sheet1").FirstCell();
            cell.FormulaA1 = "A1";
            Assert.Throws<InvalidOperationException>(() =>
            {
                var _ = cell.Value;
            });```

Like this. Here we may be absolutely sure at which step the exception is expected to be thrown.

igitur: Oh, I see. Yes, that's better.

PR #758, opened 2018-03-27
Add RemoveAll method to IXLRanges

Pankraty:
```IXLRanges``` has a method ```Clear()``` which I personally tried to use for removing all ranges from the collection, while it serves to a completely different purpose (to _clear_ data, formats, etc. for cells included in the ranges). To prevent such a confusion and to simplify collection clearing I propose adding ```RemoveAll()``` method to the interface. By default, it will dispose removed ranges, but this is optional.

### DIFF
--- a/ClosedXML/Excel/Ranges/IXLRanges.cs
+++ b/ClosedXML/Excel/Ranges/IXLRanges.cs
@@ -20,10 +20,13 @@ namespace ClosedXML.Excel
         void Remove(IXLRange range);
 
         /// <summary>
-        /// Removes all ranges from the collection optionally disposing them.
+        /// Removes ranges matching the criteria from the collection, optionally releasing their event handlers.
         /// </summary>
-        /// <param name="dispose"></param>
-        void RemoveAll(bool dispose = true);
+        /// <param name="match">Criteria to filter ranges. Only those ranges that satisfy the criteria will be removed.
+        /// Null means the entire collection should be cleared.</param>
+        /// <param name="releaseEventHandlers">Specify whether or not should removed ranges be unsubscribed from 
+        /// row/column shifting events. Until ranges are unsubscribed they cannot be collected by GC.</param>
+        void RemoveAll(Predicate<IXLRange> match = null, bool releaseEventHandlers = true);
 
         Int32 Count { get; }
 
--- a/ClosedXML/Excel/Ranges/IXLRanges.cs
+++ b/ClosedXML/Excel/Ranges/IXLRanges.cs
@@ -19,6 +19,12 @@ namespace ClosedXML.Excel
         /// <param name="range">The range to remove from this group.</param>
         void Remove(IXLRange range);
 
+        /// <summary>
+        /// Removes all ranges from the collection optionally disposing them.
+        /// </summary>
+        /// <param name="dispose"></param>
+        void RemoveAll(bool dispose = true);
+
         Int32 Count { get; }
 
         Boolean Contains(IXLRange range);
--- a/ClosedXML_Tests/Excel/Ranges/XLRangeBaseTests.cs
+++ b/ClosedXML_Tests/Excel/Ranges/XLRangeBaseTests.cs
@@ -415,13 +415,30 @@ namespace ClosedXML_Tests
             ranges.Add(ws.Range("B1:B2"));
             var rangesCopy = ranges.ToList();
 
-            ranges.RemoveAll(false);
+            ranges.RemoveAll(null, false);
             ws.FirstColumn().InsertColumnsBefore(1);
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by igitur** [comment]
I'm not convinced about this one. This seems like an internal matter. For example, what would happen if the user has references to ranges in his own local variables. If you dispose them, wouldn't it cause big problems?

**Comment by Pankraty** [comment]
What is the valid way to clear the ```IXLRange``` collection then?

**Comment by igitur** [comment]
By "clear" we mean emptying the `_ranges` list, right? There is not current way. But why would the user want to clear them? Can you give me a real-world use-case?

**Comment by Pankraty** [comment]
If conditional formats will store a collection of ranges instead of a single range (as was discussed in #757) the user may want to modify this collection. In this case, it would be natural to clear the ```Ranges``` and then add new ranges to it. Without the ```RemoveAll``` the user most probably will choose the ```Clear``` method, and after he'll found that this is wrong he'll need to do something like ```while (cf.Ranges.Any()) cf.Ranges.Remove(cf.Ranges.First());``` which is ridiculous.

Another use case - the user may want to clear a collection of ```Worksheet.SelectedRanges```.

**Comment by igitur** [comment]
Ok, I understand more. I first thought it was an internal optimisation trick, but I now see the real use-cases that would apply. However, I'm a bit concerned that there might be side effects for `XLRanges` instances where they should not be clearable. 

e.g. this doesn't make sense, but luckily it doesn't crash (from what I can tell):
```c#
ws.Cell("A1").AsRange().Ranges("A1").RemoveAll();
```
So I'm accepting this PR, but with modified API. If we encounter problems in the future, we could subclass `XLRanges` to something like `XLClearableRanges` and have the `RemoveAll` exposed on the subclass.

ClosedXML/Excel/Ranges/IXLRanges.cs
```diff
@@ -19,6 +19,12 @@ public interface IXLRanges : IEnumerable<IXLRange>, IDisposable
         /// <param name="range">The range to remove from this group.</param>
         void Remove(IXLRange range);
 
+        /// <summary>
+        /// Removes all ranges from the collection optionally disposing them.
+        /// </summary>
+        /// <param name="dispose"></param>
+        void RemoveAll(bool dispose = true);
```
igitur: What do you think of this rather:
```c#
void RemoveAll(Predicate<IXLRange> match = null, bool releaseEventHandlers = true);
```

to better match the standard Linq `RemoveAll()` method?

And I think `releaseEventHandlers` is a better description for the user than `dispose`. The user won't know what `dispose` means.

igitur: Of course, if `match` is `null`, then make `match = _ => true`.

igitur: Or do you think it would cause users to think they can do something like:

```c#
var ranges = new XLRanges();
ranges.Add(ws.Range("A1:Z10"));
ranges.RemoveAll(r => r.RangeAddress.ToString().Equals("B2:C4"));
```

which is obviously not possible without breaking up the ranges first.

Pankraty: > What do you think of this rather: ...

I agree.

> And I think releaseEventHandlers is a better description for the user than dispose.

True. Moreover, while I was working on another PR (#766) I realized that this ```dispose``` argument may become meaningless if we got rid of events subscription.

> Or do you think it would cause users to think they can do something like:

I really doubt someone can decide to do this. We have a collection and a method that accepts a predicate - for me, this would be quite obvious that the predicate will be applied to items in the collection only, not to some derivatives.

Pankraty:
> e.g. this doesn't make sense, but luckily it doesn't crash (from what I can tell):

```ws.Cell("A1").AsRange().Ranges("A1").RemoveAll();```

The user is able to call ```ranges.Remove(ranges.First())``` already, so if he really wants to shoot his foot we can't stop him anyway.

File: ClosedXML_Tests/Excel/Ranges/XLRangeBaseTests.cs [hunk 2]
```diff
@@ -415,13 +415,30 @@ public void RangesRemoveAllWithoutDispose()
...
+        public void RangesRemoveAllByCriteria()
+        {
+            var ws = new XLWorkbook().Worksheets.Add("Sheet1");
+            var ranges = new XLRanges();
+            ranges.Add(ws.Range("A1:A2"));
+            ranges.Add(ws.Range("B1:B3"));
+            ranges.Add(ws.Range("C1:C4"));
+            var otherRange = ws.Range("A3:D3");
+
+            ranges.RemoveAll(r => r.Intersects(otherRange));
```
igitur: I like this test!

PR #761, opened 2018-03-28
Сonsolidation of conditional formats and data validation rules

### LINKED ISSUE #757: Conditional formatting consolidation break priority ordering

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.92.0

The sequence of "conditional formatting rules" has been changed after export Excel by using ClosedXML in my program

The sequence should not be changed in Excel template after export excel.

No

## Reproducibility

```c#
public void Main()
{
      using (XLWorkbook workbook = new XLWorkbook(@"FAI.xlsm"))
   {
       workbook.Save();
   }
}
```
[test4.xlsx](https://github.com/ClosedXML/ClosedXML/files/1839700/test4.xlsx)

Pankraty:
This PR fixes #757.
There are many changes in this PR but that's because it is based on #709 and #758. After these two are merged I will rebase the current one. Meanwhile, it can be reviewed.

This PR consists of the following parts:
1. Implement a ```Consolidate``` method for ```XLRanges```. It rebuilds the ranges completely, combining everything that can be combined.
2. Define ```Ranges``` property for ```XLConditionalFormat``` that allows defining multiple ranges for a single conditional format without the need to split them. I am still not sure about what to do with the original ```Range``` property; for now, I marked it as an obsolete.
3. Re-implement a ```Consolidate``` method for ```XLConditionalFormats``` with respect to rules priorities. Similar rules are joined as long as this does not cause changing the order of rules affecting the same cells.
4. Implement a ```Consolidate``` method for ```XLDataValidations``` too. It is being called on a workbook saving and makes the rules definition much more compact in some cases.


igitur:
If you can rebase on `develop` that would be great, else merge and fix is fine.

Pankraty:
Rebased

ClosedXML/Excel/ConditionalFormats/IXLConditionalFormat.cs
```diff
@@ -140,8 +140,9 @@ public interface IXLConditionalFormat
         Boolean ShowIconOnly { get; }
         Boolean ShowBarOnly { get; }
         Boolean StopIfTrue { get; }
+        [Obsolete("Use Ranges instead")]
```
igitur: I actually think we should keep `Range` forever, not make it obsolete. Most/many conditional formats will refer to only 1 range and this property will ease it a lot.

ClosedXML/Excel/ConditionalFormats/Save/XLCFDatesOccuringConverter.cs
```diff
@@ -29,7 +30,7 @@ public ConditionalFormattingRule Convert(IXLConditionalFormat cf, int priority,
 
             conditionalFormattingRule.TimePeriod = cf.TimePeriod.ToOpenXml();
 
-            var address = cf.Range.RangeAddress.FirstAddress.ToStringRelative(false);
+            var address = cf.Ranges.First().RangeAddress.FirstAddress.ToStringRelative(false);
```
igitur: Just checking... shouldn't we use `cf.Range` instead of `cf.Ranges.First()` ?
So the question is, if there is more than one range, should this fail or continue?

Pankraty: I just realized that here most likely we should use the top-left cell for all ranges, like [there](https://github.com/ClosedXML/ClosedXML/issues/757#issuecomment-380886527). I will check and if so will add ```BaseCellAddress``` property to XLConditionalFormat.

ClosedXML/Excel/ConditionalFormats/IXLConditionalFormat.cs
Pankraty: OK

ClosedXML/Excel/ConditionalFormats/Save/XLCFDatesOccuringConverter.cs
Pankraty: > we should use the top-left cell for all ranges
No, it's OK to get the first cell of the first range.

> shouldn't we use ```cf.Range``` instead of ```cf.Ranges.First()```?
I made ```Range``` returning ```.FirstOrDefault``` instead of ```.Single()``` and reverted these changes. Rules with no ranges in collections are removed during consolidation, so we won't get NullRef here.

PR #765, opened 2018-03-31
Cache StringFormat.GenericTypographic

DyachenkoA:
StringFormat.GenericTypographic return new object for each call.

StringFormat object has finalizer this create big load to finalization queue when call method AdjustToContents.

### DIFF
--- a/ClosedXML/Utils/GraphicsUtils.cs
+++ b/ClosedXML/Utils/GraphicsUtils.cs
@@ -20,9 +20,10 @@ namespace ClosedXML.Utils
             }
         }
 
+        private static StringFormat defaultStringFormat = StringFormat.GenericTypographic;
         public static SizeF MeasureString(string s, Font font)
         {
-            SizeF result = Graphics.MeasureString(s, font, Int32.MaxValue, StringFormat.GenericTypographic);
+            SizeF result = Graphics.MeasureString(s, font, Int32.MaxValue, defaultStringFormat);
             return result;
         }
     }


igitur:
Thanks. Do you have any metrics on the performance impact?

DyachenkoA:
No.

When i'm export data to excel, my app use 1.5 GB RAM for 3.5 MB result xlsx file.

I'm try optimize this. Profiler (dotTrace) show ~900 000 objects in finalization queue.

Pankraty:
Well, the getter of the property creates a new instance indeed

```
public static StringFormat GenericTypographic
{
    get
    {
        IntPtr ptr;
        int status = SafeNativeMethods.Gdip.GdipStringFormatGetGenericTypographic(out ptr);
        if (status != 0)
        {
            throw SafeNativeMethods.Gdip.StatusException(status);
        }
        return new StringFormat(ptr);
    }
}
```

So it seems reasonable to preserve the instance between measurements. But I am not sure about making it a private static field. I have no idea in what circumstances this instance should be invalidated but imagine it becomes invalid on a screen resolution change. In that case, a user won't be able to use part of ClosedXML functionality without a re-loading entire AppDomain.

DyachenkoA:
StringFormat returned from GenericTypographic is depends on the resolution of the screen?

Pankraty:
Perhaps it does not. That was just a wild guess. What I tried to say since we don't know what can make the cached instance invalid it is not quite safe to store it indefinitely. 
I would consider initializing it once in ```AdjustToContents``` method and pass to ```GetWidth``` and ```GetHeight``` methods, just like it is done for ```fontCache```, and then pass it to ```GraphicsUtils.MeasureString```.

DyachenkoA:
According to the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.stringformat.generictypographic?view=netframework-4.5), method GenericTypographic does not depend on anything and always returns the same value.

Pankraty:
Ah, ok, then just ignore my comments )

PR #766, opened 2018-03-31
Ranges handling redesign (proof-of-concept)

Pankraty:
Depends on #709 

Work has not finished yet, but I would like to discuss the approach before I finalize it.

One of the hugest problems of ClosedXML is a large memory consumption, especially when many ranges are involved. The reason is that ranges, when created, are subscribed to worksheet modifications, so after rows or columns are inserted or deleted the ranges are shifted accordingly. To achieve this, a pair of delegates is created on each range creation and removed on range disposing. This causes a few serious problems that already hit multiple times:

1. If the range is not disposed related delegates remain in a worksheet collection until the workbook is closed, causing an excessive memory usage and slowing down the editing operations. To avoid this, developers must carefully dispose all ranges after they are not needed anymore.
2. But often a developer cannot be sure if the particular range is referenced from somewhere else (thus, has not to be disposed) or has already been disposed (therefore will not be shifted properly). This forces a developer to use such weird constructions as 

```
using (var range = anotherRange.AsRange()) 
  list.Add(range);
return list;
```

Why does this seem weird? Because a) normally we don't want to return disposed objects out of the method, but for ClosedXML it is natural. And b) having an instance of ```XLRange``` to generate another instance of ```XLRange``` from it looks absolutely redundant. *(I totally understand the reasons but look for more "sane" ways to accomplish the task)*

AFAIK, @igitur has experimented with the weak event pattern that could solve part of the problems (at least, non-referenced ranges could be disposed automatically), but unfortunately, the implementations of this patter showed the poor performance. 

What am I proposing?

The main idea is to store used ranges in a weak repository like it was proposed earlier for styles. But this requires many changes.

1. ```XLRangeAddress``` must become immutable. ```XLAddress``` made immutable recently, and this is just another step on this road. When range addresses are immutable we are fully protected from the situation when modification of the ```LastAddress``` of the range ```A1:E5``` affects ranges ```E5:E6```, ```E5:F8```, etc., just because they accidently use the same instance of ```XLAddress```
2. Implement an ```XLRangeRepository``` that stores weak references to all created ranges. While the range is referenced from some place (either captured in a variable, or stored in a conditional formatting rule, or in a data validation rule, whatsoever) it remains in a repository too, but after all other references are released it can be safely collected by a GC.
3. Ranges having the same addresses are re-used. If for some reason user creates a million instances of the range ```A1:B5``` there is no need to shift them all independly. Instead, we return the very same range a million times and shift it only once, when needed.
4. To make a reusing possible all the calls of the ranges constructors must be replaced with corresponding methods (like, ```Worksheet.Range(...)``` instead of ```new XLRange(...)```. Inside of these methods, we check if the range is already presented in a repository, and if it is we return a preserved copy. Otherwise call a constructor (a factory method, actually, but who cares).
5. When the shifting has to be performed we simply get the entire collection of ranges from the repository and modify them one by one. Unlike the current approach we do not shift ranges that were forgotten to be disposed, and do not shift ranges with same addresses multiple times. So the overall performance is improved.
6. In contrast to styles that are stored in weak repositories too, ranges are not immutable be themselves. Therefore, when the range address changes it must be removed from the repository and put back with a new key. I called this method ```RellocateRange```, it is fired when ```RangeAddress``` property is modified. In addition, we never deleted styles, assuming that they will be drained out of the repository after a GC collected them. For ranges I had to support a deleting operation too, since deleted ranges must be released for other ranges could refer to the same addresses.
7. Another difference with style repositories is that ranges are not shared globally, they always belong to a single worksheet. So each worksheet has its own ```_rangesRepository``` which is being cleared on a worksheet disposal.

Initially, I designed the repository in such way that all ```XLRangeBase``` descendants were stored there: XLRange, XLRow, XLColumn, XLTable, and so on. Later I found that there is no use in storing columns and rows (and worksheets, BTW) in the repository, so I most likely will revise this part. Tables, on the other hand, must be there, but I have not done it yet (that is why a few tests fail). 

After the basic functionality implemented I am going to remove some parts of code that are not needed any more, like subscriptions to shifting, ranges copying and disposing, etc. ```XLRange``` will remain ```IDisposable``` (for not breaking the existing code) but I guess the body of ```Dispose``` method will become empty.

And finally, what is the benefit? Is this all worth doing? I am pretty sure it is. 
Here is the profiling session from the develop branch (not the freshest one, but still actual)
image

And that's - from the current PR (including styles optimization from the other PR).
image

The memory usage dropped very significantly, even though there are optimizations still to be done.

Let's discuss?


Pankraty:
Now that I made all tests green I measured the performance too. 
In a Release configuration, this [test scenario](https://github.com/Pitterling/ClosedXML-Memory-Issues/blob/master/ClosedXML-Memory-Issues/Program.cs) takes 37 seconds to execute on develop branch and 29.5 seconds with this PR.

igitur:
Please fix conflicts.

igitur:
Really good work. So you removed events completely. I didn't realise that previously, which is why I continued working on weak events. I thought both changes could be combined.

I think this is the way to go. 

Maybe ref structs can benefit this even more.

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -1448,11 +1442,9 @@ public IXLDataValidation SetDataValidation()
             var validation = GetDataValidation();
             if (validation == null)
             {
-                using (var range = this.AsRange())
-                {
-                    validation = new XLDataValidation(range);
-                    Worksheet.DataValidations.Add(validation);
-                }
+                var range = this.AsRange();
```
igitur: I introduced many of the `using` statements a few months back. If we're going to revert it, I'd prefer not to create `var range` variables, because they also weren't created prior to my changes. So I'd prefer code like this:

```c#
validation = new XLDataValidation(AsRange());
```

You should be able to see in the git log how they used to be. I'll help out with this change, because it's very trivial and a lot of work.

Pankraty: I tried to do as you described but probably missed some of the occurrences.

Pankraty:
Whoops, shifting of merged ranges was not covered by tests and I broke this part. Now it's fixed.

Interesting that when rows or columns are inserted after a formatted range in Excel their fill style, font style, and conditional formats are expanded on shifting, while borders and merged cells are not.

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -1864,8 +1854,8 @@ private Boolean SetRange(Object rangeObject)
                 {
                     var maxRows = asRange.RowCount();
                     var maxColumns = asRange.ColumnCount();
-                    using (var rng = Worksheet.Range(_rowNumber, _columnNumber, maxRows, maxColumns))
-                        rng.Clear();
+                    var rng = Worksheet.Range(_rowNumber, _columnNumber, maxRows, maxColumns);
```
igitur: `Worksheet.Range(_rowNumber, _columnNumber, maxRows, maxColumns).Clear();`

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -1981,8 +1971,8 @@ private bool SetEnumerable(object collectionObject)
         private void ClearMerged()
         {
             List<IXLRange> mergeToDelete;
-            using (var asRange = AsRange())
-                mergeToDelete = Worksheet.Internals.MergedRanges.Where(merge => merge.Intersects(asRange)).ToList();
+            var asRange = AsRange();
+            mergeToDelete = Worksheet.Internals.MergedRanges.Where(merge => merge.Intersects(asRange)).ToList();
```
igitur: Remove `asRange` variable and use only `mergeToDelete = Worksheet.Internals.MergedRanges.Where(merge => merge.Intersects(AsRange())).ToList();`

ClosedXML/Excel/XLWorksheet.cs
```diff
@@ -1350,35 +1336,57 @@ private void MoveNamedRangesColumns(XLRange range, int columnsShifted, IXLNamedR
...
             {
-                foreach (var item in RangeShiftedRows)
+                SuspendEvents();
+                var rangesToShift = _rangeRepository.ToList();
+
+                WorksheetRangeShiftedRows(range, rowsShifted);
+                foreach (var storedRange in rangesToShift)
                 {
-                    item.Action(range, rowsShifted);
+                    var addr = storedRange.RangeAddress;
```
igitur: Remove unused `addr` variable.

ClosedXML/Excel/Ranges/XLRange.cs
```diff
@@ -760,9 +736,11 @@ private void TransposeMerged(Int32 squareSide)
                 RangeAddress.FirstAddress.RowNumber + squareSide - 1,
                 RangeAddress.FirstAddress.ColumnNumber + squareSide - 1);
 
-            foreach (IXLRange merge in Worksheet.Internals.MergedRanges.Where(Contains))
+            foreach (XLRange merge in Worksheet.Internals.MergedRanges.Where(Contains))
```
igitur: Please use this style: `foreach (var merge in Worksheet.Internals.MergedRanges.Where(Contains).Cast<XLRange>())`

ClosedXML/Excel/Ranges/XLRangeAddress.cs
```diff
@@ -113,34 +120,76 @@ IXLAddress IXLRangeAddress.FirstAddress
...
         #endregion Public properties
 
         #region Public methods
 
+        /// <summary>
+        /// Lead a range address to a normal form - when <see cref="FirstAddress"/> points to the top-left address and
+        /// <see cref="LastAddress"/> points to the bottom-right address.
+        /// </summary>
+        /// <returns></returns>
+        public XLRangeAddress Normalize()
```
igitur: Can this method not be simplified with `Min` and `Max` functions?

ClosedXML/Excel/Ranges/XLRangeAddress.cs
```diff
@@ -1,12 +1,31 @@
...
+        public static XLRangeAddress EntireRow(XLWorksheet worksheet, int row)
+        {
+            return new XLRangeAddress(
+                new XLAddress(worksheet, row, 1, false, false),
+                new XLAddress(worksheet, row, XLHelper.MaxColumnNumber, false, false));
+        }
+
+        #endregion Static members
+
         #region Private fields
```
igitur: Remove the entire `Private fields` region, including `_firstAddress` and `_lastAddress`. Implement autoproperty in `FirstAddress` and `LastAddress`

ClosedXML/Excel/Ranges/XLRangeAddress.cs
```diff
@@ -52,22 +67,22 @@ public XLRangeAddress(XLWorksheet worksheet, String rangeAddress)
 
             if (XLHelper.IsValidA1Address(firstPart))
             {
-                FirstAddress = XLAddress.Create(worksheet, firstPart);
-                LastAddress = XLAddress.Create(worksheet, secondPart);
+                _firstAddress = XLAddress.Create(worksheet, firstPart);
```
igitur: Readonly autoproperties are settable in the constructor.

ClosedXML/Excel/Ranges/XLRangeAddress.cs
```diff
@@ -203,22 +252,46 @@ public string ToString(XLReferenceStyle referenceStyle, bool includeSheet)
...
+                   EqualityComparer<XLWorksheet>.Default.Equals(Worksheet, address.Worksheet);
         }
 
         public override int GetHashCode()
         {
-            return
-                Worksheet.GetHashCode()
-                ^ FirstAddress.GetHashCode()
-                ^ LastAddress.GetHashCode();
+            var hashCode = -778064135;
```
igitur: `GetHashCode` should not reference mutable fields. Change `_firstAddress` and `_lastAddress` to `FirstAddress` and `LastAddress`

ClosedXML/Excel/Tables/XLTable.cs
```diff
@@ -19,38 +18,31 @@ internal class XLTable : XLRange, IXLTable
...
-                id++;
-            }
+            InitializeValues(false);
         }
 
         public XLTable(XLRange range, String name, Boolean addToTables, Boolean setAutofilter = true)
             : base(new XLRangeParameters(range.RangeAddress, range.Style))
         {
-            CheckRangeNotInTable(range);
+
```
igitur: Why is `CheckRangeNotInTable` removed?

ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -2090,23 +2090,21 @@ private static void LoadDataValidations(DataValidations dataValidations, XLWorks
...
-                        if (dvs.Prompt != null) dvt.InputMessage = dvs.Prompt;
-                        if (dvs.ErrorTitle != null) dvt.ErrorTitle = dvs.ErrorTitle;
-                        if (dvs.Error != null) dvt.ErrorMessage = dvs.Error;
-                        if (dvs.ErrorStyle != null) dvt.ErrorStyle = dvs.ErrorStyle.Value.ToClosedXml();
-                        if (dvs.Type != null) dvt.AllowedValues = dvs.Type.Value.ToClosedXml();
-                        if (dvs.Operator != null) dvt.Operator = dvs.Operator.Value.ToClosedXml();
-                        if (dvs.Formula1 != null) dvt.MinValue = dvs.Formula1.Text;
-                        if (dvs.Formula2 != null) dvt.MaxValue = dvs.Formula2.Text;
-                    }
+                    var range = ws.Range(rangeAddress);
```
igitur: `var dvt = ws.Range(rangeAddress).SetDataValidation();`

ClosedXML/Excel/XLWorksheet.cs
```diff
@@ -1,27 +1,22 @@
 using ClosedXML.Excel.CalcEngine;
 using ClosedXML.Excel.Drawings;
-using ClosedXML.Excel.Misc;
 using System;
 using System.Collections.Generic;
 using System.Drawing;
 using System.IO;
 using System.Linq;
+using ClosedXML.Excel.Caching;
```
igitur: Sort using alphabetically using CodeMaid clean.

ClosedXML/Excel/XLWorksheet.cs
```diff
@@ -371,9 +378,9 @@ public IXLRows Rows(String rows)
                     firstRow = tPair;
                     lastRow = tPair;
                 }
-                using (var xlRows = Rows(Int32.Parse(firstRow), Int32.Parse(lastRow)))
-                    foreach (IXLRow row in xlRows)
-                        retVal.Add((XLRow)row);
+
+                var xlRows = Rows(Int32.Parse(firstRow), Int32.Parse(lastRow));
```
igitur: Combine into 1 statement?

ClosedXML/Excel/XLWorksheet.cs
```diff
@@ -900,22 +916,19 @@ public new IXLRanges Ranges(String ranges)
...
                 {
-                    using (var range = Range(new XLRangeAddress(Worksheet, rangeAddressStr)))
-                        retVal.Add(range);
+                    retVal.Add(Range(new XLRangeAddress(Worksheet, rangeAddressStr)));
                 }
                 else if (NamedRanges.Any(n => String.Compare(n.Name, rangeAddressStr, true) == 0))
                 {
-                    using (var xlRanges = NamedRange(rangeAddressStr).Ranges)
-                        xlRanges.ForEach(retVal.Add);
+                    var xlRanges = NamedRange(rangeAddressStr).Ranges;
```
igitur: Combine into 1 statement.

ClosedXML/Excel/XLWorksheet.cs
```diff
@@ -900,22 +916,19 @@ public new IXLRanges Ranges(String ranges)
...
+                    xlRanges.ForEach(retVal.Add);
                 }
                 else
                 {
-                    using (var xlRanges = Workbook.NamedRanges.First(n =>
+                    var xlRanges = Workbook.NamedRanges.First(n =>
```
igitur: Combine into 1 statement.

ClosedXML/Excel/XLWorksheet.cs
```diff
@@ -1350,35 +1336,57 @@ private void MoveNamedRangesColumns(XLRange range, int columnsShifted, IXLNamedR
 
         public void NotifyRangeShiftedRows(XLRange range, Int32 rowsShifted)
         {
-            if (RangeShiftedRows != null)
+            try
```
igitur: Why inside a `try`?

ClosedXML_Tests/Excel/Caching/SampleRepositoryTests.cs
```diff
@@ -5,6 +5,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 using System.Collections.Generic;
+using ClosedXML.Excel;
```
igitur: Sort usings alphabetically using CodeMaid clean.

ClosedXML_Tests/Excel/Ranges/MergedRangesTests.cs
```diff
@@ -1,3 +1,4 @@
+using System.Linq;
```
igitur: Sort usings alphabetically using CodeMaid clean.

ClosedXML/Excel/Ranges/XLRangeAddress.cs
Pankraty: I doubt so. Depending on which row (column) is used for `firstRow` (`firstColumn`) we use different values for `FixedRow` (`FixedColumn`). So `Min` / `Max` would not help very much.

ClosedXML/Excel/Ranges/XLRangeAddress.cs
Pankraty: Here I had a short discussion on this matter. https://www.jetbrains.com/help/resharper/2017.3/NonReadonlyMemberInGetHashCode.html#comment-3862369295 Am I wrong?

Pankraty: I made `Worksheet` read-only anyway.

ClosedXML/Excel/Tables/XLTable.cs
Pankraty: Moved to `XLWorksheet.Table()`

ClosedXML/Excel/XLWorksheet.cs
Pankraty: For a single failed `shift` would not leave events for entire sheet turned off. Exception is not swallowed though.

Pankraty:
All remarks have been addressed, I believe. Ready for another round ;)

ClosedXML/Excel/Ranges/XLRangeAddress.cs
igitur: You're right. I forgot it's a struct. But this change helps for the static analysis tools to stop complaining and simplifies the code. And it simplifies the code. Thanks

ClosedXML/Excel/XLWorksheet.cs
igitur: Oh, I see... no `catch` block.

ClosedXML_Tests/Excel/Caching/SampleRepositoryTests.cs
```diff
@@ -1,10 +1,8 @@
...
 using ClosedXML.Excel.Caching;
 using NUnit.Framework;
-using System;
```
igitur: Add back this `using`. It's needed for the Release configuration GC tests. Or explicitly use `System.WeakReference` in the test itself, so that we can omit the `using`.

igitur:
GC tests are failing under Release config.

igitur:
So just to confirm, there are no breaking API changes in this PR?

PR #780, opened 2018-04-14
Preserve chart

igitur:
tldr: Preserves charts upon saving. Doesn't interpret any charts.

Long:
ClosedXML deletes worksheets' drawingparts when nothing in the drawingparts are used. Previously only pictures were considered "used". This PR adds charts. We might have to consider other shapes too in the future, or remove the code that tests whether to delete the drawingspart.


ClosedXML_Tests/Excel/Saving/SavingTests.cs
```diff
@@ -254,5 +254,20 @@ public void CanSaveFileWithPictureAndComment()
                 wb.SaveAs(ms);
             }
         }
+
+        [Test]
+        public void PreserveChartsWhenSaving()
```
Pankraty: What do you think about adding another test for the opposite - say, delete all pictures and make sure no empty Drawing.xml is presented?

igitur: Added new test below this one.

Pankraty (approved):

bthnosma:
Hello @igitur,

I have similar problem to that https://stackoverflow.com/questions/49814281/closedxml-charts-missing-after-saving#comment86765047_49814281.

All drawings in my excel being removed when I edit excel with closedxml. Is it possible for you to develop a solution?

PR #782, opened 2018-04-15
Small bugfixes and improvements

Pankraty:
After I ran a code analysis tool I found a few issues - some are more important, others are less.

The obvious bug here is fixed with commit e9c22cec85cbc17ddbde31e526a0ffb35cb29676. Aside from this,  some checks were reorganized (removed, added or added), ```IDisposable``` implemented for classes containing disposable members, other small enhancements made.

### DIFF
--- a/ClosedXML/Excel/AutoFilters/XLAutoFilter.cs
+++ b/ClosedXML/Excel/AutoFilters/XLAutoFilter.cs
@@ -116,62 +116,59 @@ namespace ClosedXML.Excel
             SortColumn = columnToSortBy;
 
             // Recalculate shown / hidden rows
-            if (Enabled)
-            {
                 using (var rows = Range.Rows(2, Range.RowCount()))
                 {
                     foreach (IXLRangeRow row in rows)
                         row.WorksheetRow().Unhide();
                 }
 
-                foreach (var kp in Filters)
+            foreach (var kp in Filters)
+            {
+                Boolean firstFilter = true;
+                foreach (XLFilter filter in kp.Value)
                 {
-                    Boolean firstFilter = true;
-                    foreach (XLFilter filter in kp.Value)
+                    var condition = filter.Condition;
+                    var isText = filter.Value is String;
+                    var isDateTime = filter.Value is DateTime;
+
+                    using (var rows = Range.Rows(2, Range.RowCount()))
                     {
-                        var condition = filter.Condition;
-                        var isText = filter.Value is String;
-                        var isDateTime = filter.Value is DateTime;
-
-                        using (var rows = Range.Rows(2, Range.RowCount()))
+                        foreach (IXLRangeRow row in rows)
                         {
-                            foreach (IXLRangeRow row in rows)
+                            //TODO : clean up filter matching - it's done in different place
+                            Boolean match;
+
+                            if (isText)
+                                match = condition(row.Cell(kp.Key).GetFormattedString());
+                            else if (isDateTime)
+                                match = row.Cell(kp.Key).DataType == XLDataType.DateTime && condition(row.Cell(kp.Key).GetDateTime());
+                            else
+                                match = row.Cell(kp.Key).DataType == XLDataType.Number && condition(row.Cell(kp.Key).GetDouble());
+
+                            if (firstFilter)
                             {
-                                //TODO : clean up filter matching - it's done in different place
-                                Boolean match;
-
-                                if (isText)
-                                    match = condition(row.Cell(kp.Key).GetFormattedString());
-                                else if (isDateTime)
-                                    match = row.Cell(kp.Key).DataType == XLDataType.DateTime && condition(row.Cell(kp.Key).GetDateTime());
+                                if (match)
+                                    row.WorksheetRow().Unhide();
                                 else
-                                    match = row.Cell(kp.Key).DataType == XLDataType.Number && condition(row.Cell(kp.Key).GetDouble());
-
-                                if (firstFilter)
-                                {
-                                    if (match)
-                                        row.WorksheetRow().Unhide();
-                                    else
-                                        row.WorksheetRow().Hide();
-                                }
-                                else
-                                {
-                                    if (filter.Connector == XLConnector.And)
-                                    {
-                                        if (!row.WorksheetRow().IsHidden)
-                                        {
-                                            if (match)
-                                                row.WorksheetRow().Unhide();
-                                            else
-                                                row.WorksheetRow().Hide();
-                                        }
-                                    }
-                                    else if (match)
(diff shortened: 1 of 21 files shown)


ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -477,7 +477,7 @@ private void LoadSpreadsheetDocument(SpreadsheetDocument dSpreadsheet)
                                 continue;
                         }
 
-                        if (target != null && source != null)
+                        if (source != null)
```
igitur: I assume the static analysis tool said that `target` cannot be `null` here. So, for checks like these, I prefer to keep the check, to future-proof it. It's possible that something happens in the future that changes code higher up and then we would need the check. Rather than introducing a possible regression in the future, I'd prefer to keep this check her. It doesn't hurt to keep it.

ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -1361,12 +1361,9 @@ private static void ParseReference(string item, out string sheetName, out string
                 else
                     formula = cell.CellFormula.Text;
 
-                if (cell.CellFormula.Reference != null)
-                {
-                    // Parent cell of shared formulas
-                    // Child cells will use this shared index to set its R1C1 style formula
-                    xlCell.FormulaReference = ws.Range(cell.CellFormula.Reference.Value).RangeAddress;
-                }
+                // Parent cell of shared formulas
```
igitur: Why did you remove this `if`?

ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -1956,47 +1953,47 @@ private static void LoadAutoFilterColumns(AutoFilter af, XLAutoFilter autoFilter
 
                         if (xlDateGroupFilter.DateTimeGrouping >= XLDateTimeGrouping.Year)
                         {
-                            if (dateGroupItem?.Year?.HasValue ?? false)
+                            if (dateGroupItem.Year?.HasValue ?? false)
```
igitur: You don't have to change this one, but again, including the `?` is my personal preference to future-proof the code.

ClosedXML/Excel/XLWorkbook_Save.cs
```diff
@@ -3476,7 +3477,7 @@ private void AddDifferentialFormats(WorkbookStylesPart workbookStylesPart, SaveC
             var differentialFormat = new DifferentialFormat();
 
             var diffFont = GetNewFont(new FontInfo { Font = cf.Style.Font as XLFont }, false);
-            if (diffFont?.HasChildren ?? false)
+            if (diffFont.HasChildren)
```
igitur: Definitely keep the `?` and `??`, please. I have been working on branches that might return `null`s. Not sure if I'm going to merge them, but again, I'd rather keep the code future-proof. Please revert these changes for all the "style" variables in these methods.

ClosedXML/XLHelper.cs
```diff
@@ -116,8 +116,10 @@ internal static int TrimRowNumber(int rowNumber)
 
         public static bool IsValidColumn(string column)
         {
+            if (String.IsNullOrWhiteSpace(column))
+                return false;
             var length = column.Length;
-            if (String.IsNullOrWhiteSpace(column) || length > 3)
+            if (length > 3)
```
igitur: Probably not even necessary for a `length` variable?

ClosedXML/Excel/Rows/XLRow.cs
```diff
@@ -336,7 +336,7 @@ public IXLRow AdjustToContents(Int32 startColumn, Int32 endColumn, Double minHei
                         else
                         {
                             Double rotation;
-                            if (textRotation == 90 || textRotation == 180 || textRotation == 255)
+                            if (textRotation == 90 || textRotation == 180)
```
igitur: Actually, I have no idea why there is a special case for `textRotation == 255` a few lines back.

ClosedXML/Excel/AutoFilters/XLAutoFilter.cs
```diff
@@ -116,62 +116,59 @@ public XLAutoFilter Sort(Int32 columnToSortBy, XLSortOrder sortOrder, Boolean ma
...
-            if (Enabled)
-            {
                 using (var rows = Range.Rows(2, Range.RowCount()))
                 {
                     foreach (IXLRangeRow row in rows)
                         row.WorksheetRow().Unhide();
                 }
 
-                foreach (var kp in Filters)
+            foreach (var kp in Filters)
```
igitur: Fix indentation of the preceding block.

ClosedXML/Excel/AutoFilters/XLAutoFilter.cs
```diff
@@ -116,62 +116,59 @@ public XLAutoFilter Sort(Int32 columnToSortBy, XLSortOrder sortOrder, Boolean ma
             SortColumn = columnToSortBy;
 
             // Recalculate shown / hidden rows
-            if (Enabled)
```
igitur: Why did you remove this? It seems like a good check to have?

Pankraty: At the top of the method, there is already 
`if (!Enabled) throw new InvalidOperationException("Filter has not been enabled.");`

igitur: ah ok

ClosedXML/Excel/XLWorkbook_Load.cs
Pankraty: OK, got it. Will restore it.

ClosedXML/Excel/XLWorkbook_Load.cs
Pankraty: Here we are already inside the condition 
`if (cell.CellFormula != null && cell.CellFormula.SharedIndex != null && cell.CellFormula.Reference != null)`

igitur: It would benefit me to expand the preceding code blocks! :-)

ClosedXML/XLHelper.cs
Pankraty: It is used in the code below so I think it's OK to leave it.

ClosedXML/Excel/Rows/XLRow.cs
Pankraty: `textRotation == 255` stands for a vertical orientation, so it is natural to determine a height for this case independently.
image
I wonder if this special case was ever tested?..

igitur:
Thanks!

PR #786, opened 2018-04-17
Implemented support of storing cell calculated values using DateTime

Pankraty:
This is the alternative to #679 using ```DateTime``` value of last modification date instead of a ```long``` counter. After we choose which one to use I will delete a benchmark from the Sandbox.


igitur:
Those 23 tests fail on my machine too. I'll inspect.

igitur:
As discussed, the .NET DateTime.UTC isn't fine-grained enough to handle this. This is why the unit tests are failing randomly.

Issue #787, opened 2018-04-18
Use 'in' and 'ref' for value types where applicable


Pankraty:
**Do you want to request a *feature* or report a *bug*?**
- [ ] Bug
- [ ] Feature
- [x] Enhancement

**Version of ClosedXML**

e.g. 0.93

**What is the current behavior?**

With a new approach to styles and addresses processing structs became widely used in ClosedXML.

C# 7.2 introduces new features regarding value types (https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-7-2#reference-semantics-with-value-types) so we can have some use of them after CloseXML is upgraded to C# 7.2. Specifically, marking input parameters with ```in``` keyword looks promising for me - on the synthetic benchmark I achieved about 34% of performance improvement with this.

For now, we can park this issue.

igitur:
Does it work on dotnet40?

>

Pankraty:
Yes

PR #788, opened 2018-04-18
Change projects to use C# 7.2 by default.

igitur:
Change projects to use C# 7.2 by default.

I can't think of any possible side-effects, except that developers have to have a fairly recent version of Visual Studio or VS Code.


igitur:
Addresses part of #787

Pankraty:
I'm OK with it. Do you have any preferences for which new language features we can use and which we'd better avoid? I.e. out variables, pattern matching, tuple types, etc.

igitur:
I don't know it well enough to really give a good opinion. I like the new `out` parameter syntax. I like pattern matching. But let's review on a case by case basis.

PR #789, opened 2018-04-18
Use inline out parameters provided by C# 7.2

igitur:
Syntactic sugar: Use inline out parameters provided by C# 7.2
Also consolidate some nested `if` statements that are now possible.

### DIFF
--- a/ClosedXML/Excel/XLWorkbook_Load.cs
+++ b/ClosedXML/Excel/XLWorkbook_Load.cs
@@ -1227,8 +1227,7 @@ namespace ClosedXML.Excel
                         }
                         else
                         {
-                            string sheetName, sheetArea;
-                            ParseReference(area, out sheetName, out sheetArea);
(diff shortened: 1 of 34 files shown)


ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -1227,8 +1227,7 @@ private void LoadDefinedNames(Workbook workbook)
                         }
                         else
                         {
-                            string sheetName, sheetArea;
-                            ParseReference(area, out sheetName, out sheetArea);
+                            ParseReference(area, out String sheetName, out String sheetArea);
```
Pankraty: Just a side note: later this method may be redefined to return a tuple ```(string shetName, string sheetArea)```.

igitur: Yes, we should review each of 7.2's additions and see where we can use it. I know of some other places where we can use tuples too.

PR #790, opened 2018-04-18
Improve XLCell.SetValue(object)

igitur:
Improve XLCell.SetValue(object) by removing unnecessary TryParse functions

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -218,57 +218,127 @@ namespace ClosedXML.Excel
             }
 
             var style = GetStyleForRead();
+            Boolean parsed;
+            string parsedValue;
+
+            // For SetValue<T> we set the cell value directly to the parameter
+            // as opposed to the other SetValue(object value) where we parse the string and try to decude the value
             if (value is String || value is char)
             {
-                _cellValue = value.ToString();
+                parsedValue = value.ToInvariantString();
                 _dataType = XLDataType.Text;
-                if (_cellValue.Contains(Environment.NewLine) && !style.Alignment.WrapText)
+                if (parsedValue.Contains(Environment.NewLine) && !style.Alignment.WrapText)
                     Style.Alignment.WrapText = true;
-            }
-            else if (value is TimeSpan)
-            {
-                _cellValue = value.ToString();
-                _dataType = XLDataType.TimeSpan;
-                if (style.NumberFormat.Format == String.Empty && style.NumberFormat.NumberFormatId == 0)
-                    Style.NumberFormat.NumberFormatId = 46;
-            }
-            else if (value is DateTime)
-            {
-                _dataType = XLDataType.DateTime;
-                var dtTest = (DateTime)Convert.ChangeType(value, typeof(DateTime));
-                if (style.NumberFormat.Format == String.Empty && style.NumberFormat.NumberFormatId == 0)
-                    Style.NumberFormat.NumberFormatId = dtTest.Date == dtTest ? 14 : 22;
 
-                _cellValue = dtTest.ToOADate().ToInvariantString();
-            }
-            else if (value.GetType().IsNumber())
-            {
-                if ((value is double || value is float) && (Double.IsNaN((Double)Convert.ChangeType(value, typeof(Double)))
-                    || Double.IsInfinity((Double)Convert.ChangeType(value, typeof(Double)))))
-                {
-                    _cellValue = value.ToString();
-                    _dataType = XLDataType.Text;
-                }
-                else
-                {
-                    _dataType = XLDataType.Number;
-                    _cellValue = ((Double)Convert.ChangeType(value, typeof(Double))).ToInvariantString();
-                }
-            }
-            else if (value is Boolean)
-            {
-                _dataType = XLDataType.Boolean;
-                _cellValue = (Boolean)Convert.ChangeType(value, typeof(Boolean)) ? "1" : "0";
+                parsed = true;
             }
             else
             {
-                _cellValue = Convert.ToString(value);
-                _dataType = XLDataType.Text;
+                var tuple = SetKnownTypedValue(value, style);
+                parsedValue = tuple.Item1;
+                parsed = tuple.Item2;
             }
+
+            // If parsing was unsuccessful, we throw an ArgumentException
+            // because we are using SetValue<T> (typed).
+            // Only in SetValue(object value) to we try to fall back to a value of a different type
+            if (!parsed)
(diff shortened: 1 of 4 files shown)


Pankraty:
Please wait before #679 is merged. I rebased it ~6 times already )

igitur:
So you should be able to rebase in your sleep by now!

>

Pankraty:
:-P 
I do actually, but with rebasing, I should resolve merge conflicts too ) That's not fair!

vbjay:
https://hackernoon.com/fix-conflicts-only-once-with-git-rerere-7d116b2cec67

Pankraty:
Thanks, @vbjay! In this particular case, the problem was that the same code has been modified for a few times during the PR life, but in some other PRs this trick will be very useful.

igitur:
This PR is ready for review. `XLCell.SetValue` has been refactored substantially, so the diffs won't make sense. I do believe the method is a lot more readable now.

ClosedXML_Tests/Excel/Cells/XLCellTests.cs
```diff
@@ -369,6 +369,36 @@ public void ValueSetDateWithShortUserDateFormat()
             Assert.AreEqual(expected, actual);
         }
 
+        [Test]
+        public void SetStringCellValues()
```
Pankraty: I would suggest adding a few more test cases:

* Specify an empty string
* Specify a string longer than 32767 characters
* Specify null
* Specify number as a string
* Specify DateTime less than `BaseDate`
* Specify DateTime, Decimal, Integer, TimeSpan, Boolean without converting to strings
* Specify `FormulaA1` and then set cell value and ensure that both `FormulaA1` and `FormulaR1C1` are empty
* Specify a string having line break included and ensure that `WrapText` is set to true
* Specify a string containing a single quote and number (e.g. `'10`) and make sure the `DataType` is Text.

Perhaps some of those cases are already covered with tests but they can help not to get into regression in future.

igitur:
Ready for another review. Made more improvements.

igitur:
Breaking change: setting a cell value using `.SetValue()` to a date that's lower than the base date will throw an `ArgumentException`.

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -218,57 +218,125 @@ internal IXLCell SetValue<T>(T value, bool setTableHeader)
...
+            {
+                parsedValue = b ? "1" : "0";
+                _dataType = XLDataType.Boolean;
+                parsed = true;
+            }
+            else if (value.IsNumber())
+            {
+                if ((value is double || value is float)
+                    && (Double.IsNaN((Double)Convert.ChangeType(value, typeof(Double)))
+                        || Double.IsInfinity((Double)Convert.ChangeType(value, typeof(Double)))))
```
Pankraty:
```
if ((value is double d && (double.IsNaN(d) || double.IsInfinity(d)) ||
    (value is float f && (float.IsNaN(f) || float.IsInfinity(f))
```
looks cleaner and does not require `(Double)Convert.ChangeType(value, typeof(Double))` to be executed twice.

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -218,57 +218,125 @@ internal IXLCell SetValue<T>(T value, bool setTableHeader)
...
+                throw new ArgumentException($"Unable to set cell value to {value.ToInvariantString()}", nameof(value));
+
+            SetInternalCellValueString(parsedValue, validate: true, parseToCachedValue: false);
+
+            CachedValue = null;
+
+            return this;
+        }
+
+        private Tuple<string, bool> SetKnownTypedValue<T>(T value, XLStyleValue style)
```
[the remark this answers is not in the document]
igitur: Tried that, but requires new libraries to be referenced (ValueTuple) and couldn't get it working for netstandard, so left it deliberately until we implement ValueTuple in many places.

ClosedXML/Excel/Cells/XLCell.cs
igitur: True. And I must add a test for this.

ClosedXML/Excel/Cells/XLCell.cs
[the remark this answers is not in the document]
Pankraty: Got it.

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -434,12 +502,18 @@ public Object Evaluate(Boolean force = false)
...
-            SetInternalCellValueString(cellValue, this.HasFormula);
+            SetInternalCellValueString(cellValue, validate: false, parseToCachedValue: this.HasFormula);
         }
 
-        private void SetInternalCellValueString(String cellValue, Boolean parseToCachedValue)
+        private void SetInternalCellValueString(String cellValue, Boolean validate, Boolean parseToCachedValue)
         {
+            if (validate)
+            {
+                if (cellValue.Length > 32767) throw new ArgumentException("Cells can hold a maximum of 32,767 characters.");
```
Pankraty: BTW, `ArgumentOutOfRangeException` would be more specific and it is inherited from `ArgumentException` so it is safe to replace it (no external code would be broken). Same for dates.
Up to you.

ClosedXML/Excel/Cells/XLCell.cs
igitur: done.

igitur:
Ready for review.

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -2021,85 +2098,69 @@ private void ClearMerged()
...
             }
-            if (val.Length > 32767) throw new ArgumentException("Cells can only hold 32,767 characters.");
 
-            if (SetTableHeaderValue(val)) return;
-            if (SetTableTotalsRowLabel(val)) return;
+            if (SetTableHeaderValue(parsedValue)) return;
+            if (SetTableTotalsRowLabel(parsedValue)) return;
+
+            SetInternalCellValueString(parsedValue, validate: true, parseToCachedValue: false);
+            CachedValue = null;
```
Pankraty: `CachedValue` won't become null if `SetTableHeaderValue(parsedValue)` or `SetTableTotalsRowLabel(parsedValue)` succeed. I am not sure but this seems wrong. Better move it to the top where `FormulaA1` and `_richText` are reset.

igitur: `SetTableHeaderValue` sets `XLTableField.Name`, which calls `XLCell.SetValue<T>`(T value, setTableHeader: false)`, which will set `CachedValue` to null.

PR #794, opened 2018-04-19
Set XLCell.CachedValue from file

igitur:
Populate XLCell.CachedValue from file and use it in other operation. 
And simplify the lots of parsing in XLCell


ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -1288,7 +1317,25 @@ private IEnumerable<XLCell> GetAffectingCells()
...
+            {
+                if (!HasFormula && cachedValue == null)
+                    cachedValue = Value;
+
+                return cachedValue;
+            }
+            private set
+            {
+                if (value != null && !HasFormula)
+                    throw new InvalidOperationException("Cached values can be set only for cells with formulas");
```
Pankraty: Maybe add a test for this?

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -493,7 +536,11 @@ public object Value
                 }
 
                 var cellValue = HasRichText ? _richText.ToString() : _cellValue;
-                return StringToCellDataType(cellValue);
+                var parsedValue = ParseCellValueFromString(cellValue, _dataType, out String error);
+                if ("" == error)
+                    return parsedValue;
+                else
+                    throw new ArgumentException(error);
```
Pankraty: I am not fond of throwing exceptions in getter, especially assuming that the caller do not pass any arguments and can be very surprised with `ArgumentException`. I didn't find the way to get this exception (maybe if a workbook contains errors?) but still this looks dangerous.

Pankraty:
Consider this.

```
            {
                var c = ws.Cell("A1");
                c.Value = "ABC123";
                c.DataType = XLDataType.Number;
                Console.WriteLine(c.Value);
            }
```

Before this PR this code would fail at `c.DataType = XLDataType.Number;`. With this PR it will silently set `c.Value` to `String.Empty`. I think it's wrong.

igitur:
Good points. I'll look into it.

ClosedXML/Excel/Cells/XLCell.cs
igitur: Well, currently that exception cannot occur. It's a private setter, and where it's used it's always either `value == null` or `HasFormula == true`. Included this more for future-proofing the code. I don't really want to change the setter to `internal` just for the sake of testing an exception that doesn't currently occur. You ok with that?

ClosedXML/Excel/Cells/XLCell.cs
igitur: You're right. I shouldn't be throwing an exception here.

igitur:
Added that piece of code as a new unit test in #835

ClosedXML/Excel/Cells/XLCell.cs
Pankraty: OK

PR #798, opened 2018-04-20
Speed up AdjustToContents

### LINKED ISSUE #796: Method AdjustToContents is very slow on sheets with many merged ranges

- [x] Bug
- [ ] Feature

**Version of ClosedXML**
0.92

Calling method `AdjustToContents` on a sheet with many merged ranges is painfully slow. That's because every cell in the range being adjusted is searched for in _every_ merged range, making the complexity of the operation O(n^2).

The method should be faster - for example by keeping a `HashSet` of all merged cells in the target range, so that each individual check is a O(1) operation, making the whole method a O(n) operation.

No.

## Reproducibility
Attached is a sample [sheet](https://github.com/ClosedXML/ClosedXML/files/1931041/test.xlsx) with 20000 merged ranges. Calling the code below takes (in a Debug build) over 150 seconds.

```c#
using( var wb = new XLWorkbook("test.xlsx") )
{
    wb.Worksheet( 1 ).ColumnsUsed().AdjustToContents();
}
```

crypto-rsa:
Fixes #796.

Introduces a cache for merged cell addresses in class `XLRanges`. It is only updated once, ie. subsequent changes in merged ranges are not considered. Therefore, it should only be used in a context where we are sure the merged ranges will not change. To this end, I also added class `XLWorksheetMergedCellsCalculatorWrapper` (feel free to suggest a shorter name 😄) which runs the calculation in the constructor and resets the cache in the `Dispose` method, allowing a simple usage with the `using` statement.
In the sheet from the linked issue, the time to run `AdjustToContents` has dropped from 150 to 2 seconds.

### DIFF
--- a/ClosedXML/Excel/Ranges/XLRanges.cs
+++ b/ClosedXML/Excel/Ranges/XLRanges.cs
@@ -10,6 +10,8 @@ namespace ClosedXML.Excel
     {
         private readonly List<XLRange> _ranges = new List<XLRange>();
 
+        private HashSet<IXLAddress> _mergedCellAddresses;
+
         public XLRanges() : base(XLWorkbook.DefaultStyleValue)
         {
         }
@@ -50,7 +52,7 @@ namespace ClosedXML.Excel
         /// </summary>
         /// <param name="match">Criteria to filter ranges. Only those ranges that satisfy the criteria will be removed.
         /// Null means the entire collection should be cleared.</param>
-        /// <param name="releaseEventHandlers">Specify whether or not should removed ranges be unsubscribed from 
+        /// <param name="releaseEventHandlers">Specify whether or not should removed ranges be unsubscribed from
         /// row/column shifting events. Until ranges are unsubscribed they cannot be collected by GC.</param>
         public void RemoveAll(Predicate<IXLRange> match = null, bool releaseEventHandlers = true)
         {
@@ -82,6 +84,9 @@ namespace ClosedXML.Excel
 
         public Boolean Contains(IXLCell cell)
         {
+            if (_mergedCellAddresses != null)
+                return _mergedCellAddresses.Contains(cell.Address);
+
             return _ranges.Any(r => r.RangeAddress.IsValid && r.Contains(cell));
         }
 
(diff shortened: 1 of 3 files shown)


igitur:
Thanks. LGTM. I just notice some whitespace issues. Does your IDE support editorconfig?

crypto-rsa:
I use Visual Studio 2017 so yes, it supports `.editorconfig` - but it just specifies the indentation whitespace. Other spacing (for example after control keywords) is not specified in the file - and my default VS settings are very different from the convention used in the repository. So I couldn't just run "Format document" and I had to fix the spaces manually which perhaps wasn't perfect.

crypto-rsa:
I will happily improve the whitespace if you point me to the specific lines.

igitur:
There are still a few places where the whitespace is weird. Should I point them out?

vbjay:
Why not just update editor config with correct settings and then just run
format document?

igitur:
@vbjay Because my weekend starts in 10minutes and then I'm outta here.

igitur:
But is that kind of config even possible in editorconfig?

crypto-rsa:
https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference

ClosedXML/Excel/Ranges/XLRanges.cs
```diff
@@ -256,5 +261,15 @@ public IXLRanges Consolidate()
             var engine = new XLRangeConsolidationEngine(this);
             return engine.Consolidate();
         }
+
+        public void CalculateMergedCells()
+        {
+            _mergedCellAddresses = new HashSet<IXLAddress>(_ranges.Where(r => r.RangeAddress.IsValid).SelectMany(r => r.Cells().Select(c => c.Address)));
```
Pankraty: This will cause a huge memory allocation for large merged ranges. Consider this:
```
            {
                ws.Range("A:C").Merge();
                ws.Range("A1:C1").Value = "ASDFGHJ";
                ws.Columns("A:C").AdjustToContents();
            }
```
With this PR it takes more than 7 seconds to execute, and cost 1 GB of RAM while without the PR it takes only about 0.7 sec.

On the other hand, having many small merged ranges is much more common than having a few huge ones. So maybe it's fine.

crypto-rsa: Good point(s). I'll give it a thought.

igitur:
> https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference

Awesome thanks! Never knew that Visual Studio supports additional parameters. Will definitely implement some of them. Enjoy the weekend, guys.

igitur:
Superceded by #803

xxbigbacon:
I am running into this as well, was this ever resolved? trying to adjust columns 1 to 35 of 1200 rows takes over 7 minutes! pulling my hair out wondering my data can get pulled from SQL and dumped in the worksheet in like 30 seconds yet the file wasn't ready for another 7.

PR #804, opened 2018-04-23
Catch the exceptions when the formula is not valid (i.e. shifted too far)

### LINKED ISSUE #631: Unhandled exception when saving (etc) after copying a cell containing a 'left pointing' formula
- [x] Bug
- [ ] Feature

**Version of ClosedXML**
0.90

An incorrect formula can be 'generated' in ClosedXML by copying a formula cell, with a left pointing reference, to a cell far enough to the left, making the resulting referenced cell to have a column number below 1.
As a result, an unhandled exception is thrown in GetColumnLetterFromNumber, which impacts a lot of functionality: saving is not possible, adjusttocontents is not working, etc.

The Excel GUI uses '=#REF!' to replace the invalid cell formula when doing a similar copy action.
What would be the best approach in ClosedXML ?

Unknown

## Reproducibility

```vb.net
        Dim theWorkbook = New Excel.XLWorkbook()
        Dim theSheet1 = theWorkbook.Worksheets.Add("Sheet1")
        theSheet1.Cell("A1").Value = "ValueA1"
        theSheet1.Cell("B2").FormulaA1 = "=A1"
        theSheet1.Range("A2").Value = theSheet1.Range("B2")
        theWorkbook.SaveAs(Path.Combine(Application.StartupPath, "issue7b_out.xlsx"), False, False)
```

Pankraty:
This PR fixes #631.

Here I wrap `GetFormulaA1` into a try-catch block to prevent exceptions when formula became invalid due to shifting. We discussed that maybe we have to throw an exception at the moment such formula occurs but I assumed that since the formula could come from the saved file too and we actually support these improper formulas it is safer to return `#REF!` rather than throw an exception.

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -2072,36 +2072,43 @@ namespace ClosedXML.Excel
             if (String.IsNullOrWhiteSpace(strValue))
                 return String.Empty;
 
-            var value = ">" + strValue + "<";
-
-            var regex = conversionType == FormulaConversionType.A1ToR1C1 ? A1Regex : R1C1Regex;
-
-            var sb = new StringBuilder();
-            var lastIndex = 0;
-
-            foreach (var match in regex.Matches(value).Cast<Match>())
+            try
             {
-                var matchString = match.Value;
-                var matchIndex = match.Index;
-                if (value.Substring(0, matchIndex).CharCount('"') % 2 == 0
-                    && value.Substring(0, matchIndex).CharCount('\'') % 2 == 0)
+                var value = ">" + strValue + "<";
+
+                var regex = conversionType == FormulaConversionType.A1ToR1C1 ? A1Regex : R1C1Regex;
+
+                var sb = new StringBuilder();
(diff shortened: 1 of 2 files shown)


igitur:
Ideally this test should pass, but I think it's a much bigger job:
```c#
[Test]
public void InvalidFormulaShiftProducesREF()
{
    {
        {
            ws.Cell("A1").Value = 1;
            ws.Cell("B1").Value = 2;
            ws.Cell("B2").FormulaA1 = "=A1+B1";
            Assert.AreEqual(3, ws.Cell("B2").Value);

            ws.Range("A2").Value = ws.Range("B2");
            var fA2 = ws.Cell("A2").FormulaA1;

            wb.SaveAs(ms);

            Assert.AreEqual("=#REF!+A1", fA2);
        }

        using (var wb2 = new XLWorkbook(ms))
        {
            var fA2 = wb2.Worksheets.First().Cell("A2").FormulaA1;
            Assert.AreEqual("=#REF!+A1", fA2);
        }
    }
}
```

igitur:
Superceded by #814

PR #806, opened 2018-04-23
Add LibreOffice unit test files

igitur:
Adds a test to load all the .xlsx and .xlsm files that I found in the LibreOffice test suite.
Doesn't attempt to save and compare the files.

Unsurprisingly, some of the files fail to load.


ClosedXML_Tests/Excel/Loading/LoadingTests.cs
```diff
@@ -44,6 +44,38 @@ public void CanSuccessfullyLoadFiles()
             }
         }
 
+        [Test]
+        public void CanSuccessfullyLoadLOFiles()
+        {
+            // TODO: unpark all files
+            var parkedForLater = new[]
```
igitur: @Pankraty Not urgent, but at some stage we should investigate one by one why theses files don't load in ClosedXML. I see we have to do some work on pivot tables. That doesn't surprise me. Maybe you can look at the first file, `formats.xlsx`, since you've been doing a lot of work on it.

Pankraty: OK. Created a separate task to not forget about it.

PR #809, opened 2018-04-24
 Add support for named ranges referring to expressions or multiple ranges

### LINKED ISSUE #137: Named constants throw exception.
If I assign via MS Excel for RefersTo of named range does not range address but the value and load the file into XLWorkbook, I get an exception when trying to calculate the value of a cell referring to this range.
image
Example xlsx: 
[7_vars.xlsx](https://github.com/ClosedXML/ClosedXML/files/684721/7_vars.xlsx)

Pankraty:
This PR fixes #137 

When evaluating the literal during formula evaluation I check if the specified identifier is presented in the collection of named ranges (either worksheet's or workbook's) and then check where it refers to. If the reference is not a valid range address then I call a calc engine to evaluate the expression.

In addition, I found that evaluation of expressions that contain named ranges referring to multiple ranges (e.g. `=SUM(NAMEDRANGE)` where `NAMEDRANGE` stands for `A1:C1,A3:C3`) did not work too. This part is also fixed.


ClosedXML_Tests/Excel/NamedRanges/NamedRangesTests.cs
```diff
@@ -162,10 +163,92 @@ public void NamedRangesWhenCopyingWorksheets()
...
+                    Assert.AreEqual(0.1, (double) ws1.Cell(1, 1).Value, XLHelper.Epsilon);
+                    Assert.AreEqual(1.0, (double) ws1.Cell(2, 1).Value, XLHelper.Epsilon);
+                    Assert.AreEqual(0.2, (double) ws1.Cell(3, 1).Value, XLHelper.Epsilon);
+                    Assert.AreEqual(0.6, (double) ws1.Cell(4, 1).Value, XLHelper.Epsilon);
+                }
+            }
+        }
+
+        [Test]
+        public void CanEvalateNamedMultiRange()
```
igitur: CanEvaluateNamedMultiRange

ClosedXML/Excel/CalcEngine/XLCalcEngine.cs
```diff
@@ -95,11 +95,36 @@ public override object GetExternalObject(string identifier)
...
+
                 return GetCellRangeReference(_ws.Range(identifier));
+            }
             else
                 return identifier;
         }
 
+        private bool TryGetNamedRange(string identifier, IXLWorksheet worksheet, out IXLNamedRange namedRange)
+        {
+            Func<string, string, bool> equals = (a, b) =>
```
igitur: `bool equals(string a, string b) => string.Equals(a, b, StringComparison.InvariantCultureIgnoreCase);` 
simpler?

igitur: Hmm, or it's kind of the same complexity.

igitur:
Mind squashing the commits and rebasing? I want to try to keep the history cleaner, especially now that big changes are happening.

Pankraty:
No problem

PR #810, opened 2018-04-24
Easier pivot tables from tables

### LINKED ISSUE #799: New helper method: IXLTable.SummarizeWithPivotTable()

igitur:
Fixes #799 

Turns out there is already an `IXLRangeBase.CreatePivotTable()` method, but it wasn't working correctly. This can be used for subclasses like `XLTable` too.


ClosedXML/Excel/PivotTables/IXLPivotTable.cs
```diff
@@ -126,7 +129,10 @@ public interface IXLPivotTable
         String RowHeaderCaption { get; set; }
 
         IXLCell TargetCell { get; set; }
+
         IXLRange SourceRange { get; set; }
+        IXLTable SourceTable { get; set; }
```
Pankraty: Apply auto formatting with CodeMaid once more?

ClosedXML/Excel/PivotTables/XLPivotTable.cs
```diff
@@ -27,7 +27,39 @@ public XLPivotTable(IXLWorksheet worksheet)
...
+        }
+
+        public IXLTable SourceTable
+        {
+            get
+            {
+                if (this.SourceType == XLPivotTableSourceType.Table)
+                    return SourceRange as IXLTable;
+                else
+                    throw new InvalidOperationException("This pivot table's source is not a table.");
```
Pankraty: I would prefer `return null` but do not insist.

ClosedXML/Excel/PivotTables/IXLPivotTable.cs
igitur: I did. And redid it now. No change. What looks wrong?

ClosedXML/Excel/PivotTables/XLPivotTable.cs
igitur: Done.

ClosedXML/Excel/PivotTables/IXLPivotTable.cs
Pankraty: All the properties have empty lines between them but these three stand tight so I thougth this line was added after the auto formatting was performed.

igitur: Oh, I thought you meant the indentation or something was wrong. It seems CodeMaid doesn't add new lines between elements of interfaces. I left these three grouped together because they're kind of related.

Pankraty: Ah, OK then

igitur: image

This is why CodeMaid doesn't fix that. Going to leave it as is for now.

PR #811, opened 2018-04-24
Add and implement XLNamedRangeScope

igitur:
Named Ranges can be at Worksheet or Workbook scope. Currently we use private properties and check for `Worksheet == null` to deduce the scope. This is a more explicit implementation.


ClosedXML/Excel/NamedRanges/XLNamedRanges.cs
```diff
@@ -70,14 +73,18 @@ internal IXLNamedRange Add(String rangeName, String rangeAddress, String comment
                 {
                     if (XLHelper.IsValidRangeAddress(rangeAddress))
                     {
-                        var range = Worksheet?.Range(rangeAddress) ?? Workbook.Range(rangeAddress);
+                        IXLRange range = null;
+                        if (Scope == XLNamedRangeScope.Worksheet)
+                            range = Worksheet.Range(rangeAddress);
+                        else if (Scope == XLNamedRangeScope.Workbook)
+                            range = Workbook.Range(rangeAddress);
```
Pankraty: I'd also add
```
else
    throw new NotSupportedException($"Scope {Scope} is not supported");
```

Issue #267, opened 2017-04-10
.xltx / .xltm support


Dryhoof:
**Do you want to request a *feature* or report a *bug*?**
- [ ] Bug
- [x] Feature

**What is the current behavior?**
Currently the library only supports xlsx / xlsm files.

**What is the expected behavior or new feature?**
Would be very nice to be able to open xltm files!

**Did this work in previous versions of our tool?  Which versions?**
Not that I know! :D

igitur:
Are those file types also the same zipped OpenXML format as `.xlsx` and `.xlsm`? If so, it should be easy to extend.

Dryhoof:
Goood question, i have no idea :D

There certainly are differences, because Excel itself will not allow me to open a file that was saved as xltm after changing it to xlsm by hand...

Saving a opened xltm as xlsm would also open up new problems, I guess.

PR #816, opened 2018-04-26
Template fix

StevenGodin:
Fixes saving as .xltx and .xltm files. #267 

The `SpreadsheetDocumentType` was not being set correctly for these file types, and Excel would not open them.


ClosedXML_Tests/Excel/Saving/SavingTests.cs
```diff
@@ -9,6 +9,8 @@
 using System.Linq;
 using System.Reflection;
 using System.Threading;
+using DocumentFormat.OpenXml;
```
igitur: Please sort the `using` statements alphabetically.

igitur:
Nice catch, thanks. Please see the minor review comment.

PR #817, opened 2018-04-26
Fix determining the FirstColumnUsed

### LINKED ISSUE #339: RangeUsed() returns wrong start column if first cells of first row are blank
RangeUsed() returns wrong start column when the first used row starts with some blank cells.
Example: with data like this (A1 and B1 blank):
,,1,2,3
1,2,3,4,5
1,2,3,4,5

RangeUsed() returns C1:E3 instead of A1:E3.

Sample test code:
```c#
            var sheet = wb.AddWorksheet("page1");

            sheet.Range(1, 3, 1, 5).Value = "row1";
            sheet.Range(2, 1, 2, 5).Value = "row2";
            sheet.Range(3, 1, 3, 5).Value = "row3";
            var used = sheet.RangeUsed().ToString();
            sheet.Cell(2, 8).Value = used;

            string path = @"test.xlsx";
            wb.SaveAs(path);
```

Pankraty:
Fixes #339 
I can't believe it has not been unfixed for so long!

The original algorithm failed in cases like this:
```
| | |1|
| |1|1|
|1|1|1|
```

### DIFF
--- a/ClosedXML/Excel/Cells/XLCellsCollection.cs
+++ b/ClosedXML/Excel/Cells/XLCellsCollection.cs
@@ -212,22 +212,28 @@ namespace ClosedXML.Excel
         {
             int finalRow = rowEnd > MaxRowUsed ? MaxRowUsed : rowEnd;
             int finalColumn = columnEnd > MaxColumnUsed ? MaxColumnUsed : columnEnd;
+            int firstColumnUsed = finalColumn;
+            var found = false;
             for (int ro = rowStart; ro <= finalRow; ro++)
             {
                 if (rowsCollection.TryGetValue(ro, out Dictionary<Int32, XLCell> columnsCollection))
                 {
-                    for (int co = columnStart; co <= finalColumn; co++)
+                    for (int co = columnStart; co <= firstColumnUsed; co++)
                     {
                         if (columnsCollection.TryGetValue(co, out XLCell cell)
                             && !cell.IsEmpty(includeFormats)
-                            && (predicate == null || predicate(cell)))
-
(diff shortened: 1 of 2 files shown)


ClosedXML/Excel/Cells/XLCellsCollection.cs
```diff
@@ -212,22 +212,28 @@ public int FirstColumnUsed(int rowStart, int columnStart, int rowEnd, int column
...
-                    for (int co = columnStart; co <= finalColumn; co++)
+                    for (int co = columnStart; co <= firstColumnUsed; co++)
                     {
                         if (columnsCollection.TryGetValue(co, out XLCell cell)
                             && !cell.IsEmpty(includeFormats)
-                            && (predicate == null || predicate(cell)))
-
-                            return co;
+                            && (predicate == null || predicate(cell))
+                            && co <= firstColumnUsed)
```
zybexXL: this line is not needed, it's always true

ClosedXML/Excel/Cells/XLCellsCollection.cs
```diff
@@ -212,22 +212,28 @@ public int FirstColumnUsed(int rowStart, int columnStart, int rowEnd, int column
...
+                        {
+                            firstColumnUsed = co;
+                            found = true;
+                            break;
```
zybexXL: you can optimize this for most use cases - big speed improvement for files with many rows:
```
if (co == columnStart) return co;
break;
```

zybexXL:
Thank you for fixing this :)

Issue #808, opened 2018-04-24
Fix issues with opening files from LibreOffice test suite


Pankraty:
Continuation of #806.

Quote:
>Not urgent, but at some stage, we should investigate one by one why these files don't load in ClosedXML. I see we have to do some work on pivot tables. That doesn't surprise me. Maybe you can look at the first file, `formats.xlsx`, since you've been doing a lot of work on it.

Pankraty:
From the collection of files that fail to load in ClosedXML

1. 11 files relate to pivot tables and I don't feel brave enough to cope with them
2. 2 files are invalid (`formats.xlsx` and `tdf100709.xlsx`). ClosedXML is not designed to manage corrupted files so I suggest removing them.
3. 2 files have comments of non-rectangular shape (`universal-content-strict.xlsx` and `universal-content.xlsx`) and wait for #645.
4. `xf_default_values.xlsx` has pictures with the duplicate ids. Apparently, duplicate IDs are allowed in XLSX so I removed the validation that threw exceptions on attempts to assign the existing ID to the picture.

igitur:
I feel that pivot tables should be refactored. The current code is quite messy.

Issue #444, opened 2017-08-11
When opening the given file, ClosedXML crash with the message 'The picture name 'Group 1' already exists.'


Tarboeuf:
**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

**ClosedXML 0.88.0**

**What is the current behavior?**

When opening the given file, ClosedXML crash with the message 'The picture name 'Group 1' already exists.'
This file can be opened in Excel 2016.

**Did this work in previous versions of our tool?  Which versions?**
This worked with the version 0.87.1

- [x] I attached a sample spreadsheet.  (You can drag files on to this issue)
[Headers.xlsx](https://github.com/ClosedXML/ClosedXML/files/1218379/Headers.xlsx)

**Code to reproduce problem:**
```c#
public void Main()
{
   using (XLWorkbook workbook = new XLWorkbook(@"Headers.xlsx"))
   {
   }
}
```

igitur:
Wow, I didn't know Excel allows duplicate shape names. It's going to take some time to sort that out. In the meantime, I suggest you remove the duplicate group box or rename it to something else.

igitur:
[ClosedXML v0.89.0-beta1](https://www.nuget.org/packages/ClosedXML/0.89.0-beta1) has been published and is ready for your testing.

PR #820, opened 2018-04-27
Correctly handle files with duplicate picture ids. XLPictures refactoring

Pankraty:
This PR partially addresses #808 

1. 2 files are invalid (`formats.xlsx` and `tdf100709.xlsx`). ClosedXML is not designed to manage corrupted files so I suggest removing them.
2. `xf_default_values.xlsx` has pictures with the duplicate ids. Apparently, duplicate IDs are allowed in XLSX so I removed the validation that threw exceptions on attempts to assign the existing ID to the picture.


kenleese:
@Pankraty Should ClosedXml have a deterministic behavior for invalid / corrupted files? If so, those files could be the basis of a useful test case.

ClosedXML/Excel/Drawings/XLPicture.cs
```diff
@@ -140,17 +139,7 @@ public Int32 Height
...
-            get { return id; }
-            internal set
-            {
-                if ((_worksheet.Pictures.FirstOrDefault(p => p.Id.Equals(value)) ?? this) != this)
-                    throw new ArgumentException($"The picture ID '{value}' already exists.");
-
-                id = value;
-            }
-        }
+        public Int32 Id { get; internal set; }
```
igitur: Private set?

igitur:
I remember how surprised I was to find out picture names were not unique. Now I'm even more surprised that picture IDs are not unique. What's the point of having an ID if it's not unique.

Anyway, compare this to #819, which is about a fix for having unique IDs for the non visual drawing properties. I

igitur:
For the corrupt files, does OpenXML throw an exception? If so we could handle it. But otherwise it's difficult to tell whether a loading problem will be due to a corrupt file or a ClosedXML bug.

vbjay:
I'd just st as well fix it in load. When a duplicate I'd is found, replace
it with a valid one.

ClosedXML/Excel/Drawings/XLPicture.cs
Pankraty: With `private set`, we won't be able to assign the value read from a file.

igitur: Sorry, stupid me.

igitur:
Please run those files through the validator first.

Pankraty:
The validator says "Attribute 'id' should have unique value. Its current value '1' duplicates with others".

vbjay:
Question. Does there need to be any relationship fixing?

igitur:
This changes the names of pictures if there are duplicates. This could possibly break embedded VBA code, e.g. `ActiveSheet.Shapes("Picture 2").IncrementLeft 30`. If it's a choice between supporting files that LibreOffice can load/save and not breaking Excel VBA code, I think I'm leaning towards keeping the VBA code safe.

Pankraty:
> This could possibly break embedded VBA code,

The thing is, when such file is opened in Excel, the pictures with duplicate names receive new ones, so we won't make it worse.

igitur:
@Pankraty Are you sure? In #444 there is an attached file with images with duplicate names `Picture 2`. If I open this in Excel and resave it, the names remain `Picture 2`.

Pankraty:
You are right. For `xf_default_values.xlsx` it behaves differently but probably that's because in this file pictures have the same IDs as well.

Shall we park it for now?

igitur:
Or maybe because in #444 's file the pictures are inside groups. But even the groups' names are duplicates. Yes, let's park.

Issue #824, opened 2018-05-01
cannot read string #N/A


davidbirchwork:
## Read and complete the full issue template

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

<package id="ClosedXML" version="0.92.1" targetFramework="net40" />

**What is the current behavior?**

When trying to read a spreadsheet cells with the text "#N/A" are read as blank
yes this is a weird thing to do I was seeking to write a unit test comparing ClosedXML with COM Interop - most things pass so far :-)

**What is the expected behavior or new feature?**

read the text "#N/A"

**Did this work in previous versions of our tool?  Which versions?**

Unknown

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

Without a code sample, it is unlikely that your issue will get attention. Don't be lazy. Do the effort and assist the developers to reproduce your problem. Code samples should be [minimal complete and verifiable](https://stackoverflow.com/help/mcve). Sample spreadsheets should be attached whenever applicable. Remove sensitive information.

**Code to reproduce problem:**
```c#
public void Main()
{
    var _wb = new XLWorkbook(fname);
    if(_wb.Cell("Sheet1!A1").isEmpty()) Console.WriteLine(":(");
    var str = _wb.Cell("Sheet1!A1").GetString();
    if(str != "#N/A") Console.WriteLine(":(");
}
```
- [x] I attached a sample spreadsheet.  (You can drag files on to this issue)
[error.xlsx](https://github.com/ClosedXML/ClosedXML/files/1963407/error.xlsx)

Pankraty:
What version of ClosedXML do you use?
I have a NullReferenceException at `_wb.Cell("Sheet1!A1").isEmpty()`.
If I change to `_wb.Worksheet("Sheet1").Cell("A1")` the reported issue reproduces too.

igitur:
Latest version on NuGet is v 0.93.0-beta2. Please test with that.

Pankraty:
> Latest version on NuGet is v 0.93.0-beta2. Please test with that.

It fails with the latest version as well. I just wondered how could `_wb.Cell("Sheet1!A1").isEmpty()` ever work. In code, I see that the method `XLWorkbook.Cell` did change for almost a year, and it does not work with cell addresses - only with named ranges.

davidbirchwork:
[...]

PR #825, opened 2018-05-01
Get cell by full address

Pankraty:
As revealed in #824, the cell cannot be accessed by its full address. E.g.
```
wb.Range("Sheet1!A1:B2");               // returns range
wb.Ranges("Sheet1!A1:B2,Sheet1!C1:E3"); // returns a collection of ranges
wb.Cell("Sheet1!C123");                 // returns null
```

In method `Cell()` we only searched for named ranges, not parsing the cell address.
This PR fixes this and adds a few unit tests for all these methods.

### DIFF
--- a/ClosedXML/Excel/XLWorkbook.cs
+++ b/ClosedXML/Excel/XLWorkbook.cs
@@ -353,12 +353,24 @@ namespace ClosedXML.Excel
             if (!rangeAddress.Contains('!')) return null;
 
             var split = rangeAddress.Split('!');
-            var first = split[0];
-            var wsName = first.StartsWith("'") ? first.Substring(1, first.Length - 2) : first;
-            var localRange = split[1];
+            var wsName = split[0].UnescapeSheetName();
             if (TryGetWorksheet(wsName, out ws))
             {
-                return ws.Range(localRange);
+                return ws.Range(split[1]);
+            }
+            return null;
+        }
+
+        public IXLCell CellFromFullAddress(String cellAddress, out IXLWorksheet ws)
+        {
+            ws = null;
+            if (!cellAddress.Contains('!')) return null;
+
+            var split = cellAddress.Split('!');
+            var wsName = split[0].UnescapeSheetName();
+            if (TryGetWorksheet(wsName, out ws))
+            {
+                return ws.Cell(split[1]);
             }
             return null;
         }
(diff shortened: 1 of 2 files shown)


igitur:
I vaguely remember a discussion from the old Codeplex days that this wasn't implemented to encourage users rather to use `wb.Worksheet("Sheet1").Cell("A1")` style lookups. I think that would be slightly faster. But I still agree that `wb.Cell("Sheet1!C123");` should be implemented. We'll need some metrics to verify the initial claim, but if it's true, maybe add some notes to the doctext.

ClosedXML/Excel/XLWorkbook.cs
```diff
@@ -353,12 +353,24 @@ public IXLRange RangeFromFullAddress(String rangeAddress, out IXLWorksheet ws)
...
 
             var split = rangeAddress.Split('!');
-            var first = split[0];
-            var wsName = first.StartsWith("'") ? first.Substring(1, first.Length - 2) : first;
-            var localRange = split[1];
+            var wsName = split[0].UnescapeSheetName();
             if (TryGetWorksheet(wsName, out ws))
             {
-                return ws.Range(localRange);
+                return ws.Range(split[1]);
```
igitur: I think we should check `split.Length` first. If `rangeAddress` is something like `Sheet1!`, we'd want an informative message in the exception.

ClosedXML/Excel/XLWorkbook.cs
```diff
@@ -353,12 +353,24 @@ public IXLRange RangeFromFullAddress(String rangeAddress, out IXLWorksheet ws)
...
+        public IXLCell CellFromFullAddress(String cellAddress, out IXLWorksheet ws)
+        {
+            ws = null;
+            if (!cellAddress.Contains('!')) return null;
+
+            var split = cellAddress.Split('!');
+            var wsName = split[0].UnescapeSheetName();
+            if (TryGetWorksheet(wsName, out ws))
+            {
+                return ws.Cell(split[1]);
```
igitur: Same here.

ClosedXML_Tests/Excel/Misc/XLWorkbookTests.cs
```diff
@@ -74,6 +74,85 @@ public void Cells3()
             Assert.AreEqual(1, cells.First().GetValue<Int32>());
         }
 
+        [Test]
+        public void GetCellFromFullAddress()
```
igitur: Maybe just include a test for a sheet name with some spaces or other oddities?

igitur:
I think this kind of parsing should be extracted so that `XLWorkbook` and `XLWorksheet` can share the logic. But let's leave that for later.

ClosedXML/Excel/XLWorkbook.cs
Pankraty: There is a safeguard already:
`if (!cellAddress.Contains('!')) return null;`

And if split `sheet1!` we still have an array of two elements (the second is an empty string. Added a few more test cases.

ClosedXML_Tests/Excel/Misc/XLWorkbookTests.cs
Pankraty: Done

PR #830, opened 2018-05-02
Change call .ToString() to .ToInvariantString()

DyachenkoA:
Change call .ToString() to .ToInvariantString() in save workbook.


igitur:
Thanks for the contribution. Can you please give more context and maybe a unit case which displays what this is trying to solve?

DyachenkoA:
In Russian locale test "LoadFormulaCachedValue" fail.
When I began to understand why, I found at line 5566 call .ToString (), which caused the test to fail and replaced it with a call .ToInvatiantString (). Then looked for places where similar error was found in locations also changed the call .ToString () to the call .ToInvatiantString ().

DyachenkoA:
ru-RU

igitur:
@DyachenkoA Merge https://github.com/DyachenkoA/ClosedXML/pull/1 , then we're good to go.

PR #833, opened 2018-05-04
Treat warnings as errors

igitur:
Depends on #794

### DIFF
--- a/ClosedXML/Excel/Ranges/XLRangeFactory.cs
+++ b/ClosedXML/Excel/Ranges/XLRangeFactory.cs
@@ -46,10 +46,8 @@ namespace ClosedXML.Excel
                     return CreateTable(key.RangeAddress);
 
                 case XLRangeType.Worksheet:
-                //break;
                 default:
                     throw new NotImplementedException(key.RangeType.ToString());
-                    break;
             }
         }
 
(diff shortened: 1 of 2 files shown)


igitur:
Only once we remove `ValueCached` can we implement this.

Pankraty:
> Only once we remove ValueCached can we implement this.

Not necessarily. If we mark all the methods (test including) where ValueCached is used as [Obsolete] too Visual Studio will be fine with this.

igitur:
`ValueCached` is populated when a file is loaded, so we'll then have to mark basically everything as obsolete.

Pankraty:
> `ValueCached` is populated when a file is loaded, so we'll then have to mark basically everything as obsolete.

Again, not necessarily ) We can populate an internal property and make `ValueCached` to be a mere wrapper for it OR we can add an exception for this particular warning for this part of code (https://docs.microsoft.com/ru-ru/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-pragma-warning) or project-wide.

igitur:
No, too much effort for something small like this. Let's just park this until later and in the meantime we can fix warnings ourselves instead of forcing them as errors.

Pankraty:
Ehm... I pushed my commit into your branch accidentally. I did not mean it =-)

igitur:
How do you even have access to do that?

Pankraty:
> How do you even have access to do that?

image
So easy 😄

PR #839, opened 2018-05-07
Fix issues with copying worksheets across workbooks

### LINKED ISSUE #829: IXLWorksheet CopyTo does not copy well
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

e.g. 0.93.0-beta2

When copying a worksheet from a XLTX to a XLSX workbook (it looks like there is the same issue with from a XLSX), the IXLWorksheet's `CopyTo(XLWorkbook workbook, string newSheetName);` method does not:

- keep the row height in the new workbook
- keep the font and font size in all the empty cell

**What is the expected behavior?**

I expect this method to strictly copy the worksheet without changing anything within.

Yes, this used to work at least in version 0.88.0 (.NET 4.5.2)

## Reproducibility

```c#
public void Main()
{
     using (var workbookDocument = new XLWorkbook())
     using (var templateWB = new XLWorkbook(@"Test.xltx"))
     //using (var templateWB = new XLWorkbook(@"Test.xlsx"))
     {
         var ws= templateWB.Worksheet(1);
         ws.CopyTo(workbookDocument, "Copy");
         workbookDocument.SaveAs("Test.xlsx", true);
     }
}
```
- [inputs.zip](https://github.com/ClosedXML/ClosedXML/files/1966487/inputs.zip)

Pankraty:
This PR addresses #829.

Here I added unit tests covering different aspects of worksheets copying and implemented missing functionality.
To simplify the worksheet's method `CopyTo` I extracted part of the logic into corresponding classes (XLTable, XLPicture).

### DIFF
--- a/ClosedXML/Excel/ConditionalFormats/XLConditionalFormat.cs
+++ b/ClosedXML/Excel/ConditionalFormats/XLConditionalFormat.cs
@@ -149,17 +149,7 @@ namespace ClosedXML.Excel
         {
             if (targetRange != null)
                 Ranges.Add(targetRange);
-
-            ConditionalFormatType = conditionalFormat.ConditionalFormatType;
-            TimePeriod = conditionalFormat.TimePeriod;
-            IconSetStyle = conditionalFormat.IconSetStyle;
-            Operator = conditionalFormat.Operator;
-            Bottom = conditionalFormat.Bottom;
-            Percent = conditionalFormat.Percent;
-            ReverseIconOrder = conditionalFormat.ReverseIconOrder;
-            ShowIconOnly = conditionalFormat.ShowIconOnly;
-            ShowBarOnly = conditionalFormat.ShowBarOnly;
-            StopIfTrue = OpenXmlHelper.GetBooleanValueAsBool(conditionalFormat.StopIfTrue, true);
+            CopyFrom(conditionalFormat);
         }
         #endregion Constructors
 
@@ -220,6 +210,19 @@ namespace ClosedXML.Excel
         {
             this.StopIfTrue = value;
             return this;
+        }
+
+        public IXLConditionalFormat CopyTo(IXLWorksheet targetSheet)
+        {
+            var targetRange = targetSheet.Range(Range.RangeAddress.WithoutWorksheet());
+            var newCf = new XLConditionalFormat(this, targetRange);
+            if (Ranges.Count > 1)
+            {
+                Ranges.Skip(1).ForEach(r => newCf.Ranges.Add(targetSheet.Range(r.RangeAddress)));
+            }
+
+            targetSheet.ConditionalFormats.Add(newCf);
+            return newCf;
         }
 
         public void CopyFrom(IXLConditionalFormat other)
--- a/ClosedXML/Excel/ConditionalFormats/XLConditionalFormat.cs
+++ b/ClosedXML/Excel/ConditionalFormats/XLConditionalFormat.cs
@@ -234,7 +234,7 @@ namespace ClosedXML.Excel
             return this;
         }
 
-        internal IXLConditionalFormat CopyTo(IXLWorksheet targetSheet)
+        public IXLConditionalFormat CopyTo(IXLWorksheet targetSheet)
         {
             if (targetSheet == Range?.Worksheet)
                 throw new InvalidOperationException("Cannot copy conditional format to the worksheet it already belongs to.");
--- a/ClosedXML/Excel/ConditionalFormats/XLConditionalFormat.cs
+++ b/ClosedXML/Excel/ConditionalFormats/XLConditionalFormat.cs
@@ -1,7 +1,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using ClosedXML.Utils;
 
 namespace ClosedXML.Excel
 {
@@ -22,14 +21,14 @@ namespace ClosedXML.Excel
 
             public bool Equals(IXLConditionalFormat x, IXLConditionalFormat y)
             {
-                var xx = (XLConditionalFormat) x;
-                var yy = (XLConditionalFormat) y;
+                var xx = (XLConditionalFormat)x;
+                var yy = (XLConditionalFormat)y;
                 if (ReferenceEquals(xx, yy)) return true;
                 if (ReferenceEquals(xx, null)) return false;
                 if (ReferenceEquals(yy, null)) return false;
                 if (xx.GetType() != yy.GetType()) return false;
 
-                var xxValues = xx.Values.Values.Where(v => !v.IsFormula).Select(v=>v.Value);
+                var xxValues = xx.Values.Values.Where(v => !v.IsFormula).Select(v => v.Value);
                 var yyValues = yy.Values.Values.Where(v => !v.IsFormula).Select(v => v.Value);
                 var xxFormulas = x.Ranges.Any() ? xx.Values.Values.Where(v => v.IsFormula).Select(f => ((XLCell)x.Ranges.First().FirstCell()).GetFormulaR1C1(f.Value)) : null;
                 var yyFormulas = y.Ranges.Any() ? yy.Values.Values.Where(v => v.IsFormula).Select(f => ((XLCell)y.Ranges.First().FirstCell()).GetFormulaR1C1(f.Value)) : null;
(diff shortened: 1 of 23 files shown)


Pankraty:
Ooops )

ClosedXML/Excel/ConditionalFormats/XLConditionalFormat.cs
```diff
@@ -222,6 +212,19 @@ public IXLConditionalFormat SetStopIfTrue(bool value)
             return this;
         }
 
+        public IXLConditionalFormat CopyTo(IXLWorksheet targetSheet)
+        {
+            var targetRange = targetSheet.Range(Range.RangeAddress.WithoutWorksheet());
```
igitur: Do you think that all `CopyTo` methods should first check that `targetSheet` is different to the "current sheet"? For example, it doesn't make sense to copy a conditional format from a sheet into the same sheet.

ClosedXML/Excel/Tables/XLTable.cs
```diff
@@ -817,5 +817,39 @@ public DataTable AsNativeDataTable()
...
+        {
+            return CopyTo((XLWorksheet) targetSheet);
+        }
+
+        internal IXLTable CopyTo(XLWorksheet targetSheet)
+        {
+            String tableName = Name;
+            var newTable = (XLTable)targetSheet.Table(targetSheet.Range(RangeAddress.ToString()), tableName, true);
+
+            newTable.RelId = RelId;
```
igitur: The `RelId` is to store the `relId` when a workbook is opened. When creating a new table (by copying it), the `RelId` should remain empty.

ClosedXML/Excel/Tables/XLTable.cs
```diff
@@ -817,5 +817,39 @@ public DataTable AsNativeDataTable()
...
         }
+
+        public IXLTable CopyTo(IXLWorksheet targetSheet)
+        {
+            return CopyTo((XLWorksheet) targetSheet);
+        }
+
+        internal IXLTable CopyTo(XLWorksheet targetSheet)
+        {
+            String tableName = Name;
```
igitur: Should copying a table not also involve copying the contents of the cells in the table, i.e. the range should be copied too?

ClosedXML_Tests/Excel/Worksheets/XLWorksheetTests.cs
```diff
@@ -1,8 +1,11 @@
 using ClosedXML.Excel;
 using NUnit.Framework;
 using System;
+using System.Drawing;
 using System.IO;
 using System.Linq;
+using System.Reflection;
+using ClosedXML.Excel.Drawings;
```
igitur: Alphabetical

ClosedXML/Excel/ConditionalFormats/XLConditionalFormat.cs
Pankraty: True. Added a validation and two more unit tests.

ClosedXML/Excel/ConditionalFormats/XLConditionalFormat.cs
```diff
@@ -222,6 +212,19 @@ public IXLConditionalFormat SetStopIfTrue(bool value)
...
+            var newCf = new XLConditionalFormat(this, targetRange);
```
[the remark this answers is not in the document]
Pankraty: Good idea.

ClosedXML/Excel/Drawings/XLPicture.cs
```diff
@@ -326,6 +326,44 @@ public IXLPicture WithSize(Int32 width, Int32 height)
...
         }
 
+        public IXLPicture CopyTo(IXLWorksheet targetSheet)
+        {
+            return CopyTo((XLWorksheet) targetSheet);
+        }
+
+        internal IXLPicture CopyTo(XLWorksheet targetSheet)
+        {
+            var newPicture = targetSheet.AddPicture(ImageStream, Format, Name)
```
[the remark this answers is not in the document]
Pankraty: This is actually done in `XLPicture` constructor:
```
            this.ImageStream = new MemoryStream();
            {
                stream.Position = 0;
                stream.CopyTo(ImageStream);
                ImageStream.Seek(0, SeekOrigin.Begin);
```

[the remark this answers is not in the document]
igitur: I can't remember my own code! I still feel we should change to byte array. But that's a separate issue.

ClosedXML/Excel/Tables/XLTable.cs
Pankraty: Perhaps you are right. I just moved this from `XLWorksheet` class. And this code was quite ancient (2011, commit message "worksheet.CopyTo(...) should now work as expected." LOL)

igitur: If it's from 2011, then it's before my time. Let me just double-check my previous statement in this regard first.

ClosedXML/Excel/Tables/XLTable.cs
Pankraty: As long as we use this method in the scope of worksheet copying only it is not needed as cells are copied separately. But if we would like to use this method separately, we'd have to perform cells copying as well. Although I do not see any use cases that would require it.

igitur: Well, you included it in `IXLTable.cs`, same with the other `CopyTo` methods. So people are going to use it. And I think it's a good idea to have this publicly exposed.

ClosedXML_Tests/Excel/Worksheets/XLWorksheetTests.cs
```diff
@@ -296,5 +300,522 @@ public void WorksheetNameCanContainApostrophe()
...
+                var ws1 = wb1.Worksheets.Add("Original");
+
+                ws1.Cell("A1").FormulaA1 = "10*10";
+                ws1.Cell("A2").FormulaA1 = "Original!A1 * 3";
+
+                ws1.Name = "Renamed";
+
+                Assert.AreEqual("Renamed!A1 * 3", ws1.Cell("A2").FormulaA1);
+            }
+        }
```
igitur: Also add individual tests for e.g. `IXLTable.CopyTo` in the file for testing tables. Same with the other objects.

ClosedXML/Excel/Tables/XLTable.cs
Pankraty: I'll add another test tomorrow.

Pankraty: When I remove this line `CopyWorksheets` test fails. The produced file is correct either way, just table IDs differ.

ClosedXML_Tests/Excel/Worksheets/XLWorksheetTests.cs
Pankraty: Test for tables, pictures and named ranges added. Conditional formats already had them.

Pankraty:
The first set of feedback processed.

ClosedXML/Excel/Drawings/IXLPicture.cs
```diff
@@ -8,6 +8,14 @@ public interface IXLPicture : IDisposable
     {
         IXLAddress BottomRightCellAddress { get; }
 
+        /// <summary>
+        /// Create a copy of the picture.
+        /// </summary>
+        /// <param name="targetSheet">A worksheet where put the copy to. When null the copy of the picture is created
```
igitur: I don't think we should allow `targetSheet == null`. I would rather have a separate `IXLPicture.Duplicate()` method, which does the same.

Also update the parameter doctext to `The worksheet to which the picture will be copied.`

ClosedXML/Excel/Drawings/IXLPicture.cs
```diff
@@ -8,6 +8,19 @@ public interface IXLPicture : IDisposable
     {
         IXLAddress BottomRightCellAddress { get; }
 
+        /// <summary>
+        /// Create a copy of the picture on a different worksheet.
+        /// </summary>
+        /// <param name="targetSheet">The worksheet to which the picture will be copied.</param>
+        /// <returns>A created copy of the picture.</returns>
+        IXLPicture CopyTo(IXLWorksheet targetSheet = null);
```
igitur: Make the parameter non-optional, please.

igitur:
Great work, again.

Pankraty:
Thanks )

Issue #843, opened 2018-05-11
To get merged cell's range.


mnts26:
## Read and complete the full issue template

**Do you want to request a *feature* or report a *bug*?**
- [ x] feature

If you are logging a possible bug or feature request, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.75.0.0

**What is the current behavior?**
I have following excel files and i want get to know those merged cell is range. 
![image](https://user-images.githubusercontent.com/331587/39926571-8bde8f3e-556a-11e8-8b7f-d22a13ce39fd.png)

**What is the expected behavior or new feature?**
for example:
CELL: HEADER IS (A1:G4)
          : BODY IS (A5:G15) 
          : FOOTER IS (A16:G21)  

**Did this work in previous versions of our tool?  Which versions?**

No

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

**Code to reproduce problem:**
```c#
public void ReadExcel(string path, int sheet_number)
        {
            try
            {
                temp_wb = new XLWorkbook(path);
                temp_wsh = temp_wb.Worksheet(sheet_number);
                IXLRange work_sheet_range = temp_wsh.RangeUsed(true);
                max_row = work_sheet_range.RangeAddress.LastAddress.RowNumber;
                max_col = work_sheet_range.RangeAddress.LastAddress.ColumnNumber;
                for (int row = 1; row <= max_row ; row++)
                {
                    list_of_kjs_cell = new List<IXLCell>();
                    for (int col = 1; col <= max_col ; col++)
                    {
                        IXLCell cell = temp_wsh.Cell(row, col); 
                        
                        if(!string.isNullOrEmpty(cell.GetString()) && cell.IsMerged()){
                            // In here how to get that cell is range?
                        }

                        
                    }
                }
                temp_wb.Dispose();
            }
            catch (System.IO.IOException e)
            {
                Console.WriteLine(e.Message);
            }
        }
```
- [ ] I attached a sample spreadsheet.  (You can drag files on to this issue)

Pankraty:
On the latest version you can to do this:
`temp_wsh.MergedRanges.GetIntersectedRanges(cell.Address).FirstOrDefault();`

On previous version there should be 
`temp_wsh.MergedRanges.Where(r => r.Contains(cell)).FirstOrDefault();`

[...]

PR #845, opened 2018-05-12
Update XLCell.MergedRange to use GetIntersectedRanges

igitur:
As discussed in #843 . The new way is more efficient, right, @Pankraty ?


Pankraty:
Yes, especially when there are plenty of merged ranges

igitur:
> Maybe do the same for conditional formats and data validation rules

To which lines are you referring?

Pankraty:
On the other hand, the effect of it will be hardly noticable (in contrast to `MergedRange` optimization)

igitur:
Ready for review. Last one for 0.93

PR #856, opened 2018-05-16
Clear pivot table rendered range upon loading to avoid conflict on re…

### LINKED ISSUE #850: Pivot tables are broken on Save
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.92.1 (also reproduced on 0.93.0-beta4)

When opening and saving a file that contains a pivot table, something happens to the pivot table. When we open the file with Excel the following dialog appears: "There's already data in [TestTable.xlsx]Sheet1. Do you want to replace it?".
If we click the "Yes" button, the pivot table seems to be properly restored.
If we click the "No" button, the pivot table gets removed. The data is still there, but it's not part of a pivot table anymore.

Nothing should happen to the file, especially since nothing was done to it: no write or otherwise file-altering operations.

Yes, it worked on 0.85.0. On 0.86.0 the library threw a System.NullReferenceException.
The observed behavior started in version 0.87.1. I tested all versions from 0.87.1 to 0.93.0-beta4 and reproduced this issue on all of them with the same code.

## Reproducibility

```c#
public static void Main(string[] args)
{
    using (XLWorkbook wb = new XLWorkbook("TestTable.xlsx"))
    {
        wb.Save();
    }
}
```
[TestTable.xlsx](https://github.com/ClosedXML/ClosedXML/files/2001189/TestTable.xlsx)

igitur:
Fixes #850 

This PR is a workaround until we have a real pivot table renderer (which is a long way away).

Upon loading a file with an existing pivot table, it clears the range (contents only) of the current pivot table's cells.

Initially I thought to do this only upon saving, but I didn't want to create a situation where ClosedXML reports a cell as having content (that was loaded), but then when it is saved and reopened with ClosedXML the content changes. That's not good.


igitur:
Needs unit test.

Issue #778, opened 2018-04-12
Can support dynamic charts? Make the cell black。


nathanQiu1:
## Read and complete the full issue template

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.93.0

**What is the current behavior?**

The dynamic chart disappeared，Cell color changed to black。

**What is the expected behavior or new feature?**

Cell color is  normal， dynamic chart is normal。

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

Without a code sample, it is unlikely that your issue will get attention. Don't be lazy. Do the effort and assist the developers to reproduce your problem. Code samples should be [minimal complete and verifiable](https://stackoverflow.com/help/mcve). Sample spreadsheets should be attached whenever applicable. Remove sensitive information.

**Code to reproduce problem:**
```c#
public void Main()
{
using (XLWorkbook workbook = new XLWorkbook(@"FAI.xlsm"))
   {
       workbook.Save();
   }
}
```
- [x] I attached a sample spreadsheet.  (You can drag files on to this issue)
[dynamiccharts.xlsx](https://github.com/ClosedXML/ClosedXML/files/1903078/dynamiccharts.xlsx)

nathanQiu1:
Updated the latest code to support charts. the black part still exists.

Pankraty:
@nathanQiu1 please check the attached file. From my perspective, it looks OK.

[778_result.xlsx](https://github.com/ClosedXML/ClosedXML/files/2019816/778_result.xlsx)

nathanQiu1:
@Pankraty   thank you for your reply!  After the file passes closedxml, there will be a black area.
![image](https://user-images.githubusercontent.com/19723900/40285703-d4dc4f0a-5cd1-11e8-816d-3cc8e1e501dc.png)

Pankraty:
The file I attached came from ClosedXML with fixes from #862. If there are no other issues with this file then this bug will be closed when 862 is merged.

nathanQiu1:
OK, thank you very much!

igitur:
Fixed in #862

PR #862, opened 2018-05-19
Correctly process transparent background (Fix #290)

### LINKED ISSUE #290: Copy BackgroundColor from one IXLCell to each IXLCell in an IXLRange
- [x] Bug
- [ ] Feature

Unpredictable BackgroundColor value assignment. 

I try to use a default cell (A1) to have a Cell in the Default-Excel-Style, which I can then use to delete/reset Cells of different Ranges (A2:E2). (I haven't found another way to fully reset a cell, if there is, please tell me.) 

Not all Cells of Range (A2:E2) get the color of that origin Cell (A1), sometimes none (A2:E2 stay colored), sometimes only a few (only A2 gets the A1 Style, or A2 and B2 but not the rest, or A2 and C2 but not the rest, I have seen it all) and sometimes even the "origin Cell" gets the "destination Cell" BackgroundColor.  
Please have a look at code comments below.

**If the current behavior is a bug, please provide the steps to reproduce and 
if possible a minimal demo of the problem with a sample spreadsheet.**
- I open a new Excel File, 
- select Range A2:E2, 
- hit the arrow Button next to the FillColor bucket in Excel,
- and choose a color of the shown "Designcolors"(<- german version)  palette 
- select another cell on the sheet
- Hit File -> save as "C:\MyTemp\empty.xlsx"
- Run Code in release or debug mode

It worked for weeks with 
> cell.Value = "";
> cell.Style = DefaultCell.Style;
> cell.Style.Font.FontColor = XLColor.Black;
>   cell.Style.Alignment.Horizontal = XLAlignmentHorizontalValues.General;
now it wont. I don't know why, I havent changed the code or the file. (did a few runs with backup files with the same result)
In my production xlsx File (which I cannot provide) the origin Cell gets the color of a Range cell, and assigns that color to the whole Range too, I can't reproduce this behavior in my Testproject.

[empty.xlsx](https://github.com/ClosedXML/ClosedXML/files/962121/empty.xlsx)

```c#
        {
            String Filename = @"C:\MyTemp\empty.xlsx";
            String Worksheet = "Tabelle1";
            using (XLWorkbook workbook = new XLWorkbook(Filename, XLEventTracking.Disabled))
            using (IXLWorksheet sheetXML = workbook.Worksheet(Worksheet))
            {
                IXLCell defaultcell = sheetXML.Cell(1, 1);
                IXLRange setColorRange = sheetXML.Range(2, 1, 2, 5);
                foreach (IXLCell cell in setColorRange.Cells())
                {
                    //cell.Style.Fill.SetBackgroundColor(defaultcell.Style.Fill.BackgroundColor); //leaves black background

                    //cell.Style = defaultcell.Style; 
                    //Works in this szenario, but not (anymore) in my "live.xlsx" which I cannot provide, sorry.
                    //the defaultcell AND all Range cells get the blue color

                    //cell.Style.Fill = defaultcell.Style.Fill; 
                    //Unpredictable behavior, most of the times A2 gets it right, rest of Range stays blue
                    //sometimes it manages to fill B2 and C2, sometimes only A2 and C2.... 
                    //I tried it with a new File each time and with the existing File. 

                    //cell.Style.Fill.BackgroundColor = defaultcell.Style.Fill.BackgroundColor; 
                    // right now it leaves a black background color, (tested with new file)
                    // usually I get the same result as I do with the cell.Style.Fill aproach (which makes sence)
                }
                workbook.Save(false);
            }
        }
```

package id="ClosedXML" version="0.87.1" targetFramework="net45"
package id="DocumentFormat.OpenXml" version="2.5" targetFramework="net45"
Microsoft Office Standard 2010 -> Excel: Version: 14.0.7173.5000 (32-Bit)
Visual Studio 2015 community Edition

Please let me know, if anything is unclear or more information is needed, thanks.

Pankraty:
This PR fixes #290 and #778.

In this PR I modified a logic of comparing the FillKeys so that two empty fills were considered the same no matter how their "emptiness" is specified - with PatternType.None, or with Indexed color 64, or with XLColor.NoColor.

After this was done I had to modify the logic for a simultaneous change of the `BackgroundColor` and `PatternType`. Without it, the cached "empty" fill was used even though the pattern type has changed to `Solid` (if the color remained Indexed(64) the fill still was empty) or the color changed to some non-transparent (the pattern remained `None` and fill was empty again).

To prevent the empty fills from saving to file the additional checks were added. Unlike some other entities (say, `patternFill`), for colors we should check `.HasAttributes` rather than `.HasChildren` - colors never have children.

Thanks to @igitur for the POC.


ClosedXML/Excel/Style/Colors/XLColor_Static.cs
```diff
@@ -169,9 +169,20 @@ public static XLColor FromTheme(XLThemeColor themeColor, Double themeTint)
             }
         }
 
-        private static XLColor noColor = new XLColor();
+        internal static bool IsNullOrTransparent(XLColor color)
```
igitur: Isn't it better to have this a instance method on `XLColor` itself?

ClosedXML/Excel/Style/Colors/XLColor_Static.cs
```diff
@@ -169,9 +169,20 @@ public static XLColor FromTheme(XLThemeColor themeColor, Double themeTint)
...
+        {
+            return color == null
+                || !color.HasValue
+                || IsTransparent(color.Key);
+        }
+
+        internal static bool IsTransparent(in XLColorKey colorKey)
```
igitur: Same question for this one too.

ClosedXML/Excel/Style/Colors/XLColor_Static.cs
Pankraty: This would make the usage more complicated. Compare XLColor.IsNullOrTransparent(color) vs (color == null || color.IsTransparent()).

But maybe you're right.

igitur: No, leave it like it is.

igitur:
Commit f79a9080a691ab2bfacf46d1b9f648cad0cc0f53 broke the `LoadAndSaveTransparentBackgroundFill` test. Well, it made the test pass, but the reference file is now incorrect. If you open it, the cell of concern displays with a black background.

Issue #866, opened 2018-05-22
Unprotect not functioning? Example needed.


michimkc:
## Read and complete the full issue template

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

If you are logging a possible bug or feature request, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.92.1.0

**What is the current behavior?**

Unprotect on IXLWorksheet reports "Invalid Password" even if string passed to Unprotect() is the correct password, or if the excel sheet is protected without any password.

**What is the expected behavior or new feature?**

Unprotect should unprotect the sheet, given the correct password.

**Did this work in previous versions of our tool?  Which versions?**

Have not tested on previous versions.

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

Without a code sample, it is unlikely that your issue will get attention. Don't be lazy. Do the effort and assist the developers to reproduce your problem. Code samples should be [minimal complete and verifiable](https://stackoverflow.com/help/mcve). Sample spreadsheets should be attached whenever applicable. Remove sensitive information.

**Code to reproduce problem:**
```c#
public void Main()
{
    // Code standards:
    // - Fully runnable. I should be able to copy and paste this code into a 
    //   console application and run it without having to edit it much.
    // - Declare all your variables (this follows from the previous point)
    // - The code should be a minimal code sample to illustrate issue. The code 
    //   samples on the wiki are good examples of the terseness that I want. Don't
    //   post your full application.

        
            var wb = new XLWorkbook("protectedsheets.xlsx");

            var ws = wb.Worksheet(1);

            // sheet 1 - protected sheet with no password
            try
            {
                ws.Unprotect();
            }
            catch (Exception e)
            {
                Console.WriteLine("Could not unprotect sheet. Error was: " + e.Message);
            }

            // sheet 2 - protected sheet with password "abc"
            ws = wb.Worksheet(2);
            
            try
            {
                ws.Unprotect("abc");
            }
            catch (Exception e)
            {
                Console.WriteLine("Could not unprotect sheet. Error was: " + e.Message);
            }
}
```
- [ ] I attached a sample spreadsheet.  (You can drag files on to this issue)

[protectedsheets.xlsx](https://github.com/ClosedXML/ClosedXML/files/2024969/protectedsheets.xlsx)

[...]

PR #876, opened 2018-05-28
Correctly unprotect worksheet without password

igitur:
This PR partially addresses #866 . It fixes a bug with unprotected worksheets that are protected without a password.

Worksheets protected with the SHA512 algorithm are not supported yet.


ClosedXML_Tests/Excel/Worksheets/XLSheetProtectionTests.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        [Test]
+        public void TestUnprotectWorksheetWithNoPassword()
+        {
+            using (var stream = TestHelper.GetStreamFromResource(TestHelper.GetResourcePath(@"Misc\SHA512PasswordProtection.xlsx")))
+            using (var wb = new XLWorkbook(stream))
+            {
+                wb.Worksheet("Sheet1").Unprotect();
+
+                // Protected with SHA-512 password - not yet supported.
```
Pankraty: I think this test should be split in two: the first unprotecting `Sheet1` and the second unprotecting `Sheet2`. Until SHA-512 is implemented the second one must be muted, but both must have correct assertions from the beginning (e.g. `Assert.IsFalse(ws.Protection.Protected);`).

ClosedXML_Tests/Excel/Worksheets/XLSheetProtectionTests.cs
```diff
@@ -0,0 +1,23 @@
...
+                Assert.Throws<ArgumentException>(() => wb.Worksheet("Sheet2").Unprotect());
```
Pankraty: `NotSupportedException` or `NotImplementedException` would be more informative.

igitur: That would be a breaking change. I'd rather spend any effort on actually implementing this algorithm.

ClosedXML_Tests/Excel/Worksheets/XLSheetProtectionTests.cs
igitur: Ok, the split them in to, but note that the SHA-512 tests is just a temporary measure until we actually implement the algorithm.

PR #879, opened 2018-05-28
WIP First proof of concept for adding context to functions

igitur:
Work in progress.

Adds `CalculationContext` to functions to enable self-referencing functions e.g. `ROW()`

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Expression.cs
+++ b/ClosedXML/Excel/CalcEngine/Expression.cs
@@ -10,7 +10,18 @@ namespace ClosedXML.Excel.CalcEngine
 {
     internal abstract class ExpressionBase
     {
+        private CalculationContext _context;
+        protected ExpressionBase(in CalculationContext ctx)
+        {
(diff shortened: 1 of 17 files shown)


ClosedXML/Excel/CalcEngine/Expression.cs
```diff
@@ -10,7 +10,18 @@ namespace ClosedXML.Excel.CalcEngine
...
     {
+        private CalculationContext _context;
+        protected ExpressionBase(in CalculationContext ctx)
+        {
+            _context = ctx;
+        }
+
         public abstract string LastParseItem { get; }
+
+        public ref CalculationContext CurrentContext
```
Pankraty: I would like to move toward stateless parser and calculation engine, though I am not sure if it is acheivable. 
This would, theoretically, allow doing calculations in parallel eventually.

igitur:
@Pankraty Ok, submit your PoC when you're ready, then we can discuss further. No rush.

PR #887, opened 2018-05-31
Correctly determine used range when there are merged ranges on a worksheet (fix #881)

### LINKED ISSUE #881: Worksheet.RangeUsed( true ) ignores a cell with value

- [x] Bug
- [ ] Feature

**Version of ClosedXML**
e.g. 0.93-beta4 (tested also in the `develop` branch)

Calling `Worksheet.RangeUsed( true )` returns a range which doesn't include a non-empty cell.

`Worksheet.RangeUsed` (regardless of the value of the `includeFormats` parameter) should return a range which includes all cells which have a value.

Don't know.

## Reproducibility
```c#
public void Main()
{
	using( var wb = new XLWorkbook( "Test.xlsx" ) )
	{
		Console.WriteLine( ws.RangeUsed( false ).ToString() );
		Console.WriteLine( ws.RangeUsed( true ).ToString() );
	}
}
```

This results in
```
Test!A1:B2
Test!B2:D2
```

The second call ignores cell A1.

[Test.xlsx](https://github.com/ClosedXML/ClosedXML/files/2053878/Test.xlsx)

Pankraty:
This PR fixes #881.

The code for getting the first and the last used cells of the range had following problems:

1. It analyzed merged ranges only when `includeFormat` flag was true. I am convinced that merged ranges had to be considered even if `includeFormat` is false because it's not just the matter of formatting but it specifies how far the data extends.
2. When merged ranges were analyzed all other cells were ignored. Say, for the `FirstUsedCell` the first cell of merged ranges was returned even if there exist other cells with lower coordinates (see the original issue)
3. Besides, when merged ranges were analyzed the predicate passed into method was ignored. It did not affect `RangeUsed` method as it does not use predicate but produced an incorrect result if `FirstCellUsed` and `LastCellUsed` methods were called directly.

There still is the problem with applying a predicate to rows and columns (it is ignored there too) but that is another issue (#889), which is harder to fix.

### DIFF
--- a/ClosedXML/Excel/Ranges/XLRangeBase.cs
+++ b/ClosedXML/Excel/Ranges/XLRangeBase.cs
@@ -635,16 +635,19 @@ namespace ClosedXML.Excel
                         if (!cell.IsEmpty(true)) return cell;
                     }
                 }
+            }
 
-                var intersectedRanges = Worksheet.MergedRanges.GetIntersectedRanges(RangeAddress)
-                    .Where(r => predicate == null || predicate(r.FirstCell())).ToList();
-                if (intersectedRanges.Any())
-                {
-                    Int32 minRo = intersectedRanges.Min(r => r.RangeAddress.FirstAddress.RowNumber);
-                    Int32 minCo = intersectedRanges.Min(r => r.RangeAddress.FirstAddress.ColumnNumber);
+            var intersectedRanges = Worksheet.MergedRanges.GetIntersectedRanges(RangeAddress)
+                .Where(r => predicate == null || predicate(r.FirstCell())).ToList();
+            if (intersectedRanges.Any())
+            {
+                Int32 minRo = intersectedRanges.Min(r => r.RangeAddress.FirstAddress.RowNumber);
+                Int32 minCo = intersectedRanges.Min(r => r.RangeAddress.FirstAddress.ColumnNumber);
 
-                    return Worksheet.Cell(minRo, minCo);
-                }
+                if (0 < sp.Row && sp.Row < minRo) minRo = sp.Row;
+                if (0 < sp.Column && sp.Column < minCo) minCo = sp.Column;
+
+                return Worksheet.Cell(minRo, minCo);
             }
 
             if (sp.Row > 0)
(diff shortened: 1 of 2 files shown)


ClosedXML/Excel/Ranges/XLRangeBase.cs
```diff
@@ -635,15 +635,19 @@ public XLCell FirstCellUsed(Boolean includeFormats, Func<IXLCell, Boolean> predi
...
                 }
+            }
 
-                var intersectedRanges = Worksheet.MergedRanges.GetIntersectedRanges(RangeAddress).ToList();
-                if (intersectedRanges.Any())
-                {
-                    Int32 minRo = intersectedRanges.Min(r => r.RangeAddress.FirstAddress.RowNumber);
-                    Int32 minCo = intersectedRanges.Min(r => r.RangeAddress.FirstAddress.ColumnNumber);
+            var intersectedRanges = Worksheet.MergedRanges.GetIntersectedRanges(RangeAddress)
+                .Where(r => predicate == null || predicate(r.FirstCell())).ToList();
```
igitur: I think we should add overloads to `GetIntersectedRanges` to accept a predicate which is passed through. Also the nullable predicate can be invoked with a shorter syntax:

```c#
            var intersectedRanges = Worksheet
                .MergedRanges
                .GetIntersectedRanges(RangeAddress, predicate?.Invoke(r.FirstCell()) ?? true);
```
And doesn't look like the `ToList()` is necessary.

ClosedXML/Excel/Ranges/XLRangeBase.cs
```diff
@@ -725,15 +729,19 @@ public XLCell LastCellUsed(Boolean includeFormats, Func<IXLCell, Boolean> predic
...
                     }
                 }
+            }
 
-                var intersectedRanges = Worksheet.MergedRanges.GetIntersectedRanges(RangeAddress).ToList();
-                if (intersectedRanges.Any())
-                {
-                    Int32 minRo = intersectedRanges.Max(r => r.RangeAddress.LastAddress.RowNumber);
-                    Int32 minCo = intersectedRanges.Max(r => r.RangeAddress.LastAddress.ColumnNumber);
+            var intersectedRanges = Worksheet.MergedRanges.GetIntersectedRanges(RangeAddress)
```
igitur: Same here.

ClosedXML/Excel/Ranges/XLRangeBase.cs
igitur: Oh the other hand, I see `Quandrant.GetIntersectedRanges` uses `yield return`, so passing through the predicate probably doesn't add much value? Or does it?

ClosedXML/Excel/Ranges/XLRangeBase.cs
```diff
@@ -635,15 +635,19 @@ public XLCell FirstCellUsed(Boolean includeFormats, Func<IXLCell, Boolean> predi
...
+                .Where(r => predicate == null || predicate(r.FirstCell())).ToList();
+            if (intersectedRanges.Any())
+            {
+                Int32 minRo = intersectedRanges.Min(r => r.RangeAddress.FirstAddress.RowNumber);
+                Int32 minCo = intersectedRanges.Min(r => r.RangeAddress.FirstAddress.ColumnNumber);
 
-                    return Worksheet.Cell(minRo, minCo);
-                }
+                if (0 < sp.Row && sp.Row < minRo) minRo = sp.Row;
```
igitur: There's a nice extension method to simplify this: `if (sp.Row.Between(1, minRo - 1)) ...`

ClosedXML/Excel/Ranges/XLRangeBase.cs
Pankraty: Without `ToList()`, the collection will be enumerated twice (`.Any()` will break on the first item, but then there are two `Min()` calls) which is undesirable.

> `Quandrant.GetIntersectedRanges` uses `yield return`, so passing through the predicate probably doesn't add much value

I agree with this - the effect will be the same.

Pankraty: >  Also the nullable predicate can be invoked with a shorter syntax:...

Whole 5 characters shorter :)  OK, will use it.

ClosedXML/Excel/Ranges/XLRangeBase.cs
Pankraty: Cool, thanks!

ClosedXML/Excel/Ranges/XLRangeBase.cs
igitur: It's not about the number of characters. It's (for me at least) just easier to understand.

ClosedXML/Excel/Ranges/XLRangeBase.cs
igitur: PS: In some places Visual Studio actually recommends to use the `predicate?.Invoke()` syntax.

PR #888, opened 2018-05-31
Use current locale in the test instead of invariant culture

Pankraty:
`SetStringCellValues` test fails on my machine. It converts a typed value to a string and then applies string value to a cell expecting that cell format will be deduced from the value. But this works with string formatted in the current locale if I am not mistaken. So I replaced `.ToInvariantString()` to `.ToString()` and test became green.


igitur:
Has this been failing on your machine all along?

igitur:
Oh, it's my recent PR that introduced this?

Pankraty:
> it's my recent PR that introduced this?

Exactly.

PR #891, opened 2018-05-31
Bump version to 0.94

igitur:


Pankraty:
Maybe change .csproj as well so local builds were up-to-date?

igitur:
> Maybe change .csproj as well so local builds were up-to-date?

Oh yeah! Odd, I did a global search for `93`, but the entries in `csproj` weren't hit.

PR #904, opened 2018-06-07
Implement the full range of Item subclasses for pivot tables 

igitur:
Implement the full range of Item subclasses for pivot tables 
https://msdn.microsoft.com/en-us/library/documentformat.openxml.spreadsheet.shareditems.aspx


igitur:
Also slight refactoring to use pattern matching and deduplicate code.

PR #909, opened 2018-06-09
Reapply autofilter

### LINKED ISSUE #900: Reapply autofilter
When loading a workbook from a file, the row visibility in a table might be out of sync with what the filter implies. We need a way to reapply the filter so that row visibility is recalculated.

igitur:
Fixes #900 

- First does some Codemaid cleanup.
- Refactors autofilters which removes `IXLBaseAutoFilter` (doesn't seem necessary).
- Adds `IXLAutoFilter.HiddenRows` and `IXLAutoFilter.Reapply()`


igitur:
@Pankraty When you review this, take it one commit at a time. The diffs will make more sense.

igitur:
Breaking change: `IXLBaseAutoFilter` removed and replaced with the existing `IXLAutoFilter`.

PR #910, opened 2018-06-12
Consolidate XLCell.InsertData and XLCell.InsertTable

igitur:
Before continuing work on #908 , let's first clean up and consolidate `XLCell.InsertData` and `XLCell.InsertTable`, which share a lot of logic.

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -655,79 +655,214 @@ namespace ClosedXML.Excel
             return InsertTable(data, tableName, true);
         }
 
-        public IXLTable InsertTable<T>(IEnumerable<T> data, string tableName, bool createTable)
+        public IXLTable InsertTable<T>(IEnumerable<T> data, String tableName, Boolean createTable)
+        {
+            return InsertTable(data, tableName, createTable, addHeadings: true, transpose: false);
+        }
+
+        public IXLTable InsertTable<T>(IEnumerable<T> data, String tableName, Boolean createTable, Boolean addHeadings, Boolean transpose)
         {
             if (createTable && this.Worksheet.Tables.Any(t => t.Contains(this)))
                 throw new InvalidOperationException(String.Format("This cell '{0}' is already part of a table.", this.Address.ToString()));
 
-            if (data != null && !(data is String))
+            var range = InsertDataInternal(data, addHeadings, transpose);
+
+            if (createTable)
+                // Create a table and save it in the file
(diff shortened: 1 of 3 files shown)


ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -655,79 +655,210 @@ public IXLTable InsertTable<T>(IEnumerable<T> data, string tableName)
...
+        public IXLTable InsertTable(DataTable data, String tableName, Boolean createTable)
+        {
+            if (data == null || data.Columns.Count == 0)
+                return null;
+
+            if (createTable && this.Worksheet.Tables.Any(t => t.Contains(this)))
+                throw new InvalidOperationException(String.Format("This cell '{0}' is already part of a table.", this.Address.ToString()));
+
+            if (data.Rows.Cast<DataRow>().Any()) return InsertTable(data.Rows.Cast<DataRow>(), tableName, createTable);
+            var ro = _rowNumber;
```
Pankraty: This variable is not used.

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -655,79 +655,210 @@ public IXLTable InsertTable<T>(IEnumerable<T> data, string tableName)
...
+
+        public IXLTable InsertTable(DataTable data, String tableName, Boolean createTable)
+        {
+            if (data == null || data.Columns.Count == 0)
+                return null;
+
+            if (createTable && this.Worksheet.Tables.Any(t => t.Contains(this)))
+                throw new InvalidOperationException(String.Format("This cell '{0}' is already part of a table.", this.Address.ToString()));
+
+            if (data.Rows.Cast<DataRow>().Any()) return InsertTable(data.Rows.Cast<DataRow>(), tableName, createTable);
```
Pankraty: `data.Rows.Cast<DataRow>()` will throw an exception in case of invalid cast, if I am not mistaken. `data.Rows.OfType<DataRow>()` would fit better.

ClosedXML/Excel/Cells/XLCell.cs
igitur: Which one? `ro`?

ClosedXML/Excel/Cells/XLCell.cs
igitur: `data` is of type `DataTable` and `DataTable.Rows` should contain only items of type `DataRow`. I'd prefer an exception if, for some reason, this wasn't the case, rather than "losing" any items.

Pankraty: You're right, I missed this is a `DataTable`

PR #924, opened 2018-06-26
 Fix determining the column letter by index 

### LINKED ISSUE #923: Getting Exception of System.ArgumentOutOfRangeException type
- [x] Bug
- [ ] Feature

**Version of ClosedXML**
0.93.0

**What is the current behaviour?**
System.ArgumentOutOfRangeException: 'Specified argument was out of the range of valid values.
Parameter name: AZA is not recognized as a column letter'

**What is the expected behaviour or new feature?**
It is expected to open this file just like other files. 

No

## Reproducibility

```
var path = "gu.xlsx";

```

System.ArgumentOutOfRangeException: 'Specified argument was out of the range of valid values.
Parameter name: AZA is not recognized as a column letter'

That this the error I get when I try to open an excel file (attached here)

[gu.xlsx](https://github.com/ClosedXML/ClosedXML/files/2133659/gu.xlsx)

Pankraty:
Fixes #923.

@igitur thank you for tests. I should do this from the start.

### DIFF
--- a/ClosedXML/XLHelper.cs
+++ b/ClosedXML/XLHelper.cs
@@ -71,7 +71,7 @@ namespace ClosedXML.Excel
                 else if (i < 26 * 27)
                     letter = letters[i / 26 - 1] + letters[i % 26];
                 else
-                    letter = letters[i / 26 / 26 - 1] + letters[(i / 26 - 1) % 26] + letters[i % 26];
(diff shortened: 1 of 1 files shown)


igitur:
I want to release this as a patch for `0.93`. Shouldn't we base this against `release-candidate`?

igitur:
Even my branch was against `develop`.

PR #928, opened 2018-06-27
Exclude full span addresses from range change notifications

igitur:
Full span ranges (e.g. full rows, full columns and entire worksheets) should not be affected by range deletions and insertions.

Basing this on `release-candidate`


igitur:
Closing this as this is a problem in `0.92` too.

PR #930, opened 2018-06-27
Store picture markers against a range

### LINKED ISSUE #927: Pictures do not change location on row insert
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.94

When inserting a new Row, images in the document to not respect their placement setting. Image remains in the same location despite rows being inserted.

Images set to XLPicturePlacement.Move or XLPicturePlacement.MoveAndSize should shift their position down when row is inserted above.

Presumably, No

## Reproducibility

```c#  
    namespace ClosedXML_Examples
    {
        public class CreateFiles
        {
            public static void CreateAllFiles()
            {
                var path = Program.BaseCreatedDirectory;

                new ImageShift().Create(Path.Combine(path, "ImageShift.xlsx"));
```

---

```c# 
    public class ImageShift : IXLExample
    {
        public void Create(string filePath)
        {
            var wb = new XLWorkbook();
            IXLWorksheet ws;

            using (Stream fs = Assembly.GetExecutingAssembly().GetManifestResourceStream("ClosedXML_Examples.Resources.ImageHandling.jpg"))
            {

                ws = wb.Worksheets.Add("ImageShift");
                ws.AddPicture(fs, XLPictureFormat.Jpeg, "JpegImage")
                    .MoveTo(ws.Cell(5, 2).Address) // Insert to Row 5
                    .WithPlacement(XLPicturePlacement.Move);

                ws.Row(2).InsertRowsBelow(20); // At Row 2, Insert 20 rows. Image should now be on Row 25.

                wb.SaveAs(filePath);
            }
        }
    }
```

[ImageShift.xlsx](https://github.com/ClosedXML/ClosedXML/files/2141968/ImageShift.xlsx)

igitur:
Fixes #927 

Breaking change: picture position parameters (e.g. for `MoveTo`) are now based on `IXLCell` instead of `IXLAddress`.


Pankraty:
Something wrong with the test project. Did you forget to commit some changes?

igitur:
Yeah, I messed up

ClosedXML/Excel/Drawings/IXLMarker.cs
```diff
@@ -1,10 +0,0 @@
-using System.Drawing;
-
-namespace ClosedXML.Excel.Drawings
-{
-    internal interface IXLMarker
```
Pankraty: Out of curiosity, why did you decide to remove the interface?

ClosedXML/Excel/Drawings/XLMarker.cs
```diff
@@ -1,23 +1,38 @@
...
 
 namespace ClosedXML.Excel.Drawings
 {
     [DebuggerDisplay("{Address} {Offset}")]
-    internal class XLMarker : IXLMarker
+    internal class XLMarker
     {
-        internal XLMarker(IXLAddress address)
-            : this(address, new Point(0, 0))
+        // Using a range to store the lcoation so that it gets added to the range repository
```
Pankraty: "lcoation"-  a typo

ClosedXML/Excel/Drawings/XLMarker.cs
```diff
@@ -1,23 +1,38 @@
...
 
-        internal XLMarker(IXLAddress address, Point offset)
+        internal XLMarker(IXLCell cell, Point offset)
+            : this(cell.AsRange(), offset)
+        { }
+
+        internal XLMarker(IXLRange rangeCell, Point offset)
         {
-            this.Address = address;
+            this.rangeCell = rangeCell;
```
Pankraty: I'd suggest adding the validation that `rangeCell` is actually one cell high and one cell wide and throw the `ArgumentException` otherwise. Thus we'll protect ourselves from misusage. Besides, markers constructed for ranges `A1:A1` and `A1:B2` functionally would be equal, but their `rangeCell` would differ

ClosedXML/Excel/Drawings/IXLMarker.cs
igitur: Neither `IXLMarker` nor `XLMarker` is public. Both are internal, so I don't really see the point of an interface. Or am I missing something?

ClosedXML/Excel/Drawings/XLMarker.cs
igitur: True, I'm going to change the constructor modifier too. Should be `private`.

ClosedXML/Excel/Drawings/IXLMarker.cs
Pankraty: I don't see any problems with this either. We can add the interface back any time if we need it.

PR #937, opened 2018-07-01
Implement OpenXML's CellFormat.QuotePrefix

### LINKED ISSUE #935: Set Cell Format To Text

- [] Bug
- [x] Feature

I searched what is currently being developed and I see nothing in this regard.

**Version of ClosedXML**

0.92.1

Can't set the cell format to text when the value is a number, even if the DataTable has that column set to string nor by setting the cell data type. So for example, if I have a string column that has numbers with leading 0s, then when the excel file is outputted, it displays properly. The problem is when a user edits this field, the leading 0s are not retained. So say I have 00023, the moment the user edits this and just changes it to 00022, then excel changes it to 22. This of course does NOT happen if I manually set the cell's format to text. It always retains the leading 0s no matter how many times a user edits the value. In the sample code below, I have some lines commented out. I tried to set the number format. When this occurs, the leading 0s are displayed when the value is edited, but the value truncates the leading 0s. So displaying the leading 0s is not the problem, it's when editing the value that the 0s disappear or are NOT part of the value anymore.

To be able to set cell format to text, regardless of data type so in cases of numbers with leading 0s, the leading 0s will retain when the cell is changed. It could be some property to set the cell format. I'm not sure if perhaps that's possible already and I couldn't find it in the samples. It would be nice if there was documentation to search through, but even so I don't believe it's anywhere (at least when I searched each sample I couldn't find this at all).

I'm not sure, but I believe no

## Reproducibility

```c#
public void Main()
{
    System.Data.DataTable dt = new System.Data.DataTable("Test sheet");
    dt.Columns.Add("ID", typeof(int));
    dt.Columns.Add("AlphaNumericID", typeof(string));
	dt.Columns.Add("MoneyValue", typeof(decimal));
	dt.Columns.Add("Name", typeof(string));
	dt.Columns.Add("IntValue", typeof(int));
	dt.Columns.Add("AlphaNumericID2", typeof(string));

			new System.Collections.Generic.List<int> { 23, 1, 2, 5, 0, 5, 8 }.ForEach(x =>
			{
				System.Data.DataRow r = dt.NewRow();
				int i = -1;
				r[++i] = x;
				r[++i] = x.ToString("00000000#");
				r[++i] = x * 23;
				r[++i] = "Name " + x;
				r[++i] = x * 10;
				r[++i] = x.ToString("00000000#0000");
				dt.Rows.Add(r);
			});
			
			ClosedXML.Excel.XLWorkbook xlsx = new ClosedXML.Excel.XLWorkbook(ClosedXML.Excel.XLEventTracking.Disabled);
			
            xlsx.Worksheets.Add(dt);
            
            ClosedXML.Excel.IXLWorksheet xlWorksheet = xlsx.Worksheet(1);
            
            foreach(int c in new[] {2, 6})
            {
				xlWorksheet.Column(c).CellsUsed().SetDataType(XLDataType.Text);
				for (int r = 2; r < 9; ++r)
				{
					ClosedXML.Excel.IXLCell cell = xlWorksheet.Cell(r, c);
					string val = cell.Value + "";

					//cell.Style.NumberFormat.Format = "".PadRight(val.Length - 1, '0') + "#";
					//cell.SetValue<string>(val);
					cell.Value = "'" + val;
					
					cell.DataType = XLDataType.Text;
                                        //it would be nice if we had something like this since the above lines don't do much, except help in the presentation not when editing a cell
                                        //cell.CellFormat = CellFormat.Text;
				}//for row
            }//foreach col
			
			string path = "test.xlsx";
			xlsx.SaveAs(path);
}
```
[201806300120296235.xlsx](https://github.com/ClosedXML/ClosedXML/files/2151304/201806300120296235.xlsx)

igitur:
Fixes #935

I'm just wondering whether `cell.style.IncludeQuotePrefix` should be `true` by default when using this way of setting a value: `cell.value = "'500"`. This would be a breaking change, but would align with what Excel does.


igitur:
Note: add example to wiki

PR #945, opened 2018-07-05
Allow Guid values to be inserted

### LINKED ISSUE #918: Error Exporting datatable to Excel (.xlsx) 'Unable to set cell value to ' + uniqueidentifier

- [x] Bug
- [ ] Feature

Less of a bug, more of a workaround, but I the person who provided the answer thought it might be valuable to communicate it to the developers.

**Version of ClosedXML**

0.93.0-rc3 (through Visual Studio NuGet)

When exporting a datatable gathered from a SQL table, the uniqueidentifier field (at column index 0) cannot be exported using the standard export to Excel feature I get the error: "Unable to set cell value to [uniqueidentifier]". I have attempted to change the index of the column in the table but this has no effect.

I expected the uniqueidentifer to be treated as a text/string field.

Unknown - Haven't attempted on previous versions.

## Reproducibility

Converted from original VB.NET
````c# 
private void ExportToxlsxUsingClosedXML()
{
    string appPath = Directory.GetCurrentDirectory();
    string filename = appPath + "\mytest_" + Strings.Format(DateTime.Now, "yyyyMMddHHmmss") + ".xlsx";

    string SQL_Query = "select * FROM [mydb].[dbo].[mytable] where myDate >= '2017-01-01' ";

    DataTable myTable = CLS_SQL.SQL_Retrieve(SQL_Query);

    wb.Worksheets.Add(myTable, "MyData");
    wb.SaveAs(filename);
}
```
I managed to fix the issue by casting my uniqueidentifier as VARCHAR(36) in the SQL statement, thanks to the help of AndrewMorton in the below StackOverflow thread:
https://stackoverflow.com/questions/50968359/closedxml-export-to-xlsx-error-unable-to-set-cell-value-to-uniqueidentifie

igitur:
Fixes #918

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -223,7 +223,7 @@ namespace ClosedXML.Excel
 
             // For SetValue<T> we set the cell value directly to the parameter
             // as opposed to the other SetValue(object value) where we parse the string and try to decude the value
-            if (value is String || value is char)
+            if (value is String || value is char || value is Guid)
             {
                 parsedValue = value.ToInvariantString();
                 _dataType = XLDataType.Text;
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -223,21 +223,9 @@ namespace ClosedXML.Excel
 
             // For SetValue<T> we set the cell value directly to the parameter
             // as opposed to the other SetValue(object value) where we parse the string and try to decude the value
-            if (value is String || value is char || value is Guid)
-            {
-                parsedValue = value.ToInvariantString();
-                _dataType = XLDataType.Text;
-                if (parsedValue.Contains(Environment.NewLine) && !style.Alignment.WrapText)
-                    Style.Alignment.WrapText = true;
-
-                parsed = true;
-            }
-            else
-            {
-                var tuple = SetKnownTypedValue(value, style);
-                parsedValue = tuple.Item1;
-                parsed = tuple.Item2;
-            }
+            var tuple = SetKnownTypedValue(value, style, acceptString: true);
+            parsedValue = tuple.Item1;
+            parsed = tuple.Item2;
 
             // If parsing was unsuccessful, we throw an ArgumentException
             // because we are using SetValue<T> (typed).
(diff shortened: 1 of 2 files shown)


ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -223,7 +223,7 @@ internal IXLCell SetValue<T>(T value, bool setTableHeader)
 
             // For SetValue<T> we set the cell value directly to the parameter
             // as opposed to the other SetValue(object value) where we parse the string and try to decude the value
-            if (value is String || value is char)
+            if (value is String || value is char || value is Guid)
```
Pankraty: Now we have one branch in the code to process strings, chars, and GUIDs, and another branch to process everything else. There we have separate conditions for dates, timespans, numbers, booleans, and again for everything else. The later case converts the value using `.ToString()` and - unlike this fragment - does not care about line wrapping.

For me, it looks unclear why the particular type's processing should be added to the `SetValue` method and not to the `SetKnownTypedValue`. 
I suggest putting this logic into `SetKnownTypedValue`: add another `else if` condition for strings and chars and maybe GUIDs as well.

igitur: String values are treated differently depending on whether you use `.SetValue(s)` or `.Value = s` .  The latter one will try to parse the value into e.g. a `DateTime`. The logic is therefore deliberately excluded from the other known types. I can shift the logic, but I will have to add a flag to `SetKnownTypeValue` to indicate whether to accept strings as they are or try to parse them into a value. Example look at this test: `SetStringCellValues`

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -253,11 +241,20 @@ internal IXLCell SetValue<T>(T value, bool setTableHeader)
         }
 
         // TODO: Replace with (string, bool) ValueTuple later
-        private Tuple<string, bool> SetKnownTypedValue<T>(T value, XLStyleValue style)
+        private Tuple<string, bool> SetKnownTypedValue<T>(T value, XLStyleValue style, Boolean acceptString)
         {
             string parsedValue;
             bool parsed;
-            if (value is DateTime d && d >= BaseDate)
+            if (value is String && acceptString || value is char || value is Guid)
```
igitur: @Pankraty Logic now moved into `SetKnownTypeValue`

sopelt:
Hi, do you happen to know when this change will ship? Having to stick with pre-0.93 to avoid compatibility issues feels somewhat icky ;)
Thank you for your work!

igitur:
@sopelt I don't commit to timelines. You're welcome to build it from source. Me releasing it right now on Nuget doesn't make it any more official than it would be if you compiled it yourself.

sopelt:
Thank you for the swift reply @igitur  ... I was just wondering if I missed something in terms of a roadmap ;)
In cases where ClosedXML is used as a dependency of another NuGet package it is nice to have an official version to depend on.
Thanks again.

PR #947, opened 2018-07-06
Fix pivot table integrity when saving multiple times.

igitur:
Fixes problem with saving file multiple times. Pivot table relId has to be set after first save.

### DIFF
--- a/ClosedXML/Excel/XLWorkbook_Save.cs
+++ b/ClosedXML/Excel/XLWorkbook_Save.cs
@@ -2012,14 +2012,19 @@ namespace ClosedXML.Excel
                 GeneratePivotTableCacheDefinitionPartContent(pivotTableCacheDefinitionPart, pt, context);
 
                 PivotTablePart pivotTablePart;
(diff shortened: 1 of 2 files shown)


ClosedXML/Excel/XLWorkbook_Save.cs
```diff
@@ -2012,14 +2012,20 @@ private static void GenerateTableDefinitionPartContent(TableDefinitionPart table
                 GeneratePivotTableCacheDefinitionPartContent(pivotTableCacheDefinitionPart, pt, context);
 
                 PivotTablePart pivotTablePart;
+                bool createdNewPart = false;
```
Pankraty: Maybe do this way?
```
var createNewPart = String.IsNullOrWhiteSpace(pt.RelId);
if (createNewPart)
{
    ...
```
A bit more laconic.

igitur: Had to google "laconic". :-P

Issue #940, opened 2018-07-03
Regressions from v0.92.1


igitur:
Placeholder issue.

- ~#923~
- Regression in ~#862~
- ~#938~

PR #949, opened 2018-07-12
Aggregate color conversion in one place

Pankraty:
Currently, we have 6 nearly identical `switch`-es to convert ClosedXML colors to OpenXML format. For sparklines, we'd have to add 8 more as each color to be saved must be converted to its own class (`FirstMarkerColor`, `LastMarkerColor`, etc.). That's obviously insane.

This PR replaces all those `switch`-es with a single method that may be utilized by sparklines too.

### DIFF
--- a/ClosedXML/Excel/XLWorkbook_Load.cs
+++ b/ClosedXML/Excel/XLWorkbook_Load.cs
@@ -1706,7 +1706,7 @@ namespace ClosedXML.Excel
                 if (source.Style != null)
                     setBorder(source.Style.Value.ToClosedXml());
                 if (source.Color != null)
-                    setColor(GetColor(source.Color));
+                    setColor(source.Color.ToClosedXMLColor(_colorList));
             }
         }
 
@@ -1731,22 +1731,22 @@ namespace ClosedXML.Excel
                     if (differentialFillFormat)
                     {
                         if (openXMLFill.PatternFill.BackgroundColor != null)
-                            closedXMLFill.BackgroundColor = GetColor(openXMLFill.PatternFill.BackgroundColor);
+                            closedXMLFill.BackgroundColor = openXMLFill.PatternFill.BackgroundColor.ToClosedXMLColor(_colorList);
                     }
                     else
                     {
                         // yes, source is foreground!
                         if (openXMLFill.PatternFill.ForegroundColor != null)
-                            closedXMLFill.BackgroundColor = GetColor(openXMLFill.PatternFill.ForegroundColor);
+                            closedXMLFill.BackgroundColor = openXMLFill.PatternFill.ForegroundColor.ToClosedXMLColor(_colorList);
                     }
                     break;
 
                 default:
                     if (openXMLFill.PatternFill.ForegroundColor != null)
-                        closedXMLFill.PatternColor = GetColor(openXMLFill.PatternFill.ForegroundColor);
+                        closedXMLFill.PatternColor = openXMLFill.PatternFill.ForegroundColor.ToClosedXMLColor(_colorList);
 
                     if (openXMLFill.PatternFill.BackgroundColor != null)
-                        closedXMLFill.BackgroundColor = GetColor(openXMLFill.PatternFill.BackgroundColor);
+                        closedXMLFill.BackgroundColor = openXMLFill.PatternFill.BackgroundColor.ToClosedXMLColor(_colorList);
                     break;
             }
         }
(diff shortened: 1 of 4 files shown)


igitur:
Good idea, but I think we have to fix the regressions ( #940 ) first before taking this on, as some of the regressions are fill color related.

Pankraty:
Good point, but it turned out, these two issues are completely unrelated: the regression happened in loading, and this PR addresses saving only.

vbjay:
Next question. Is there a case of some color difference between different
objects?  Like sparkines not supporting transparent or only rgb and not
argb for one ibject type vs another. If so, then consolidating color
handling may not be ideal.

Pankraty:
Once they all inherit ColorType class we can safely expect the same behavior. Otherwise it would be a violation of Liskov principle.

Pankraty:
Not mentioning DRY principle which was violated too

vbjay:
I'm talking about how Excel handles things, not closedxml. If they did
something odd, we need to handle the odd.

vbjay:
I do agree. I'm just pointing out if there is a difference somewhere,
consolidating would cause issues for the odd ones.

ClosedXML/Utils/OpenXmlHelper.cs
```diff
@@ -13,5 +18,42 @@ public static bool GetBooleanValueAsBool(BooleanValue value, bool defaultValue)
...
+        /// </summary>
+        /// <typeparam name="T">OpenXML color type to convert the argument to.</typeparam>
+        /// <param name="xlColor">Color in ClosedXML format.</param>
+        /// <param name="isDifferential">Flag specifiying that the color should be saved in
+        /// differential format (affects the transparent color processing).</param>
+        /// <returns>The original color in OpenXXL format.</returns>
+        public static T ToOpenXmlColor<T>(XLColor xlColor, bool isDifferential = false)
+            where T : DocumentFormat.OpenXml.Spreadsheet.ColorType, new()
+        {
+            var openXmlColor = new T();
```
igitur: `new T()` is quite slow and there's another pattern that is preferred. See https://blogs.msdn.microsoft.com/seteplia/2017/02/01/dissecting-the-new-constraint-in-c-a-perfect-example-of-a-leaky-abstraction/

Pankraty: Interesting, thanks. Never tested the performance of this approach.

Pankraty:
Besides, I was wrong in my assumptions about re-using the same code for sparklines. I missed that they use color classes inherited from `ColorType` in _different_ namespace. Those two `ColorType`s have the same properties and functionally are the same but they don't share any common interface so some code duplication is inevitable, I am afraid.

Pankraty:
I've re-implemented my solution in such a way that would allow reusing the same code for `ColorType` from both namespaces (`DocumentFormat.OpenXml.Spreadsheet` and `DocumentFormat.OpenXml.Office2010.Excel`). And I discarded `new T()` too.

ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -2365,8 +2365,7 @@ private void LoadExtensions(WorksheetExtensionList extensions, XLWorksheet ws)
                 if (xlConditionalFormat != null)
                 {
                     var negativeFillColor = conditionalFormattingRule.Descendants<DocumentFormat.OpenXml.Office2010.Excel.NegativeFillColor>().SingleOrDefault();
-                    var color = new DocumentFormat.OpenXml.Spreadsheet.Color { Rgb = negativeFillColor.Rgb };
-                    xlConditionalFormat.Colors.Add(this.GetColor(color));
```
Pankraty: I like this chuck most of all )

ClosedXML_Tests/Excel/Styles/ColorTests.cs
```diff
@@ -38,5 +41,137 @@ public void DefaultColorIndex64isTransparentWhite()
             Assert.AreEqual(64, color.Indexed);
             Assert.AreEqual(Color.Transparent, color.Color);
         }
+
+        [Test]
+        public void CanConvertXlColorToColorType()
```
igitur: Please rename to `CanConvertXLColorToColorType`

ClosedXML/Utils/OpenXmlHelper.cs
```diff
@@ -1,14 +1,50 @@
...
+using Drawing = System.Drawing;
+using X14 = DocumentFormat.OpenXml.Office2010.Excel;
 
 namespace ClosedXML.Utils
 {
     internal static class OpenXmlHelper
     {
         #region Public Methods
 
+        /// <summary>
```
igitur: Please capitalise `XML`

igitur:
Squash and rebase too please

ClosedXML/Utils/OpenXmlHelper.cs
```diff
@@ -13,5 +18,42 @@ public static bool GetBooleanValueAsBool(BooleanValue value, bool defaultValue)
...
+
+        /// <summary>
+        /// Convert color in ClosedXML representation to specified OpenXML type.
+        /// </summary>
+        /// <typeparam name="T">OpenXML color type to convert the argument to.</typeparam>
+        /// <param name="xlColor">Color in ClosedXML format.</param>
+        /// <param name="isDifferential">Flag specifiying that the color should be saved in
+        /// differential format (affects the transparent color processing).</param>
+        /// <returns>The original color in OpenXXL format.</returns>
+        public static T ToOpenXmlColor<T>(XLColor xlColor, bool isDifferential = false)
```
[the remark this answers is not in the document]
Pankraty: Should I leave the class name `OpenXmlHelper` or change it as well?

Pankraty:
Rebased

ClosedXML/Utils/OpenXmlHelper.cs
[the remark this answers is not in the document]
igitur: It's OK, let's leave it as is.

PR #950, opened 2018-07-12
Fix loading transparent fill

Pankraty:
This PR fixes the regression introduced in #862 regarding transparent fills.


igitur:
Thanks. Not sure if it's related, but I see the loss of the horizontal alignment is also a regression. Can you please check if it's related to this PR?

Pankraty:
Already started

PR #962, opened 2018-07-26
Add ability to set cell value to an enum value

### LINKED ISSUE #953: IXLCell.SetValue handles Enums differently in 0.93
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

e.g. 0.93/0.92.1

Currently, in 0.93, IXLCell.SetValue(parameter) throws an ArgumentException ("Unable to set cell value to...") when the type of the parameter is an enum.

In previous versions, calling SetValue with an enum argument would not throw an exception, and seemed to call ToString in the background before setting the value.

Yes (tested v0.92.1 and 0.89).

## Reproducibility
```c#
 class Program
    {
        {
            var wb = new XLWorkbook();
            var ws = wb.AddWorksheet("test");
            var value = TestEnum.I;
            ws.Cell(1, 1).SetValue(value);
        } 
    }
    public enum TestEnum
    {
        I,
        Am,
        An,
        Enum
    }
```

igitur:
Fixes #953 

@Pankraty We can consider your data type register later.

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -245,7 +245,7 @@ namespace ClosedXML.Excel
         {
             string parsedValue;
             bool parsed;
-            if (value is String && acceptString || value is char || value is Guid)
+            if (value is String && acceptString || value is char || value is Guid || value is Enum)
             {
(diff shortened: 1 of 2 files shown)


Pankraty:
I don't mind. Will review the changes later.

PR #978, opened 2018-08-22
Add table helper methods

igitur:
Some helper methods added.

I'm not convinced about replacing `ArgumentException` with `ClosedXMLEntityNotFoundException`. I can see the use of having our own exception, but it also feels inconsistent with other places where `ArgumentException` is thrown.

### DIFF
--- a/ClosedXML/Excel/IXLWorkbook.cs
+++ b/ClosedXML/Excel/IXLWorkbook.cs
@@ -223,6 +223,15 @@ namespace ClosedXML.Excel
 
         XLWorkbook SetUse1904DateSystem(Boolean value);
 
+        /// <summary>
+        /// Gets the Excel table of the given name
+        /// </summary>
+        /// <param name="tableName">Name of the table to return.</param>
+        /// <param name="comparisonType">One of the enumeration values that specifies how the strings will be compared.</param>
+        /// <returns>The table with given name</returns>
+        /// <exception cref="ArgumentOutOfRangeException">If no tables with this name could be found in the workbook.</exception>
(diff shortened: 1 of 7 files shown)


ClosedXML_Tests/Excel/Tables/TablesTests.cs
```diff
@@ -1009,6 +1009,17 @@ public void SavingTableWithNullDataRangeThrowsException()
...
         }
 
+        [Test]
+        public void TableNotFound()
+        {
+            using (var wb = new XLWorkbook())
+            {
+                var ws = wb.AddWorksheet("Sheet1");
+                Assert.Throws<ClosedXMLEntityNotFoundException>(() => ws.Table("dummy"));
+                Assert.Throws<ClosedXMLEntityNotFoundException>(() => wb.Table("dummy"));
```
Pankraty: Why two times?

igitur:
Ready for review.

ClosedXML/Excel/IXLWorkbook.cs
```diff
@@ -223,6 +223,14 @@ public interface IXLWorkbook : IDisposable
 
         XLWorkbook SetUse1904DateSystem(Boolean value);
 
+        /// <summary>
+        /// Gets the Excel table of the given name
+        /// </summary>
+        /// <param name="tableName">Name of the table to return.</param>
+        /// <param name="comparisonType">One of the enumeration values that specifies how the strings will be compared.</param>
+        /// <returns>The table with given name or </returns>
```
Pankraty: The table with given name or _what_? The implementation below throws the exception if not found.

igitur: ok done

PR #999, opened 2018-09-12
Allow names to start with backslash

### LINKED ISSUE #997: When workbook contains NamedRange starting with "\" the workook can't be opened. Exception message is "The named range name '\y' does not begin with a letter or an underscore"
- [x] Bug

**Version of ClosedXML**

0.93.0
0.93.1
master
develop

It is not possible to create named range starting with \ in excel but it happens that files contains it.

I don't think so

## Reproducibility
```c#
public void Main()
{
   var wb = new XLWorkbook(@"D:\namedRange.xlsx");
}
```
[namedRange.xlsx](https://github.com/ClosedXML/ClosedXML/files/2375616/namedRange.xlsx)

igitur:
http://www.criticaltosuccess.com/allowed-characters-in-range-names/

> Range names must start with a letter, underscore, or backslash. Following characters must be letters or numbers, periods, or underscores. You cannot use spaces or any character combination that looks like a cell reference.

Fixes #997

### DIFF
--- a/ClosedXML/XLHelper.cs
+++ b/ClosedXML/XLHelper.cs
@@ -333,9 +333,10 @@ namespace ClosedXML.Excel
                 return false;
             }
 
-            if (newName[0] != '_' && !char.IsLetter(newName[0]))
+            var allowedFirstCharacters = new[] { '_', '\\' };
+            if (!allowedFirstCharacters.Contains(newName[0]) && !char.IsLetter(newName[0]))
             {
-                message = $"The {objectType} name '{newName}' does not begin with a letter or an underscore";
+                message = $"The {objectType} name '{newName}' does not begin with a letter, an underscore or a backslash.";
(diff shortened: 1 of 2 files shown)


ClosedXML/XLHelper.cs
```diff
@@ -333,7 +333,8 @@ public static Boolean ValidateName(String objectType, String newName, String old
                 return false;
             }
 
-            if (newName[0] != '_' && !char.IsLetter(newName[0]))
+            var allowedFirstCharacters = new[] { '_', '\\' };
+            if (!allowedFirstCharacters.Contains(newName[0]) && !char.IsLetter(newName[0]))
             {
                 message = $"The {objectType} name '{newName}' does not begin with a letter or an underscore";
```
Pankraty: "... or a backslash"

PR #1001, opened 2018-09-17
 Implement sensible ToString override for XLCell

### LINKED ISSUE #980: Implement something sensible for IXLCell.ToString()

igitur:
Fixes #980

### DIFF
--- a/ClosedXML/Excel/Cells/IXLCell.cs
+++ b/ClosedXML/Excel/Cells/IXLCell.cs
@@ -384,6 +384,13 @@ namespace ClosedXML.Excel
 
         XLTableCellType TableCellType();
 
+        /// <summary>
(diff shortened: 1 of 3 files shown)


ClosedXML/Excel/Cells/IXLCell.cs
```diff
@@ -383,6 +383,10 @@ public interface IXLCell
 
         XLTableCellType TableCellType();
 
+        string ToString();
```
Pankraty: Is it necessary to define a `ToString()` method in the interface? All CLR types have it anyway.

igitur: True!

Pankraty (approved):

PR #1012, opened 2018-09-28
Added '[' as a valid character to begin Named Ranges

### LINKED ISSUE #1011: When workbook contains NamedRange starting with "[" the workbook can't be opened. Exception message is "The named range name '\y' does not begin with a letter or an underscore"
- [x ] Bug

**Version of ClosedXML**
0.93.1

It's possible to create a named range starting with "[" in excel. ClosedXml should be able to open a spreadsheet using this format.

No

## Reproducibility

```c#
public void Main()
{
   var wb = new XLWorkbook("Test Workbook.xlsx");
}
```
[Test Workbook.xlsx](https://github.com/ClosedXML/ClosedXML/files/2425615/Test.Workbook.xlsx)

MexicanCustard:
#### What's this PR do?
Added '[' as a valid starting character in named ranges (fixes #1011)

### DIFF
--- a/ClosedXML/XLHelper.cs
+++ b/ClosedXML/XLHelper.cs
@@ -333,7 +333,7 @@ namespace ClosedXML.Excel
                 return false;
             }
 
-            var allowedFirstCharacters = new[] { '_', '\\' };
+            var allowedFirstCharacters = new[] { '_', '\\', '[' };
             if (!allowedFirstCharacters.Contains(newName[0]) && !char.IsLetter(newName[0]))
             {
                 message = $"The {objectType} name '{newName}' does not begin with a letter, an underscore or a backslash.";


Pankraty:
Hmm, the logic in Excel seems to be more complicated. 
`[test` - invalid
`[test]` - invalid
`[test]1` - invalid
`[test]test1` - **valid**
`[test][test]test` - invalid
`[test]test]test` - invalid
`test[test]test` - invalid
`[]test` - invalid
`[ ]test` - invalid
`[_]test` - **valid**
`[123]test` - **valid**
`[ 123]test` - invalid

So the named range title may only have square brackets if and only if 
1) there is a single pair of square brackets
2) the named range name starts with the `[`
3) the expression in brackets contains letters, digits, `_` or `\`
4) the expression after the brackets contains letters, digits, `_` or `\` and does not start with a digit.

igitur:
Is there any documentation available about what the exact rules are for named ranges' names? Maybe we should peak into LibreOffice's source code?

igitur:
Interesting. LibreOffice does not allow this.

igitur:
Superceded by #1036

Issue #1010, opened 2018-09-24
Formula not supported: TEXTJOIN()


James-Whitfield:
**Do you want to request a *feature* or report a *bug*?**
- [] Bug
- [X] Feature

**Version of ClosedXML**

0.93.1

**What is the current behavior?**

TEXTJOIN() formula function not supported

**What is the expected behavior or new feature?**

https://support.office.com/en-us/article/textjoin-function-357b449a-ec91-49d0-80c3-0e8fc845691c

**Did this work in previous versions of our tool?  Which versions?**

No

## Reproducibility

**Code to reproduce problem:**
```c#
using ClosedXML.Excel;
using System;

namespace ConsoleApp1
{
    class Program
    {
        static void Main(string[] args)
        {
            var wb = new XLWorkbook();
            var ws = wb.AddWorksheet("Sheet1");
            ws.Cell("A1").FormulaA1 = "=TEXTJOIN(\",\",TRUE,0,1,2,3)";
            // TEXTJOIN() Formula not supported 
            var x = ws.Cell("A1").GetString();
            Console.WriteLine(x);
        }
    }
}

```
- [ ] I attached a sample spreadsheet.  (You can drag files on to this issue)

igitur:
Thanks. Many formulas are not supported yet. We rely on community contributions to extend ClosedXML in this regard. So have a go at it.

James-Whitfield:
Thank you for the quick reply. I will look into adding support for this.

igitur:
Closed by #1015

PR #1015, opened 2018-10-05
Feature/#1010 text join func

James-Whitfield:
#### What's this PR do?

This PR implements the excel function TEXTJOIN which joins strings together with a delimiter.
the docs for this can be viewed [here](https://support.office.com/en-us/article/textjoin-function-357b449a-ec91-49d0-80c3-0e8fc845691c).

They can start by running the new unit test added to the project.

This can be manually tested by running the unit test or creating a formula which uses TEXTJOIN.
 
This function was not supported. So I raised it as a issue #1010 and it was suggested that I have a go at implementing it. Here is my attempt. 
#### Questions:
- Is there a blog post?
No
Might need to add it to the list of supported functions 
No


ClosedXML/Excel/CalcEngine/Functions/Text.cs
```diff
@@ -275,6 +276,63 @@ private static object _Text(List<Expression> p)
...
+        /// <param name="p">Parameters</param>
+        /// <returns> string </returns>
+        /// <exception cref="ApplicationException">
+        /// Delimiter in first param must be a string
+        /// or
+        /// Second param must be a boolen (TRUE/FALSE)
+        /// </exception>
+        private static object TextJoin(List<Expression> p)
+        {
+            if( p.Count > 254) throw new ApplicationException("Max number of params 254");
```
Pankraty: I suspect this check is not necessary since the maximum number of parameters is set in `RegisterFunction`. But to be completely sure I'd add a test for this too...

ClosedXML_Tests/Excel/CalcEngine/FunctionsTests.cs
```diff
@@ -598,6 +598,42 @@ public void SumSq()
             Assert.AreEqual(25.0, actual);
         }
 
+        [Test]
```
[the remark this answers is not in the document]
Pankraty: And move it to TextTests class please.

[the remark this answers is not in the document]
James-Whitfield: Done

ClosedXML/Excel/CalcEngine/Functions/Text.cs
James-Whitfield: You are correct this test is not necessary. Do you still want a test case for it?

ClosedXML/Excel/CalcEngine/Functions/Text.cs
```diff
@@ -275,6 +276,61 @@ private static object _Text(List<Expression> p)
...
+        /// or
+        /// Second param must be a boolen (TRUE/FALSE)
+        /// </exception>
+        private static object TextJoin(List<Expression> p)
+        {
+            List<string> values = new List<string>();
+            var p0 = p[0].Evaluate();
+            var p1 = p[1].Evaluate();
+
+            if (!(p0 is string)) throw new ApplicationException("Delimiter in first param must be a string");
```
igitur: I don't think we should be throwing `ApplicationException` here. What does Excel return when you input a non-string value in that parameter?

ClosedXML/Excel/CalcEngine/Functions/Text.cs
```diff
@@ -275,6 +276,61 @@ private static object _Text(List<Expression> p)
...
+            var delimiter = (string) p0;
+
+            if (!(p1 is bool)) throw new ApplicationException("Second param must be a boolen (TRUE/FALSE)");
+            var ignoreEmptyStrings = (bool) p1;
+
+
+            foreach(Expression param in p.Skip(2))
```
igitur: Please fix whitespace formatting. We use the CodeMaid Visual Studio plugin. Run CodeMaid Clean on all affected files.

igitur: Use implicit variable type declaration, i.e. `foreach (var param in p.Skip(2))`

ClosedXML/Excel/CalcEngine/Functions/Text.cs
```diff
@@ -275,6 +276,61 @@ private static object _Text(List<Expression> p)
...
+            {
+                if (param is XObjectExpression)
```
igitur: Can this logic (and the few lines below) not be simplified?

James-Whitfield: Probably, I am not a strong c# developer. If you have any specific suggestions I would be happy to try and implement them.

ClosedXML/Excel/CalcEngine/Functions/Text.cs
```diff
@@ -192,7 +193,7 @@ private static string WildcardToRegex(string pattern)
 
         private static object Search(List<Expression> p)
         {
-            var search = WildcardToRegex((string)p[0]);
+            var search = WildcardToRegex(p[0]);
```
igitur: Why did this line change? Did CodeMaid do this?

James-Whitfield: yes, I ran code made on the files I had changed

ClosedXML_Tests/Excel/CalcEngine/FunctionsTests.cs
```diff
@@ -1,10 +1,9 @@
-using System;
 using ClosedXML.Excel;
 using NUnit.Framework;
+using System;
```
igitur: I realise CodeMaid fixed the whitespace issues, but as this file was essentially unchanged (logically), all whitespace changes should also be reverted. Just revert the entire file.

Pankraty: Done

igitur:
Thanks for the work on this, @Pankraty .

I think this PR should depend on #1035 . Once #1035 has been merged, I think the `TextJoin` function can be improved. I'll make notes in the review.

Pankraty:
I agree

ClosedXML/Excel/CalcEngine/Functions/Text.cs
```diff
@@ -276,6 +277,72 @@ private static object _Text(List<Expression> p)
...
+        /// <returns> string </returns>
+        /// <exception cref="ApplicationException">
+        /// Delimiter in first param must be a string
+        /// or
+        /// Second param must be a boolen (TRUE/FALSE)
+        /// </exception>
+        private static object TextJoin(List<Expression> p)
+        {
+            List<string> values = new List<string>();
+            var p0 = p[0].Evaluate();
```
igitur: We should prefer `var delimiter = (string)p[0];`

The `Expression.cs` file contains overloads for implicit casting that takes into account all the data types. The fix in #1035 is however required for this.

ClosedXML/Excel/CalcEngine/Functions/Text.cs
```diff
@@ -276,6 +277,72 @@ private static object _Text(List<Expression> p)
...
+        /// <param name="p">Parameters</param>
+        /// <returns> string </returns>
+        /// <exception cref="ApplicationException">
+        /// Delimiter in first param must be a string
+        /// or
+        /// Second param must be a boolen (TRUE/FALSE)
+        /// </exception>
+        private static object TextJoin(List<Expression> p)
+        {
+            List<string> values = new List<string>();
```
igitur: Small things, but use `var values = ...` where possible.

ClosedXML_Tests/Excel/CalcEngine/TextTests.cs
```diff
@@ -489,6 +489,52 @@ public void Text_String_Input()
...
+            ws.Cell("B2").Value = "D";
+
+            ws.Cell("C1").FormulaA1 = formula;
+            var a = ws.Cell("C1").Value;
+
+            Assert.AreEqual(expectedOutput, a.ToString());
+        }
+
+        [TestCase("=TEXTJOIN(\",\", FALSE, D1:D32769)", "The value is too long")]
+        [TestCase("=TEXTJOIN(\",\", \"0\", A1:B2)", "The second argument is invalid")]
```
igitur: Is this what Excel does for this case? I have Excel 2013, so can't test it. I would actually prefer if we could cast the `"0"` to an actual `bool` of value `false`. That makes more sense to me.

ClosedXML/Excel/CalcEngine/Functions/Text.cs
```diff
@@ -276,6 +277,72 @@ private static object _Text(List<Expression> p)
...
+
+                    values.AddRange((range as XLRange).CellValues().Cast<object>().Select(ValueToString));
+                }
+                else
+                {
+                    values.Add(ValueToString(param.Evaluate()));
+                }
+            }
+
+            if (ignoreEmptyStrings)
```
igitur: Rather test this in the `foreach` loop and not add the value if applicable.

ClosedXML/Excel/CalcEngine/Functions/Text.cs
```diff
@@ -276,6 +277,72 @@ private static object _Text(List<Expression> p)
...
+                    if (!(tableArray.Value is CellRangeReference rangeReference))
+                        throw new NoValueAvailableException("tableArray has to be a range");
+
+                    var range = rangeReference.Range;
+
+                    values.AddRange((range as XLRange).CellValues().Cast<object>().Select(ValueToString));
+                }
+                else
+                {
+                    values.Add(ValueToString(param.Evaluate()));
```
igitur: Rather `values.Add((string)param);`

igitur:
Here is my alternative version of `TextJoin()`, which depends on #1035 

```c#
private static object TextJoin(List<Expression> p)
{
    var values = new List<string>();
    var delimiter = (string)p[0];
    var ignoreEmptyStrings = (bool)p[1];

    foreach (var param in p.Skip(2))
    {
        if (param is XObjectExpression tableArray)
        {
            if (!(tableArray.Value is CellRangeReference rangeReference))
                throw new NoValueAvailableException("tableArray has to be a range");

            var range = rangeReference.Range;
            IEnumerable<string> cellValues;
            if (ignoreEmptyStrings)
                cellValues = range.CellsUsed()
                    .Select(c => c.GetString())
                    .Where(s => !string.IsNullOrEmpty(s));
            else
                cellValues = range.Cells().Select(c => c.GetString());

            values.AddRange(cellValues);
        }
        else
        {
            values.Add((string)param);
        }
    }

    var retVal = string.Join(delimiter, values);

    if (retVal.Length > 32767)
        throw new CellValueException();

    return retVal;
}
```

Pankraty:
> Here is my alternative version of TextJoin(), which depends on #1035

Looks much nicer! But pay attention that `range.Cells().Select(c => c.GetString());` is quite insufficient for column-wide ranges (e.g. `A:B`).

igitur:
> Looks much nicer! But pay attention that range.Cells().Select(c => c.GetString()); is quite insufficient for column-wide ranges (e.g. A:B).

Yeah, was about to discuss that with you on WhatsApp. I think using `CellRangeReference.GetEnumerator()` would be better, right?

ClosedXML_Tests/Excel/CalcEngine/TextTests.cs
igitur: I suggest we accept strings that are parsable to `bool` there too. Of course `TEXTJOIN(",", "asdf", A1:B2)` should throw a `CellValueException`.

igitur:
Thanks @James-Whitfield and @Pankraty for the contributions.

PR #1021, opened 2018-10-11
Return simpler range address values

igitur:
Let XLRange.ToString() and XLRangeAddress.ToString() return simpler values when appropriate.

For example, for full row, return `3:5` instead of `A3:XFD5`. Same for columns. 

For entire sheet, return `1:1048576`. This is what Excel does.


ClosedXML/Excel/Ranges/XLRangeAddress.cs
```diff
@@ -213,11 +212,20 @@ public String ToStringFixed()
...
-                    FirstAddress.ToStringRelative(), ":",
-                    LastAddress.ToStringRelative())
-                : "#REF!";
+            string address;
+            if (!IsValid)
+                address = "#REF!";
+            else
+            {
+                if (IsEntireSheet())
+                    address = "1:1048576";
```
Pankraty: I think we'd better use `XLHelper.MaxRowNumber` everywhere in this file.

ClosedXML/Excel/Ranges/XLRangeBase.cs
```diff
@@ -513,18 +513,21 @@ public virtual Boolean IsEmpty(Boolean includeFormats)
 
         public virtual Boolean IsEntireRow()
         {
-            return RangeAddress.FirstAddress.ColumnNumber == 1
-                   && RangeAddress.LastAddress.ColumnNumber == XLHelper.MaxColumnNumber;
+            return RangeAddress.IsEntireRow();
```
Pankraty: Add check `RangeAddress.IsValid`, otherwise the exception will occur (I suspect). Same in the next 2 methods

ClosedXML/Excel/Ranges/XLRangeBase.cs
```diff
@@ -513,18 +513,21 @@ public virtual Boolean IsEmpty(Boolean includeFormats)
...
         public virtual Boolean IsEntireColumn()
         {
-            return RangeAddress.FirstAddress.RowNumber == 1
-                   && RangeAddress.LastAddress.RowNumber == XLHelper.MaxRowNumber;
+            return RangeAddress.IsEntireColumn();
+        }
+
+        public virtual Boolean IsEntireSheet()
+        {
+            return RangeAddress.IsEntireSheet();
```
Pankraty: You made these methods `virtual` to override them in XLRow, XLColumn, XLWorksheet, I assume. Did you forget this?

ClosedXML/Excel/Ranges/XLRangeBase.cs
igitur: Ok, I'll add the `IsValid` check to the methods in `RangeAddress`

ClosedXML/Excel/Ranges/XLRangeBase.cs
igitur: Ah, yes, the `IsEntireSheet()` method doesn't need to be virtual. Copy and paste error.

PR #1022, opened 2018-10-11
Small bug fixes and enhancements

igitur:
Some small bug fixes and enhancements.

- Fix small bug with last address' parameters in Worksheet.Range() lookup
- Use pattern matching
- Add test for setting cell value to range
- Rework rangesToMerge to standard Linq functions
- Throw ArgumentOutOfRangeException for invalid parameters in Row() and Column() lookup


ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -1934,12 +1934,8 @@ private bool SetRichText(object value)
...
         {
             var asRange = rangeObject as XLRangeBase;
-            if (asRange == null)
-            {
-                var tmp = rangeObject as XLCell;
-                if (tmp != null)
-                    asRange = tmp.AsRange();
-            }
+            if (asRange == null && rangeObject is XLCell tempCell)
+                asRange = tempCell.AsRange();
```
Pankraty: How about this?
```
var asRange = (rangeObject as XLRangeBase)
           ?? (rangeObject as XLCell)?.AsRange();
```

PR #1025, opened 2018-10-11
Lazy initialize graphics

### LINKED ISSUE #1020: ClosedXML fails in Azure Functions v2 for non-trivial operations
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.93.1 (but also latest dev build)

Using ClosedXML in Azure Functions v2 throws a "Platform Not Supported" exception when requiring the use of any functionality from XLHelper (e.g. resolving an A1-type reference)

Ideally, normal functionality would be available - but since this appears to be a System.Drawing issue, having non-graphics related operations succeed would be sufficient.

Yes, in Azure Functions v1 (.NET Framework based)

This needs to be run as part of an Azure Functions v2 project, not a console app.  Sorry!

```c#
    public static class Function1
    {
        [FunctionName("TestClosedXML")]
        public static async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = null)]HttpRequest req, ILogger log)
        {
            log.LogInformation("Function start.");

            var worksheet = workbook.AddWorksheet("Sheet 1");
            worksheet.Cell("A1").SetValue<string>("Hello");

            return new OkObjectResult("Success");
        }
    }
```

**Problem identification:**

XLHelper.cs has a static dependency on System.Drawing.Bitmap (lines 24 and 25):
```c#
internal static readonly Graphics Graphic = Graphics.FromImage(new Bitmap(200, 200));
internal static readonly Double DpiX = Graphic.DpiX;
```
The constructor for System.Drawing.Bitmap throws the PlatformNotSupported exception because Azure Functions v2 (Web Apps in general) run in a sandbox which denies access to Win32 functions, even when those are supported by the .Net Standard assemblies (see [this article](https://github.com/projectkudu/kudu/wiki/Azure-Web-App-sandbox#win32ksys-user32gdi32-restrictions)).

Can we move that into a function that is only called if necessary? Graphics-related functions might still fail (but could perhaps be caught in a cleaner manner) but non-graphics operations would succeed.

igitur:
Fixes #1020

### DIFF
--- a/ClosedXML/XLHelper.cs
+++ b/ClosedXML/XLHelper.cs
@@ -19,8 +19,8 @@ namespace ClosedXML.Excel
         public const String MaxColumnLetter = "XFD";
         public const Double Epsilon = 1e-10;
 
-        internal static readonly Graphics Graphic = Graphics.FromImage(new Bitmap(200, 200));
-        internal static readonly Double DpiX = Graphic.DpiX;
+        internal static readonly Graphics Graphics = Graphics.FromImage(new Bitmap(200, 200));
+        internal static readonly Double DpiX = Graphics.DpiX;
         internal static readonly NumberStyles NumberStyle = NumberStyles.AllowDecimalPoint | NumberStyles.AllowLeadingSign | NumberStyles.AllowLeadingWhite | NumberStyles.AllowTrailingWhite | NumberStyles.AllowExponent;
         internal static readonly CultureInfo ParseCulture = CultureInfo.InvariantCulture;
 
--- a/ClosedXML/XLHelper.cs
+++ b/ClosedXML/XLHelper.cs
@@ -19,8 +19,10 @@ namespace ClosedXML.Excel
         public const String MaxColumnLetter = "XFD";
         public const Double Epsilon = 1e-10;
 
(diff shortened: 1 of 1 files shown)


ClosedXML/XLHelper.cs
```diff
@@ -21,8 +21,12 @@ public static class XLHelper
 
         private const Int32 TwoT26 = 26 * 26;
 
-        internal static readonly Graphics Graphic = Graphics.FromImage(new Bitmap(200, 200));
-        internal static readonly Double DpiX = Graphic.DpiX;
+        private static readonly Lazy<Graphics> graphics = new Lazy<Graphics>(() => Graphics.FromImage(new Bitmap(200, 200)));
+        internal static Graphics Graphic { get => graphics.Value; }
+
+        private static readonly Lazy<Double> dpiX = new Lazy<double>(() => Graphic.DpiX);
```
Pankraty: This field is not necessary. If you change the internal property to 
`internal static Double DpiX { get => Graphic.DpiX; }`
it'll be lazy-loaded already.

igitur:
@Pankraty Approved?

Issue #968, opened 2018-08-06
Data validation isn't saved correctly


tisaacs:
**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

Latest dev branch

**What is the current behavior?**

When saving a file with a large set of data validation rules, ClosedXML doesn't save them correctly.

For example, it transforms the XML from this:

```
<dataValidation type="textLength" operator="lessThanOrEqual" showInputMessage="1" showErrorMessage="1" prompt="not greater than 43 characters" sqref="C11:AH11 C19:AH19 C27:AH27 C35:AH35 C43:AH43 C51:AH51 C59:AH59 C67:AH67 C75:AH75 C83:AH83 C91:AH91 C99:AH99 C107:AH107 C115:AH115 C123:AH123 C131:AH131 C139:AH139 C147:AH147 C155:AH155 C163:AH163 C171:AH171 C179:AH179 C187:AH187 C195:AH195 C203:AH203 C211:AH211 C219:AH219 C227:AH227 C235:AH235 C307:AH307 C251:AH251 C259:AH259 C267:AH267 C275:AH275 C283:AH283 C291:AH291 C299:AH299 C243:AH243 C315:AH315 C323:AH323 C331:AH331 C339:AH339 C347:AH347 C355:AH355 C363:AH363 C371:AH371 C379:AH379 C387:AH387 C395:AH395 C403:AH403 C411:AH411 C419:AH419 C427:AH427 C435:AH435 C443:AH443 C451:AH451 C459:AH459 C467:AH467 C475:AH475 C483:AH483 C491:AH491 C499:AH499 C507:AH507 C515:AH515 C523:AH523 C531:AH531 C539:AH539 C547:AH547 C555:AH555 C563:AH563 C571:AH571 C579:AH579 C587:AH587 C595:AH595 C603:AH603 C611:AH611 C619:AH619 C627:AH627 C635:AH635 C643:AH643 C651:AH651 C659:AH659 C667:AH667 C675:AH675 C683:AH683 C691:AH691 C699:AH699 C707:AH707 C715:AH715 C723:AH723 C731:AH731 C739:AH739 C747:AH747 C755:AH755 C763:AH763 C771:AH771 C779:AH779 C787:AH787 C795:AH795 C803:AH803">
			<formula1>43</formula1>
		</dataValidation>
```
To this:

```
[...]

PR #1029, opened 2018-10-12
CellsUsed refactoring and fix

### LINKED ISSUE #1026: CellsUsed method does not consider data validation rules
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.93

When the data validation rules are applied to cells these cells may be not presented in the `CellsUsed()` results if they have neither data, nor styles, nor conditional formats.

Complete this.

No

```c#
        [Test]
        public void ApplyingDataValidationMakesCellNotEmpty()
        {
            var wb = new XLWorkbook();
            ws.Range("B2:B12").SetDataValidation()
                .Decimal.EqualOrGreaterThan(0);

            var usedCells = ws.CellsUsed(includeFormats: true).ToList();

            Assert.AreEqual(11, usedCells.Count);
            Assert.AreEqual("B2", usedCells.First().Address.ToString());
            Assert.AreEqual("B12", usedCells.Last().Address.ToString());
        }
```

Pankraty:
Fixes #1026 

I started this branch right after discovering the issue with CellsUsed method during #968 fix but then had to switch to other activities.

To do things right I started a large refactoring
* I added a flagged enum `XLCellsUsedOptions` similar to `XLClearOptions` (I'd prefer to keep them separate even though they have same values)
* I heavily refactored `CellsUsed`, `FirstCellUsed`, `LastCellUsed` and other methods to consider merged ranges, conditional formats, data validation rules, etc.
* I marked all methods using `includeFormat` flag as `[Obsolete]`. After a couple versions we can remove them I suppose.
* ~I found and fixed another bug, which perhaps I should address separately. When cells are merged they didn't keep their own styles and values instead of acquiring the style of the first cell in the range and clearing everything else (contents, data validations, conditional formats...)~ Moved to #1030


ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -1991,9 +1991,9 @@ private Boolean SetRange(Object rangeObject)
             {
                 if (!(asRange is XLRow || asRange is XLColumn))
                 {
-                    var maxRows = asRange.RowCount();
-                    var maxColumns = asRange.ColumnCount();
-                    Worksheet.Range(_rowNumber, _columnNumber, maxRows, maxColumns).Clear();
+                    var maxRow = _rowNumber + asRange.RowCount() - 1;
+                    var maxColumn = _columnNumber + asRange.ColumnCount() - 1;
+                    Worksheet.Range(_rowNumber, _columnNumber, maxRow, maxColumn).Clear();
```
igitur: This was fixed in https://github.com/ClosedXML/ClosedXML/pull/1022/commits/2cc52eba90992671e4d396f3b12a7207b9f7f5fe , right?

ClosedXML/Excel/Cells/XLCells.cs
```diff
@@ -24,15 +24,133 @@ public XLCells(bool usedCellsOnly, bool includeFormats, Func<IXLCell, Boolean> p
         {
             _usedCellsOnly = usedCellsOnly;
             _includeFormats = includeFormats;
-            _predicate = predicate;
+
+            if (predicate == null)
+                _predicate = c => true;
+            else
+                _predicate = predicate;
```
igitur: `_predicate = predicate ?? _ => true;`

Pankraty: Of course

ClosedXML/Excel/Cells/XLCell.cs
Pankraty: I noticed that too. I am going to rebase anyway, so this will be replaced with your solution.

Pankraty:
As for the bug with the merged ranges, I found that the issue is more complicated than I thought.
If I unmerge a range in the TransposeRange.xlsx I get the following result:
image

Both the "old" way and the "new" way produce incorrect result. Guess I'll extract this into a separate issue.

igitur:
Please see review comments in #1030

igitur:
Had a chance to look at this further? It just needs history rewriting. I'd like release soon.

Pankraty:
Will do it right away

Pankraty:
I've rebased and squashed the branch and eliminated most of the issues. But more refactoring is still needed in order to properly handle large ranges in conditional formats and data validation rules (your recent test CFShiftedTruncateRange now takes forever to execute) :(

Unfortunately, I don't have enough time for this today. Will try to do it during the week.

igitur (approved):

igitur:
Methods made obsolete in this PR can be removed now, right?

PR #1032, opened 2018-10-13
Change the way invalid ranges are saved

### LINKED ISSUE #1007: Woorkbook-scoped Named Ranges on a deleted sheet corrupt workbook.
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.93.1

If the workbook has a named range at its level which points to a sheet which then gets deleted, the saved workbook is corrupt. This can be solved by deleting the named range from the workbook.

Workbook should not be corrupt.

Unknown.

## Reproducibility

```c#
public void Main()
{
    {
        var sheetToDelete = workbook.AddWorksheet("Sheet To Delete");
        workbook.AddWorksheet("Blank Sheet"); // So we don't throw an exception on saving an empty workbook.

        var range = sheetToDelete.FirstCell().AsRange();

        // Add a range to the workbook.
        // This also works if opening an existing workbook with a
        // workbook-scoped named range.
        workbook.NamedRanges.Add("TestRange", range);

        // Delete the sheet to which the named range points.
        sheetToDelete.Delete();

        workbook.SaveAs("Test.xlsx");
    }
}
```
[Test.xlsx](https://github.com/ClosedXML/ClosedXML/files/2403287/Test.xlsx)

Pankraty:
Fixes #1007 

Ranges referring to the deleted worksheets must be saved as `#REF!`, not `#REF!$A$1:$B$2`, otherwise, Excel complains that the file is corrupted.
Besides, I found that `RangeList` is not initialized properly. It does not seem to affect any other functionality as `Ranges` property was still correct, and `ToString()` method aggregated `RangeList` items to a single string anyway. But after I changed the saving logic it became essential.

### DIFF
--- a/ClosedXML/Excel/NamedRanges/XLNamedRange.cs
+++ b/ClosedXML/Excel/NamedRanges/XLNamedRange.cs
@@ -15,7 +15,7 @@ namespace ClosedXML.Excel
         public XLNamedRange(XLNamedRanges namedRanges, String rangeName, String range, String comment = null)
             : this(namedRanges, rangeName, comment)
         {
-            RangeList.Add(range);
+            range.Split(',').ForEach(r => RangeList.Add(r));
         }
 
         public XLNamedRange(XLNamedRanges namedRanges, String rangeName, IXLRanges ranges, String comment = null)
--- a/ClosedXML/Excel/NamedRanges/XLNamedRange.cs
+++ b/ClosedXML/Excel/NamedRanges/XLNamedRange.cs
@@ -15,6 +15,7 @@ namespace ClosedXML.Excel
         public XLNamedRange(XLNamedRanges namedRanges, String rangeName, String range, String comment = null)
             : this(namedRanges, rangeName, comment)
         {
+            //TODO range.Split(',') may produce incorrect result if a worksheet name contains comma. Refactoring needed.
             range.Split(',').ForEach(r => RangeList.Add(r));
         }
 
--- a/ClosedXML/Excel/XLWorkbook_Save.cs
+++ b/ClosedXML/Excel/XLWorkbook_Save.cs
@@ -866,12 +866,15 @@ namespace ClosedXML.Excel
                 definedNames.AppendChild(definedName2);
             }
 
-            foreach (var nr in NamedRanges)
+            foreach (XLNamedRange nr in NamedRanges)
             {
+                var refersTo = string.Join(",", nr.RangeList
+                    .Select(r => r.StartsWith("#REF!") ? "#REF!" : r));
+
                 var definedName = new DefinedName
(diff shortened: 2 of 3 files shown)


ClosedXML/Excel/XLWorkbook_Save.cs
```diff
@@ -866,12 +866,15 @@ orderby worksheet.Position
                 definedNames.AppendChild(definedName2);
             }
 
-            foreach (var nr in NamedRanges)
+            foreach (XLNamedRange nr in NamedRanges)
```
igitur: I prefer
` foreach (var nr in NamedRanges.OfType<XLNamedRange>())`

igitur:
This issue again reminded me that named ranges needs to be reworked and instead of strings be stored as ranges. I started working on it and then saw that named ranges to e.g. `Table1[ColumnName]` break. They're currently stored as strings, and hence doesn't require parsing. If we change to parsing strings to ranges, we'll have to implement those types of ranges too.

ClosedXML/Excel/NamedRanges/XLNamedRange.cs
```diff
@@ -15,7 +15,7 @@ internal class XLNamedRange : IXLNamedRange
         public XLNamedRange(XLNamedRanges namedRanges, String rangeName, String range, String comment = null)
             : this(namedRanges, rangeName, comment)
         {
-            RangeList.Add(range);
+            range.Split(',').ForEach(r => RangeList.Add(r));
```
igitur: Just note that this implementation is quite naive, because a sheet name can contain a comma. But we do `Split(',')` in many places in this class. Once we implement a proper named range parser we can work on this.

ClosedXML/Excel/XLWorkbook_Save.cs
igitur: or `foreach (var nr in NamedRanges.Cast<XLNamedRange>())`

Pankraty:
> This is to mimic what Excel does. Excel changes the range only after saving (and then loses some information). It seems your code already works like this. Just add the assertions please.

We have two other tests that already cover this: `NamedRangesBecomeInvalidOnWorksheetDeleting`, `SavedNamedRangesBecomeInvalidOnWorksheetDeleting`

ClosedXML/Excel/NamedRanges/XLNamedRange.cs
Pankraty: Will add a TODO

igitur:
> We have two other tests that already cover this: NamedRangesBecomeInvalidOnWorksheetDeleting, SavedNamedRangesBecomeInvalidOnWorksheetDeleting

Ah ok

PR #1033, opened 2018-10-14
Optimize the performance of searching range intersections on small collections

### LINKED ISSUE #993: Saving a file having thousands of conditional formats takes too long
**Version of ClosedXML**

0.93.1.0

i cannot save as a new file from file available with XLWorkbook.SaveAs("newFilePath").
my file have size = 10076kb.
i was verified with file has small size (100kb) and good performance result.
 this is my code:
```
 using ClosedXML.Excel;
using System;
using System.IO;


namespace ConsoleApplication1
{
    class Program
    {
        {
            string filePath = @"[thm-viewer] Export_Data_Format_rev01";

            ExportDataExcel(filePath);

            Console.ReadLine();
        }
        public static void ExportDataExcel(string filePath)
        {
            if (File.Exists(filePath))
            {
                using (XLWorkbook wb = new XLWorkbook(filePath))
                {
                    wb.SaveAs("copyfile.xlsx");
                    Console.WriteLine("Done");
                }
            }
        }
    }
}
```

How do I copy large files with the XLWorkbook .SaveAs("copyfile.xlsx"); method?

**Did this work in previous versions of our tool? Which versions?**
No

Pankraty:
This PR partially fixes #993. I used a worksheet 2510 from the original file. Before this PR saving took 69.5 sec, now it takes only 12.7 sec. Still far from perfect but definitely much better.

To achieve this I modified a ranges indexing algorithm so it wouldn't apply to small collections (less than 20 items). The number 20 was chosen empirically: our QuadTree implementation is 10 levels deep and each level may require one or a few checks for intersections. Changing the constant value in a range (10 .. 70) barely affected the performance (the difference was <1%).

[993_sample2.xlsx](https://github.com/ClosedXML/ClosedXML/files/2476134/993_sample2.xlsx)

### DIFF
--- a/ClosedXML/Excel/ConditionalFormats/XLConditionalFormat.cs
+++ b/ClosedXML/Excel/ConditionalFormats/XLConditionalFormat.cs
@@ -30,8 +30,8 @@ namespace ClosedXML.Excel
 
                 var xxValues = xx.Values.Values.Where(v => v == null || !v.IsFormula).Select(v => v?.Value);
                 var yyValues = yy.Values.Values.Where(v => v == null || !v.IsFormula).Select(v => v?.Value);
-                var xxFormulas = x.Ranges.Any() ? xx.Values.Values.Where(v => v != null && v.IsFormula).Select(f => ((XLCell)x.Ranges.First().FirstCell()).GetFormulaR1C1(f.Value)) : null;
-                var yyFormulas = y.Ranges.Any() ? yy.Values.Values.Where(v => v != null && v.IsFormula).Select(f => ((XLCell)y.Ranges.First().FirstCell()).GetFormulaR1C1(f.Value)) : null;
+                var xxFormulas = x.Ranges.Count > 0 ? xx.Values.Values.Where(v => v != null && v.IsFormula).Select(f => ((XLCell)x.Ranges.First().FirstCell()).GetFormulaR1C1(f.Value)) : null;
+                var yyFormulas = y.Ranges.Count > 0 ? yy.Values.Values.Where(v => v != null && v.IsFormula).Select(f => ((XLCell)y.Ranges.First().FirstCell()).GetFormulaR1C1(f.Value)) : null;
 
                 var xStyle = xx.StyleValue;
                 var yStyle = yy.StyleValue;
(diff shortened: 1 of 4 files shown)


Pankraty:
Saving the entire file from the original task took ~27 minutes with this optimization.

Pankraty:
After a second round of optimizations the `Consolidate()` method is no longer the "bad guy" and the whole file is saved in 51 sec now.

image

igitur:
Great work.

sflanker:
I'm fairly certain this change introduces a significant bug. You're storing `XLRange` objects in a `HashSet` in the `Quadrant` data structure. However `XLRange` objects are mutable. For example if rows are inserted above an `XLRange` for merged cells, the `XLRange` gets updated with a new address, but the `XLRange` is still stored in it's old location in the `HashSet`. I'm planning on opening a bug with a repro case when I have time, but for now I have to revert back to a version prior  to this change. As a general rule you should not use a mutable object as the key in a sort or hash code based data structure. A quick fix might be to drop the `HashSet` from Quadrant and just use a List with linear search. You can already be sure that N will be small for those lists given the quad tree.

Pankraty:
Hi @sflanker, thanks for your feedback! 
First time I read your comment I thought: damn, how could I make such a childish mistake! Of course, I am aware of `HashSet` peculiarities and that made a mistake to be outstandingly stupid.
But then I realized that replacing a `HashSet` with a `List` would change a little because after the range is changed it can easily fall into another `Quadrant` - the lower one, the bigger one, the one to the left, etc.
Thinking further, I started to recollect that I was determined to implement the logic that would remove a range from the index during the editing and then put it back to the proper place. Could I simply forget about it?

And when I finally got to Visual Studio I found that I [had implemented](https://github.com/ClosedXML/ClosedXML/blob/develop/ClosedXML/Excel/XLWorksheet.cs#L1838) this indeed. I even covered this case with [tests](https://github.com/ClosedXML/ClosedXML/blob/develop/ClosedXML_Tests/Excel/Ranges/RangeIndexTest.cs#L116).
I admit, however, there might be bugs there, but I need repro cases.

PS. Pay attention, that if using `XLRangeIndex` caused the bug then good chances are that it was introduced by #803, half a year before #1033.

sflanker:
@Pankraty thank you for your quick response. That is a good point about the logic to update the index. Empirically I can say that there's something wrong, and that downgrading to 0.93.1 fixes the issues I'm seeing. However, I completely understand the need for a well defined repro case, and given the information you provide I'm not sure exactly where the bug is. I'll post an issue as soon as I have time to create a simple repro.

PR #1034, opened 2018-10-14
WIP Fix loading workbooks with auto filters using thousand separators.

### LINKED ISSUE #990: When opening a workbook there is a crash into the method LoadAutoFilterColumns
- [x] Bug

**Version of ClosedXML**

e.g. 0.93.1

When loading an excel file in french culture, il there is some autofilter filtering on this value "1 000", ClosedXml crash with the message

>  Input string was not in a correct format

No

```c#
public void Main()
{
    var wb = new XLWorkbook(@"D:\issue.xlsx");
}
```
- [x] [issue.xlsx](https://github.com/ClosedXML/ClosedXML/files/2361716/issue.xlsx)

Pankraty:
This PR fixes #990


ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -2033,10 +2033,14 @@ private static void LoadAutoFilterColumns(AutoFilter af, XLAutoFilter autoFilter
                     autoFilter.Filters.Add((int)filterColumn.ColumnId.Value + 1, filterList);
 
                     Boolean isText = false;
+
+                    var formatProvider = CultureInfo.InvariantCulture.Clone() as CultureInfo;
+                    formatProvider.NumberFormat.NumberGroupSeparator = " ";
```
igitur: I'm not convinced by this. Shouldn't we be using the user's own culture instead of forcefully specifying the number group separator? In the past I have dealt with Excel files that, when opened in Excel, will treat strings differently based on the current locale of the user. I suspect this is the case here.

ClosedXML_Tests/Excel/AutoFilters/AutoFilterTests.cs
```diff
@@ -212,5 +212,16 @@ public void ReapplyAutoFilter()
                 }
             }
         }
+
+        [Test]
+        public void CanLoadAutoFilterWithThousandsSeparator()
+        {
+            using (var stream = TestHelper.GetStreamFromResource(TestHelper.GetResourcePath(@"Examples\AutoFilter\AutoFilterWithThousandsSeparator.xlsx")))
```
igitur: The resource should rather be under `Other\AutoFilter\AutoFilterWithDifferentCultureThousandsSeparator.xlsx`. 

The `Examples` folder is supposed to keep reference files for subclasses of `IExample`'s output.

ClosedXML/Excel/XLWorkbook_Load.cs
Pankraty: I thought so too but after I set my local settings to use a comma as a thousand separator I still got `10 000.00` in the result file.

ClosedXML_Tests/Excel/AutoFilters/AutoFilterTests.cs
Pankraty: Done

ClosedXML/Excel/XLWorkbook_Load.cs
igitur: You must reapply the filter in Excel, then you'll see the different handling.

igitur:
Fixed by #1037

PR #1036, opened 2018-10-15
Don't validate named ranges when loading file

### LINKED ISSUE #1011: When workbook contains NamedRange starting with "[" the workbook can't be opened. Exception message is "The named range name '\y' does not begin with a letter or an underscore"

igitur:
Fixes #1011 
Supercedes #1012 

We already don't validate the named range address when loading from file. Now also don't validate the name. 

But we still enforce the rules when adding named ranges from the ClosedXML API


ClosedXML/Excel/NamedRanges/XLNamedRange.cs
```diff
@@ -76,6 +87,11 @@ public String Name
             }
         }
 
+        internal void SetNameWithoutValidation(string value)
```
Pankraty: Why not `private`?

ClosedXML/Excel/NamedRanges/XLNamedRange.cs
```diff
@@ -13,22 +13,33 @@ internal class XLNamedRange : IXLNamedRange
         internal XLWorkbook Workbook => _namedRanges.Workbook;
 
         public XLNamedRange(XLNamedRanges namedRanges, String rangeName, String range, String comment = null)
-            : this(namedRanges, rangeName, comment)
+            : this(namedRanges, rangeName, validateName: true, comment)
```
Pankraty: `: this(namedRange, rangeName, validateName: true, range, comment)`
- this way you can leave the constructor's body empty.

Pankraty:
The approve button is not visible once I submitted the review. Everything looks fine for me.

igitur:
I'm still trying to clarify a few things in #990 . Notice that the test that I added overrides the default fr-FR decimal symbol. But maybe that shouldn't happen.

PR #1037, opened 2018-10-16
Fix loading workbooks with auto filters using thousand separators.

### LINKED ISSUE #990: When opening a workbook there is a crash into the method LoadAutoFilterColumns

igitur:
Fixes #990 
Supercedes #1034


igitur:
@Pankraty Here is my suggested fix for the issue. The filter gets applied based on the user's locale.

PR #1041, opened 2018-10-22
Autofilter with multiple columns with filters

igitur:
Fixes https://stackoverflow.com/questions/43324643/closedxml-excel-filter-rows-by-values-in-multiple-columns

Filters on multiple columns weren't treated correctly. I rewrote the `Reapply()` method.
Also improved the loading of custom filters, e.g. `string starts with 'J'`


ClosedXML/Excel/AutoFilters/XLAutoFilter.cs
```diff
@@ -62,58 +62,56 @@ public IXLAutoFilter Reapply()
...
-                                        row.WorksheetRow().Unhide();
-                                    else
-                                        row.WorksheetRow().Hide();
-                                }
-                            }
-                            else if (match)
-                                row.WorksheetRow().Unhide();
+                                columnFilterMatch &= filterMatch;
+                            else
+                                columnFilterMatch |= filterMatch;
```
Pankraty: What do you think about a small optimization here? Say,
```
if (filter.Connector == XLConnector.And)
{
    columnFilterMatch &= filterMatch;
    if (!columnFilterMatch) break;
}
else
{
    columnFilterMatch |= filterMatch;
    if (columnFilterMatch) break;
}

File: ClosedXML/Excel/AutoFilters/XLFilterColumn.cs [hunk 2]
```diff
@@ -179,40 +179,40 @@ public void BelowAverage()
             LessThan(minValue).Or.GreaterThan(maxValue);
         }
 
+        public static Func<String, Object, Boolean> BeginsWithFunction => (value, input) => ((string)input).StartsWith(value, StringComparison.InvariantCultureIgnoreCase);
```
Pankraty: If I'm not mistaken, this way it'll produce a new instance of a function on each call. 
`public static Func<String, Object, Boolean> BeginsWithFunction { get; } = (value, input) => ((string)input).StartsWith(value, StringComparison.InvariantCultureIgnoreCase);` — must be cheaper, though I didn't test.

ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -1982,9 +1982,31 @@ private static void LoadAutoFilterColumns(AutoFilter af, XLAutoFilter autoFilter
 
                     foreach (var filter in filterColumn.CustomFilters.OfType<CustomFilter>())
                     {
-                        var xlFilter = new XLFilter { Value = filter.Val.Value, Connector = connector };
+                        XLFilter xlFilter = new XLFilter { Connector = connector };
```
Pankraty: `var`?

ClosedXML/Excel/AutoFilters/XLFilterColumn.cs
igitur: This syntax does actually create a property. I confirmed with ILSpy by decompiling to C#3 syntax:
```c#
public static Func<string, object, bool> BeginsWithFunction
{
    get
    {
        return (string value, object input) => ((string)input).StartsWith(value, StringComparison.InvariantCultureIgnoreCase);
    }
}
```

Property:
`public static Func<String, Object, Boolean> BeginsWithFunction => (value, input) => ((string)input).StartsWith(value, StringComparison.InvariantCultureIgnoreCase);`
Variable:
`public static Func<String, Object, Boolean> BeginsWithFunction = (value, input) => ((string)input).StartsWith(value, StringComparison.InvariantCultureIgnoreCase);`

igitur: Or wait... that still creates a new function for every `get` that is called, right?

Pankraty: I think so.

PR #1042, opened 2018-10-22
Fix for ShiftConditionalFormattingRows Error

andreasmo:
Fix ShiftConditionalFormattingRows if conditionalformat already reaches to the last row.

#### What's this PR do?
Fixes a bug that happens when rows are inserted and an existing conditional formatting already reaches to the end of a column (last row in sheet)
There are only two lines changed :-)
Create an Excel-File with a conditional formatting that reaches to the last rows. (I use an existing excel file as a template and add rows programatically)
#### Questions:
- Is there a blog post? 
No 
No
No


Pankraty:
Would you mind adding a unit test for this?

PR #1048, opened 2018-10-26
Shift data validation ranges when inserting / delete ranges

### LINKED ISSUE #1046: Inserting an adjacent row does not copy the data validation of cells

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.94.999.1469

Inserting an adjacent row does not copy the data validation of cells.

Inserting an adjacent row does copy the data validation. This would be similar to how ClosedXML copies other cell attributes (e.g. text wrapping), and would mimic the behavior of Excel.

I do not know if this worked in previous versions.

## Reproducibility

```c#
public void Main()
{
    using (var worksheet = workbook.AddWorksheet("WorksheetOne"))
    {
        worksheet.Range("A1:A3")
            .SetValue(1);

        //This will copy to new row on insert
        worksheet
            .Cell("A1")
            .Style
            .Alignment
            .SetWrapText(true);

        //This will not copy to new row on insert
        worksheet
            .Cell("A1")
            .DataValidation
            .Decimal
            .Between(0,1);

        worksheet
            .Row(2)
            .InsertRowsAbove(1);

        worksheet
            .Cell("A2")
            .InsertCellsAbove(1);

        workbook.SaveAs("WorkbookOne.xlsx");
    }
}
```
[WorkbookOne.xlsx](https://github.com/ClosedXML/ClosedXML/files/2516771/WorkbookOne.xlsx)

igitur:
Fixes #1046 
Closes #1042

### DIFF
--- a/ClosedXML/Excel/XLWorksheet.cs
+++ b/ClosedXML/Excel/XLWorksheet.cs
@@ -1151,14 +1151,14 @@ namespace ClosedXML.Excel
                         newRange = Range(cfAddress.FirstAddress.RowNumber,
                                          cfAddress.FirstAddress.ColumnNumber,
                                          cfAddress.LastAddress.RowNumber,
-                                         cfAddress.LastAddress.ColumnNumber + columnsShifted);
+                                         Math.Min(XLHelper.MaxColumnNumber, cfAddress.LastAddress.ColumnNumber + columnsShifted));
                     }
                     else if (cfAddress.FirstAddress.ColumnNumber >= firstCol)
                     {
                         newRange = Range(cfAddress.FirstAddress.RowNumber,
                                          Math.Max(cfAddress.FirstAddress.ColumnNumber + columnsShifted, firstCol),
                                          cfAddress.LastAddress.RowNumber,
-                                         cfAddress.LastAddress.ColumnNumber + columnsShifted);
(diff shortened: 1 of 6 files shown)


ClosedXML/Excel/XLWorksheet.cs
```diff
@@ -1173,6 +1174,52 @@ private void ShiftConditionalFormattingColumns(XLRange range, int columnsShifted
...
+                foreach (var dvRange in dvRanges)
+                {
+                    var dvAddress = dvRange.RangeAddress;
+                    IXLRange newRange;
+                    if (dvRange.Intersects(model))
+                    {
+                        newRange = Range(dvAddress.FirstAddress.RowNumber,
+                                         dvAddress.FirstAddress.ColumnNumber,
+                                         dvAddress.LastAddress.RowNumber,
+                                         dvAddress.LastAddress.ColumnNumber + columnsShifted);
```
Pankraty: There was an issue with the similar code (it was about a shifting of the conditional formats as I recall) when ranges became invalid after shifting if they included the last row or column. We should trim the result value by `XLHelper.MaxColumnNumber` (`MaxRowNumber` in the `ShiftDataValidationRows` method). An additional test case for this will also be useful.

igitur: Ah, you're referring to #1042?

Pankraty: Exactly

PR #1054, opened 2018-10-31
Implement workbook FileSharing and a few of the properties

igitur:
Reference: https://docs.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheet.filesharing?view=openxml-2.8.1

I implemented only `ReadOnlyRecommended` and `UserName`. The other properties are about password hashing, which I'm not touching now.

### DIFF
--- a/ClosedXML_Tests/Excel/Misc/XLWorkbookTests.cs
+++ b/ClosedXML_Tests/Excel/Misc/XLWorkbookTests.cs
@@ -1,7 +1,8 @@
-using System;
-using System.Linq;
 using ClosedXML.Excel;
 using NUnit.Framework;
+using System;
+using System.IO;
+using System.Linq;
 
 namespace ClosedXML_Tests.Excel
 {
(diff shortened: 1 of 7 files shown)


ClosedXML_Tests/Excel/Misc/XLWorkbookTests.cs
```diff
@@ -418,5 +418,28 @@ public void WbProtect5()
...
+        [Test]
+        public void FileSharingProperties()
+        {
+            using (var ms = new MemoryStream())
+            {
+                using (var wb = new XLWorkbook())
+                {
+                    wb.AddWorksheet("Sheet1").Cell("A1").Value = "Hello world!";
+                    wb.FileSharing.ReadOnlyRecommended = true;
+                    wb.FileSharing.UserName = Environment.UserName;
```
Pankraty: Better use some dummy value then rely on the environment.

igitur: This was actually deliberate, because I wanted to see whether it will work. And it does. So we might as well leave it? This will probably be the typical use case anyway.

Pankraty: OK, let's leave it

igitur:
Parked for v0.95

PR #1055, opened 2018-11-01
Use device independent measures

### LINKED ISSUE #510: Set picture dimensions in non-dpi-dependant units
- [ ] Bug
- [x] Feature

**Version of ClosedXML**

0.89.0

You can set the dimensions of a picture only in pixels.

You can **additionally** set the dimensions of a picture in non-dpi-dependant units (cm, inches, EMU... any of those will work since the conversion is trivial).

igitur:
Fixes #510

### DIFF
--- a/ClosedXML/Excel/Drawings/IXLPicture.cs
+++ b/ClosedXML/Excel/Drawings/IXLPicture.cs
@@ -1,6 +1,5 @@
 // Keep this file CodeMaid organised and cleaned
 using System;
-using System.Drawing;
 using System.IO;
 
 namespace ClosedXML.Excel.Drawings
@@ -15,27 +14,27 @@ namespace ClosedXML.Excel.Drawings
         /// </summary>
         XLPictureFormat Format { get; }
 
-        Int32 Height { get; set; }
+        IXLMeasure Height { get; set; }
 
         Int32 Id { get; }
 
         MemoryStream ImageStream { get; }
 
-        Int32 Left { get; set; }
+        IXLMeasure Left { get; set; }
 
         String Name { get; set; }
 
-        Int32 OriginalHeight { get; }
+        IXLMeasure OriginalHeight { get; }
 
-        Int32 OriginalWidth { get; }
+        IXLMeasure OriginalWidth { get; }
 
         XLPicturePlacement Placement { get; set; }
 
-        Int32 Top { get; set; }
+        IXLMeasure Top { get; set; }
 
         IXLCell TopLeftCell { get; }
 
-        Int32 Width { get; set; }
+        IXLMeasure Width { get; set; }
 
         IXLWorksheet Worksheet { get; }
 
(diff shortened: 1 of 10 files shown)


Pankraty:
@igitur FYI

As I showed you earlier sometimes some tests fail on my machine, but only in .Net Framework configurations, the Core works fine. 
2019-03-30 at 13 03 22
The example of differences:
2019-03-30 at 13 04 30

What's weird is that the very same tests succeed on the very same machine, I don't know for sure what it depends on. I suppose you may be interested to know that this PR does not fix it.

igitur:
If you open the actual files in Excel, which ones look more correct, ie have correctly fitted columns?

Pankraty:
net46
image

netcoreapp2.0
image

The later looks more correct.

Autosize in Excel:
image

igitur:
So those screenshots are from the generated files when the tests are run? Because that adds a new problem... net46 and netcoreapp2.0 generate different files!

What does the actual reference file in the repo look like?

igitur:
We could do a naive workaround if we just want the tests to pass. And that's to force the DPI to 96 for the test cases, instead of reading the system's DPI. But that doesn't really address the issue at hand.

Pankraty:
> So those screenshots are from the generated files when the tests are run? Because that adds a new problem... net46 and netcoreapp2.0 generate different files!
> 
> What does the actual reference file in the repo look like?

netcoreapp2.0 tests are OK so I assume their output is the same as the reference files. net40 and net46 produce the wrong result, but not always. Perhaps, it has some relation to the number of screens, their resolution, and which one is treated as the primary one. But again, I have no confidence.

igitur:
Can you confirm that `GraphicsUtils.Graphics.DpiX` returns different values for netcoreapp2.0 and net46? And what are those different values?

Pankraty:
Yes, for netcoreapp2.0 it is `96.0f` and for both net40 and net46 - `120.0f`. 

If I change the scale factor to 100% it becomes `96.0f` in all configurations. 
image

Same result when I am connected via RDP even though the remote and the local machines both have 125%. I often work on a remote and in this mode tests pass so I did not bother :)

Pankraty:
Oh, and if I change the primary display to the one having 100% scale factor it is also `96.0f` in all configurations.

igitur:
Do you think we should add a parameter to `LoadingOptions` for the user to specify the DPI settings? That way he can control the resolution. And we default to 96?

Pankraty:
Do we even need a custom DPI? Considering that .net Core uses hard-coded value of `96.0f` _and_ produces correct result whichever display is used.

igitur:
Well, the idea is to have the correct column width when doing `AdustToContents()`. The width depends on the rendered text, which depends on the font size in pixels, which depends on the dpi settings. That's how I understand it.

igitur:
> As I showed you earlier sometimes some tests fail on my machine, but only in .Net Framework configurations, the Core works fine.

@Pankraty Could you redo this comparison using the latest update on this PR? Also, how exactly do I reproduce this on my side? I don't have an HDPI monitor, so not sure if I can.

Pankraty:
All tests are green for me now!
BTW, I think the issue could be reproduced without HDPI monitor, just by setting non-default scale in display settings:
image

igitur:
> All tests are green for me now!

Ok, so is this PR even necessary at all? I do like the idea of `XLMeasure`. Do the test all pass on the `develop` branch using HDPI settings?

> BTW, I think the issue could be reproduced without HDPI monitor, just by setting non-default scale in display settings:

I was under the impression that changes only scale and that HDPI is reserved for true HDPI monitors.

Pankraty:
Huh, it seems to be more complicated. Right now, I cannot reproduce the issue even on old `develop` branch, and I think this is because I have only one monitor. In this case, `GraphicsUtils.Graphics.DpiX` returns `96f` in all configurations.
But while I worked in office I had two monitors with different resolutions. And when the high-res display was set primary with scale factor 125% `GraphicsUtils.Graphics.DpiX` would return `120f` and tests would fail.
Can't test it now, unfortunately.

igitur:
OK, let's park this for now. But the point about 2 monitors with different resolutions is very germane and we need to play around with that later. I have never had that setup.

PR #1072, opened 2018-11-19
Reference FastMember (unsigned) when building unsigned ClosedXML

igitur:
To avoid unnecessary binding redirects, we can reference FastMember (unsigned) when building unsigned ClosedXML.


igitur:
Is this a breaking change? I suspect so...

Issue #1071, opened 2018-11-16
Poor performance with data validation


zabulus:
## Read and complete the full issue template

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

**Version of ClosedXML**
tried 0.93.1 and 0.94.999.1504

**What is the current behavior?**
Generation of excel file takes too long. Have waited for 10 minutes, and still not completed.

**Did this work in previous versions of our tool?  Which versions?**
Yes. Using the 0.87.1 version the same generation takes around 1 sec.

## Reproducibility
**Code to reproduce problem:**
```c#
public void Main()
{
            XLWorkbook template = new XLWorkbook(@"Template.xlsx");
            var ws = template.Worksheet("Template");
            XLWorkbook template2 = new XLWorkbook();
            ws.CopyTo(template2, "2");
}
```
The issue is with the template.xlsx content + CopyTo method call. There are 1000+ rows with data validation and formatting of cells. 
Template is narrowed down sample, that demonstrates the same issue we have with our prod template.
[Template.xlsx](https://github.com/ClosedXML/ClosedXML/files/2589139/Template.xlsx)

zabulus:
Do you need any additional info?

igitur:
I'm focusing on the release of 0.94 for now. Will look at this later.

zabulus:
@igitur 
I've debugged a little to find the difference between current develop and 0.87, found out the difference here:
XLWorksheet.CopyTo method.
0.87
```
public IXLWorksheet CopyTo(XLWorkbook workbook, String newSheetName, Int32 position)
        {
            var targetSheet = (XLWorksheet)workbook.WorksheetsInternal.Add(newSheetName, position);
            Internals.ColumnsCollection.ForEach(kp => targetSheet.Internals.ColumnsCollection.Add(kp.Key, new XLColumn(kp.Value)));
            Internals.RowsCollection.ForEach(kp => targetSheet.Internals.RowsCollection.Add(kp.Key, new XLRow(kp.Value))); // <----- Here
            Internals.CellsCollection.GetCells().ForEach(c => targetSheet.Cell(c.Address).CopyFrom(c, false));
```
0.94:
```
public IXLWorksheet CopyTo(XLWorkbook workbook, String newSheetName, Int32 position)
        {
            var targetSheet = (XLWorksheet)workbook.WorksheetsInternal.Add(newSheetName, position);
[...]

PR #1093, opened 2018-12-04
Poor performance with data validation #1071

zabulus:
Fixes this:
https://github.com/ClosedXML/ClosedXML/issues/1071


zabulus:
Test failure because of different order inside xlworksheet content. Comparer of worksheet parts should not rely on particular order inside xml file.

PR #1098, opened 2018-12-11
WIP Fix for adjusting formulas while consolidation conditional formats

b0bi79:
Fix for adjusting formulas while consolidation conditional formats.
This function worked fine in v0.92 (regression).
Fix is needed for migration ClosedXML.Report to last version of ClosedXML.

### DIFF
--- a/ClosedXML/Excel/ConditionalFormats/XLConditionalFormats.cs
+++ b/ClosedXML/Excel/ConditionalFormats/XLConditionalFormats.cs
@@ -94,7 +94,10 @@ namespace ClosedXML.Excel
                 item.Ranges.RemoveAll();
                 consRanges.ForEach(r => item.Ranges.Add(r));
 
-                var targetCell = item.Ranges.First().FirstCell() as XLCell;
+                var targetCell = item.Ranges
+                    .OrderBy(r => r.RangeAddress.FirstAddress.RowNumber)
+                    .ThenBy(r => r.RangeAddress.FirstAddress.ColumnNumber)
+                    .First().FirstCell() as XLCell;
                 (item as XLConditionalFormat).AdjustFormulas(baseCell, targetCell);
 
                 _conditionalFormats.Add(item);


b0bi79:
WIP

vbjay:
If you want to mark it WIP, change the title. Add WIP to the title.

b0bi79:
@igitur @Pankraty  I cannot write a test that reproduces this issue. ClosedXML.Report has tests that do not pass, but this fix allows tests to pass.
Maybe you have an idea why this could happen? The problem is that in some cases during the consolidation process item.Ranges may not be sorted, but in synthetic tests I cannot reproduce it.

igitur:
Which ClosedXML.Report tests don't pass?

b0bi79:
There are several such tests. They do not pass on ClosedXML 0.94.2.
One of these tests is GroupTagTests.Simple with the tLists1_sort.xlsx template.
But only I have not yet committed the code for version 0.94.2, if you want, I can commit.

igitur:
Yes, please commit ona branch and tell me

b0bi79:
https://github.com/ClosedXML/ClosedXML.Report/pull/66

Pankraty:
As I can say from code, consolidated ranges are constructed in the determined order: from top to bottom and from left to right in the scope of a single worksheet, and then follows the next worksheet, etc. The changes you propose do basically the same but ignore the worksheets. Can this be a reason for the issue? I mean, perhaps you had conditional formats extending to multiple worksheets so their behavior was affected?

b0bi79:
This needs to be checked. These formats were copied from one sheet to another.

b0bi79:
Here is the value of item.Ranges
{Sheet1!B15:B17,Sheet1!E15:G17,Sheet1!B12:B13,Sheet1!E12:G13,Sheet1!B19:B22,Sheet1!E19:G22}

Pankraty:
Well, I think I found it. After `XLRangeIndex` was introduced `XLRanges` enumerator started to return ranges in non-deterministic order (due to using a HashSet internally). Here is the reproduction:
```
        [Test]
        public void XLRangesPreserveOrder()
        {
            var wb = new XLWorkbook();

            var originalList = new List<IXLRange>();
            originalList.Add(ws.Range("F12:F12"));
            originalList.Add(ws.Range("F13:F13"));
            originalList.Add(ws.Range("F15:F15"));
            originalList.Add(ws.Range("F16:F16"));
            originalList.Add(ws.Range("F17:F17"));
            originalList.Add(ws.Range("F19:F19"));
            originalList.Add(ws.Range("F20:F20"));
            originalList.Add(ws.Range("F21:F21"));
            originalList.Add(ws.Range("F22:F22"));

            originalList.Add(ws.Range("B12:B12"));
            originalList.Add(ws.Range("B13:B13"));
            originalList.Add(ws.Range("B15:B15"));
            originalList.Add(ws.Range("B16:B16"));
            originalList.Add(ws.Range("B17:B17"));
            originalList.Add(ws.Range("B19:B19"));
            originalList.Add(ws.Range("B20:B20"));
            originalList.Add(ws.Range("B21:B21"));
            originalList.Add(ws.Range("B22:B22"));

            originalList.Add(ws.Range("E12:E12"));
            originalList.Add(ws.Range("E13:E13"));
            originalList.Add(ws.Range("E15:E15"));
            originalList.Add(ws.Range("E16:E16"));
            originalList.Add(ws.Range("E17:E17"));
            originalList.Add(ws.Range("E19:E19"));
            originalList.Add(ws.Range("E20:E20"));
            originalList.Add(ws.Range("E21:E21"));
            originalList.Add(ws.Range("E22:E22"));

            originalList.Add(ws.Range("G12:G12"));
            originalList.Add(ws.Range("G13:G13"));
            originalList.Add(ws.Range("G15:G15"));
            originalList.Add(ws.Range("G16:G16"));
            originalList.Add(ws.Range("G17:G17"));
            originalList.Add(ws.Range("G19:G19"));
            originalList.Add(ws.Range("G20:G20"));
            originalList.Add(ws.Range("G21:G21"));
            originalList.Add(ws.Range("G22:G22"));

            var ranges = new XLRanges();
            originalList.ForEach(r => ranges.Add(r));

            var actualList = ranges.ToList();

            Assert.AreEqual(originalList.Count, actualList.Count);
            for (int i = 0; i < actualList.Count; i++)
            {
                Assert.AreEqual(originalList[i], actualList[i]);
            }
        }
```

I'm affraid we cannot guarantee the original order of items in the `XLRanges` (we will loose all performance gain) but I also think we should not preserve it. Better return ranges ordered by their first address, similar to what @b0bi79 has proposed. 
@igitur do you agree?

Pankraty:
Alexey, #1133 should fix your problem. Could you check it?

b0bi79:
Thank you Alexey. This fixes the issue.

PR #1109, opened 2018-12-18
Don't automatically adjust columns width to contents when inserting datatable. Some users want to opt out because of PlatformNotSupportedException

### LINKED ISSUE #1107: System.Drawing is not supported on this platform.
- [x] Bug

**Version of ClosedXML**

e.g. 0.94.0

I can't create Excel file in my real project. The same code that works in a Console app or a simple Web App doesn't work in my app. This is the log:

```

System.Drawing is not supported on this platform.

   at ClosedXML.Excel.FontBaseExtensions.GetCachedFont(IXLFontBase fontBase, Dictionary`2 fontCache)
   at ClosedXML.Excel.FontBaseExtensions.GetWidth(IXLFontBase fontBase, String text, Dictionary`2 fontCache)
   at ClosedXML.Excel.XLColumn.AdjustToContents(Int32 startRow, Int32 endRow, Double minWidth, Double maxWidth)
   at ClosedXML.Excel.XLColumns.<>c__DisplayClass10_0.<AdjustToContents>b__0(XLColumn c)
   at ClosedXML.Excel.XLColumns.AdjustToContents(Int32 startRow, Int32 endRow)
   at ClosedXML.Excel.XLWorksheets.Add(DataTable dataTable, String sheetName)
```

It should work, or it should give me more data on what the underlying problem is:

Yes but I don't have access to old code and don't remember the previous version.

## Reproducibility

```c#
public class ExcelController : Controller
    {
        [HttpGet]
        public IActionResult Test()
        {
            var table = new DataTable();
            table.Columns.Add("Name", typeof(string));
            table.Columns.Add("Age", typeof(int));
            var row = table.NewRow();
            row["Name"] = "First Name";
            row["Age"] = 10;
            table.Rows.Add(row);
            workbook.Worksheets.Add(table, "Export");
            var result = new MemoryStream();
            workbook.SaveAs(result);
            result.Position = 0;
            var bytes = result.ToArray();
            var fileName = "Export.xlsx";
            return File(new MemoryStream(bytes), "application/octet-stream", fileName);
        }
    }
```

This code works just find in an isolated newly created Web App. But it throws that error in my real microservices, and I can't find out why.

igitur:
Fixes #1107

### DIFF
--- a/ClosedXML/Excel/XLWorksheets.cs
+++ b/ClosedXML/Excel/XLWorksheets.cs
@@ -161,7 +161,6 @@ namespace ClosedXML.Excel
         {
             var ws = Add(sheetName);
             ws.Cell(1, 1).InsertTable(dataTable);
-            ws.Columns().AdjustToContents(1, 75);
             return ws;
         }
 
--- a/ClosedXML_Examples/Misc/AddingDataSet.cs
+++ b/ClosedXML_Examples/Misc/AddingDataSet.cs
@@ -1,6 +1,7 @@
 using ClosedXML.Excel;
 using System;
 using System.Data;
+using System.Linq;
 
 namespace ClosedXML_Examples.Misc
 {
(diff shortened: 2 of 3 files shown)


nefcanto:
No, this doesn't solve the problem.
The real problem is, the exact code snippet works fine in one project, and throws this exception in another project, ON THE SAME PLATFORM.
Please consider that.

PR #1125, opened 2019-01-07
Change cached value type when data type changes 

### LINKED ISSUE #1124: AdjustToContents doesn't work correctly in a cell containing date difference

- [x] Bug
- [ ] Feature

**Version of ClosedXML**
0.94.2

When a cell is showing a date difference and has a date format, the method `AdjustToContents` doesn't work correctly in it. It sets the cell wider than it needs to be.

`AdjustToContents` should set the correct minimum width.

No.

## Reproducibility
```c#
public void Main()
{
  {
    var ws = wb.AddWorksheet("Test");
    ws.Cell(1, 1).Value = new DateTime(2019, 1, 1, 14, 0, 0);
    ws.Cell(1, 2).Value = new DateTime(2019, 1, 1, 17, 48, 0);
    ws.Cell(1, 3).FormulaA1 = "=B1-A1";
    ws.Columns(1, 2).Style.DateFormat.Format = "hh:mm";
    ws.Column(3).Style.DateFormat.Format = "[hh]:mm";
    ws.ColumnsUsed().AdjustToContents();
    wb.SaveAs("test.xlsx");
  }
}
```

I believe this is caused by a mismatch between the actual cell value and the cell format. The format of the cell is is `[hh]:mm`, the displayed value is `03:48` (as in "three hours and 48 minutes") and the internal value is a `double` representing the difference in days (`0.158333333398332`). When executing `AdjustToContents`, the actual string is determined by `ExcelNumberFormat.NumberFormat.Format` method (see https://github.com/ClosedXML/ClosedXML/blob/124a714a1da9818880cac125245cc9cd71594251/ClosedXML/Extensions/FormatExtensions.cs#L14). But that method expects a `DateTime` value, not a `double` one. So it throws an `InvalidCastException` internally and returns a simple `ToString` value which is quite long given the decimal representation of the value above. Therefore the cell ends up way wider than it needs to be.

igitur:
Fixes #1124 

Results of operations on Dates can return either a double or a date (or possibly timespan). Cached values should be updated to reflect the type when `IXLCell.DataType` changes.

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -503,7 +503,10 @@ namespace ClosedXML.Excel
             if (force || NeedsRecalculation)
             {
                 if (HasFormula)
+                {
                     CachedValue = RecalculateFormula(FormulaA1);
+                    UpdateCachedValueFromDataType();
+                }
                 else
                     CachedValue = null;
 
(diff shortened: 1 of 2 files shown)


crypto-rsa:
Great, thanks!

PR #895, opened 2018-06-04
Pivot formatting adding/loading/saving

b0bi79:
#### What's this PR do?
PR adds formatting (styles) to the pivot table.

#### How should this be manually tested?
Run test XLPivotTableTests.PivotFieldFormatsTest.

#### Screenshots (if appropriate)
2018-06-04 10-39-17 microsoft excel - pvt xlsx

PR #1129, opened 2019-01-10
Alternative implementation of pivot table formats

igitur:
@b0bi79 @Pankraty 

Here is my alternative implementation. It removes the need for lambda functions and "builders". I think this is a cleaner API, althought the internals are a bit more messy than your implementation. Some of the new enum's could be better named. We can finalise that later.

I made some changes to e.g. `XLPivotValues` that should probably be submitted in a separate PR.

What do you think of this? Look at the unit test to see how to use it.

Closes #895

### DIFF
--- a/ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotFieldStyleFormats.cs
+++ b/ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotFieldStyleFormats.cs
@@ -1,0 +1,82 @@
+﻿// Keep this file CodeMaid organised and cleaned
+
+namespace ClosedXML.Excel
+{
+    internal class XLPivotFieldStyleFormats : IXLPivotFieldStyleFormats
+    {
+        private IXLPivotValueStyleFormat dataValuesFormat;
+        private IXLPivotStyleFormat headerFormat;
+        private IXLPivotStyleFormat labelFormat;
+        private IXLPivotStyleFormat subtotalFormat;
+
+        public XLPivotFieldStyleFormats(IXLPivotField field)
+        {
+            this.PivotField = field;
+        }
+
+        public IXLPivotField PivotField { get; }
+
+        #region IXLPivotFieldStyleFormats
+
+        public IXLPivotValueStyleFormat DataValuesFormat
+        {
+            get
+            {
+                if (dataValuesFormat == null)
+                {
+                    dataValuesFormat = new XLPivotValueStyleFormat(PivotField)
+                    {
+                        AppliesTo = XLPivotStyleFormatElement.Data
+                    };
+                }
+                return dataValuesFormat;
+            }
+            set { dataValuesFormat = value; }
+        }
+
+        public IXLPivotStyleFormat Header
+        {
+            get
+            {
+                if (headerFormat == null)
+                {
+                    headerFormat = new XLPivotStyleFormat(PivotField);
+                }
+                return headerFormat;
+            }
+            set { headerFormat = value; }
+        }
+
+        public IXLPivotStyleFormat Label
+        {
+            get
+            {
+                if (labelFormat == null)
+                {
+                    labelFormat = new XLPivotStyleFormat(PivotField)
+                    {
+                        AppliesTo = XLPivotStyleFormatElement.Label
+                    };
+                }
+                return labelFormat;
+            }
+            set { labelFormat = value; }
+        }
+
+        public IXLPivotStyleFormat Subtotal
+        {
+            get
+            {
+                if (subtotalFormat == null)
+                {
+                    subtotalFormat = new XLPivotStyleFormat(PivotField);
+                }
+
+                return subtotalFormat;
+            }
+            set { subtotalFormat = value; }
+        }
+
+        #endregion IXLPivotFieldStyleFormats
+    }
+}
--- a/ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotStyleFormat.cs
+++ b/ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotStyleFormat.cs
@@ -1,0 +1,27 @@
+// Keep this file CodeMaid organised and cleaned
+using System.Collections.Generic;
+
+namespace ClosedXML.Excel
+{
+    internal class XLPivotStyleFormat : IXLPivotStyleFormat
+    {
+        public XLPivotStyleFormat(IXLPivotField field = null, IXLStyle style = null)
+        {
+            PivotField = field;
+            Style = style ?? XLStyle.Default;
+        }
+
+        #region IXLPivotStyleFormat members
+
+        public XLPivotStyleFormatElement AppliesTo { get; set; } = XLPivotStyleFormatElement.Data;
+        public IXLPivotField PivotField { get; set; }
+        public IXLStyle Style { get; set; }
+
+        #endregion IXLPivotStyleFormat members
+
+        internal XLPivotAreaValues AreaType { get; set; } = XLPivotAreaValues.Normal;
+        internal bool CollapsedLevelsAreSubtotals { get; set; } = false;
+        internal IList<AbstractPivotFieldReference> FieldReferences { get; } = new List<AbstractPivotFieldReference>();
+        internal bool Outline { get; set; } = true;
+    }
+}
--- a/ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotStyleFormatEnums.cs
+++ b/ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotStyleFormatEnums.cs
@@ -1,0 +1,24 @@
+﻿using System;
+
+namespace ClosedXML.Excel
+{
+    [Flags]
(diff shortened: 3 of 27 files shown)


b0bi79:
I like it

b0bi79:
You mean templates for ClosedXML.Report? So far, there are only two templates for Pivot. When I can use it in ClosedXML.Report, I will add more tests.

For reverse-engineer I studied how any format changes are reflected in the file and after that I looked in the specification what these elements and attributes mean.

igitur:
This PR is finally ready for review. I admit we need more test cases, but let's start with this for now.

I don't think splitting it up adds any value. The tricky part is to figure out how the field references work.

ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotFieldStyleFormats.cs
```diff
@@ -0,0 +1,62 @@
...
+            this.PivotField = field;
+        }
+
+        public IXLPivotField PivotField { get; }
+
+        #region IXLPivotFieldStyleFormats
+
+        public IXLPivotValueStyleFormat DataValuesFormat
+        {
+            get { return dataValuesFormat ?? (dataValuesFormat = new XLPivotValueStyleFormat(PivotField) { AppliesTo = XLPivotStyleFormatElement.Data }); }
```
Pankraty: A bit too much for one line. Better use for `DataValuesFormat` and for `Header` the same pattern as for `Label` and `Subtotal`.

ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotStyleFormat.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public XLPivotStyleFormat(IXLPivotField field)
+            : this(field, XLStyle.Default)
+        { }
+
+        public XLPivotStyleFormat(IXLStyle style)
+            : this(null, style)
+        { }
+
+        public XLPivotStyleFormat(IXLPivotField field, IXLStyle style)
```
Pankraty: Out of 4 public constructors, one is never used and the other is used just once. Given that the class is internal, what do you think about decreasing the number of constructors? See also `XLPivotValueStyleFormat`.

Pankraty: Hos about this?

```
public XLPivotStyleFormat(IXLPivotField field = null, IXLStyle style = XLStyleDefault)
{
  PivotField = field;
  Style = style ?? XLStyle.Default;
}
```

ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotStyleFormatEnums.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+
+namespace ClosedXML.Excel
+{
+    [Flags]
+    public enum XLPivotStyleFormatElement
+    {
+        None = 0,
+        Label = 1,
+        Data = 2,
```
Pankraty:
```
None  = 1 << 0,
Label = 1 << 1,
Data  = 1 << 2,
```
This way chances someone will accidentally add the next flag with value 3 instead of 4 are lower.

igitur: Yes, except `None = 0` should remain. `1 << 0 == 1`

ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotStyleFormat.cs
igitur: Habit of mine that when I create the class I create all possible constructors I think would be used.

ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotStyleFormatEnums.cs
Pankraty: Ah, of course

igitur:
@Pankraty Ok for merging?

b0bi79:
Francois, I am now trying to refine the styles for pivot tables and more and more I realize that you shouldn’t throw out the `IEnumerable <IFieldRef> FieldReferences {get; }` from XLPivotFormat. In files that are created in Excel there can be a variety of styles in the pivotArea. Now the LoadPivotStyleFormats method is very limited in this respect and most of the styles are simply ignored. I suggest to return all the same to FieldReferences, which was in the PR proposed by me.

igitur:
It's a bit late to totally revert this. I need to spend some time again to understand the problem. Do you have any idea how the other libraries treat this, or at least what their API is? Do other libraries even have support for these pivot formats?

b0bi79:
But we can leave the existing api and expand it.

igitur:
Yes, of course.

b0bi79:
EPPlus does not have this feature. POI does everything at a very low level - you need to directly create a PivotArea object, create references objects ....
[DevExpress implementation](https://documentation.devexpress.com/OfficeFileAPI/118506/Spreadsheet-Document-API/Examples/Pivot-Tables/How-to-Apply-a-Custom-Style-to-a-Pivot-Table) does not suit us.

b0bi79:
Spire.Xls contains class [PivotTableStyle](https://www.e-iceblue.com/api_documents/53f40e627eb265-36565818/res/html/a31dad30-3c4b-11ab-84c3-9b38144fd765.htm) with property Styles: Dictionary<PivotTableElement, PivotStyle>

b0bi79:
Spire.Xls usage example 
```
            string styleName = "newone";
            pt.CustomTableStyleName = styleName;
            PivotTableStyle pivotTableStyle = new PivotTableStyle(styleName);
            PivotStyle pivotStyle = workbook.CreatePivotStyle();
            pivotStyle.Font.Color = Color.Blue;
            pivotTableStyle.Styles.Add(PivotTableElement.headerRow, pivotStyle);
            workbook.AddPivotTableStyle(pivotTableStyle);
```
Here styles are added not for the columns, but for the table.

b0bi79:
Francois, I would like to finish working with Pivot styles. Can I change the API that you developed? This API cannot implement all the features of pivot styles. The biggest difficulties arise with files generated by native Excel.

igitur:
Can you post an example of how you see the API should be, before actually starting to code the implementation? I would like to avoid lambda functions as far as possible.

b0bi79:
I have not thought through to the end what the API should be. But I think that there should be a part of a simple user API and a part of a "low-level" API with lambda functions. Later I will give you an example of a simple user API.

igitur:
OK, that sounds reasonable.

b0bi79:
A simple user API for creating pivot styles:
```
pt.AddRowGrandTotalFormats(XLPivotStyleFormatElement.All).Style.Font.FontColor = XLColor.VenetianRed;

monthPivotField.AddLabelStyle().Style.Fill.BackgroundColor = XLColor.Amber;
monthPivotField.AddHeaderStyle().Style.Font.FontColor = XLColor.Yellow;
namePivotField.AddSubtotalStyle(XLPivotStyleFormatElement.Label).Style.Fill.BackgroundColor = XLColor.Blue;
namePivotField.AddSubtotalStyle(XLPivotStyleFormatElement.Data).Style.Fill.BackgroundColor = XLColor.DarkBlue;
namePivotField.AddDataValuesStyle()
    .AndWith(monthPivotField, v => v.ToString() == "May")
    .ForValueField(numberOfOrdersPivotValue)
    .Style.Font.FontColor = XLColor.Green;

```
Extended API:
```
// styles creating
pt.Styles.Add(b => { b.ForSubtotal(row).Style.Fill.BackgroundColor = XLColor.Blue; });
pt.Styles.Add(b => { b.ForLabel(clmn).Style.Fill.BackgroundColor = XLColor.Amber; });
pt.Styles.Add(b => { b.ForHeader(clmn).Style.Font.FontColor = XLColor.Yellow; });
pt.Styles.Add(b => { b.ForGrandRow().Style.Font.FontColor = XLColor.VenetianRed; });
pt.Styles.Add(b => {
    b.ForData(row)
    .AndWith(clmn, v=>v.ToString() == "May")
    .ForValueField(numberOfOrders)
    .Style.Font.FontColor = XLColor.Green;
});

// styles getting
Assert.AreEqual("Month", pt.Styles[0].FieldName);
Assert.AreEqual(true, pt.Styles[0].DataOnly);
Assert.AreEqual(PivotAreaValues.Button, pt.Styles[0].AreaType);
```
We can make simple API methods (AddLabelStyle(), AddHeaderStyle(), AddSubtotalStyle(), AddDataValuesStyle()) as extension methods or new methods for IXLPivotField. How do you think this is best done?

b0bi79:
In fact, your API remains almost the same. The difference is that the properties are replaced by methods. We are forced to do this because it is difficult to parse correctly all the options `pivotArea` and `references`. Therefore, I propose to present for reading styles in a raw, and add new styles through a simple API.
Also, if you want, we can hide the API of builders with lambda functions in the low-level API.

PR #1135, opened 2019-01-16
Correctly load column-wide styles

### LINKED ISSUE #961: Column.NumberFormat is not loaded correctly

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.93

When a `NumberFormat` is specified for a column, it's not present in the resulting Excel file, but a different number format is set.
For example, this code:
```C#
var sourceColumn = sourceWorkbook.Worksheets.Single().Column(1);
                sourceColumn.Style.NumberFormat.Format = "0,00";
```
Produces this format when viewed in Excel: `#.000`. It's an empty string when opened again with ClosedXML.

It worked in versions earlier than 0.93.

## Reproducibility

The following test case shows the problem:

```c#
[Fact]
public void ReadAndWriteColumnNumberFormat()
{
    using (var sourceWorkbook = new XLWorkbook())
    {
        sourceWorkbook.AddWorksheet("MyCalculation");
        var sourceColumn = sourceWorkbook.Worksheets.Single().Column(1);
        sourceColumn.Style.NumberFormat.Format = "0,00";

        using (var memoryStream = new MemoryStream())
        {
            sourceWorkbook.SaveAs(memoryStream);
            memoryStream.Position = 0;

            using (var destWorkbook = new XLWorkbook(memoryStream))
            {
                var column = destWorkbook.Worksheets.Single().Column(1);
                // This test fails, the read value is an empty string
                Assert.Equal("0,00", column.Style.NumberFormat.Format);
            }
        }
    }
}
```

Here's the defined number formats in the resulting `styles.xml`:

```xml
<x:numFmts count="2">
  <x:numFmt numFmtId="0" formatCode="" />
  <x:numFmt numFmtId="165" formatCode="0,00" />
</x:numFmts>
```

So the `numFmtId` is set to `165` and present, but it's not referenced in `sheet1.xml`:

```xml
<x:cols>
  <x:col min="1" max="1" width="9.140625" style="1" customWidth="1" />
</x:cols>
```

I'm not exactly sure what happens. I think the reference should be `style="165"`. It looks like the reference to style `1` gets picked up by excel as some internal, pre-defined style (numbers with thousands separator, in this case).

Pankraty:
This PR fixes #961.


ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -2907,7 +2907,11 @@ private void SetProperties(SpreadsheetDocument dSpreadsheet)
                 xlStyle.NumberFormat = xlNumberFormat;
             }
 
-            xlStylized.InnerStyle = new XLStyle(xlStylized, xlStyle);
+            //When loading columns we must propagate style to each column but not deeper. In other cases we do not propagate at all.
+            if (xlStylized is IXLColumns columns)
+                columns.Cast<XLColumn>().ForEach(col => col.InnerStyle = new XLStyle(col, xlStyle));
+            else
+                xlStylized.InnerStyle = new XLStyle(xlStylized, xlStyle);
```
igitur: Does this logic not hold also for `IXLRows` and may other similar types?

Pankraty: Column and row styles are stored differently in Excel. Say, if we have a range of columns with the same style they will be loaded simultaneously (thus we'll get `XLColumns` here):
image
Rows are always stored separately (at least, I could not manage to generate a file where they are combined) so we will not get `XLRows`:
image

Probably we can still add similar logic for rows as well - just in case...

Issue #1030, opened 2018-10-13
Clearing cells on merging


Pankraty:
The continuation of #1029 

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

If you are logging a possible bug or feature request, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.93

**What is the current behavior?**

When cells are merged they style remain the same and contents is not cleared. When the file is opened in Excel it looks fine, but when cells are unmerged their styles change to the original instead of remaining the same.

**What is the expected behavior or new feature?**

When cells are merged their contents, comments, data validation rules, and conditional formats must be cleared (except the first one). The style of merged cells must become equal to the style of the first cell in the range. When the merged range in unmerged cells appearance must be the same as it was in the merged state. The most tricky part is to preserve borders.

**Did this work in previous versions of our tool?  Which versions?**

Don't know

## Reproducibility

```
        [Test]
public void UnmergedCellsPreserveStyle()
{
    using (XLWorkbook wb = new XLWorkbook())
    {
        var ws = wb.AddWorksheet("Sheet1");
        var range = ws.Range("B2:D4");
        range.Style.Fill.SetBackgroundColor(XLColor.Yellow);
        range.Style.Border.SetOutsideBorder(XLBorderStyleValues.Thick)
            .Border.SetOutsideBorderColor(XLColor.DarkBlue)
            .Border.SetInsideBorder(XLBorderStyleValues.Thin)
            .Border.SetInsideBorderColor(XLColor.Pink);
        range.Cells().ForEach(c => c.Value = c.Address.ToString());

        var firstCell = ws.Cell("B2");
        firstCell.Style.Fill.SetBackgroundColor(XLColor.Red)
            .Alignment.SetHorizontal(XLAlignmentHorizontalValues.Center)
            .Font.SetBold();

        range.Merge();
        range.Unmerge();

        Assert.IsTrue(range.Cells().All(c => c.Style.Fill.BackgroundColor == XLColor.Red));
        Assert.IsTrue(range.Cells().Where(c => c != firstCell).All(c => c.Value == ""));
        Assert.AreEqual("B2", firstCell.Value);

        Assert.AreEqual(XLBorderStyleValues.Thick, ws.Cell("B2").Style.Border.TopBorder);
[...]

PR #1137, opened 2019-01-17
Fix changing inner borders of the range

### LINKED ISSUE #1111: Border Style changes without changing
[border Problem.zip](https://github.com/ClosedXML/ClosedXML/files/2700579/border.Problem.zip)

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

latest

the border changes after these commands:

ws1.Cell(21, 1).Value = "a string..."
ws1.Cell(25, 2).Value = "a string..."
ws1.Cell(26, 2).Value = "a string..."
ws1.Cell(31, 1).Value = beschreibung;

just changes the values of these 4 cells, but not the border

Pankraty:
This PR fixes #1111.

In fact, the problem was made visible by #1030, though it had existed for a long time already: changing `InsideBorder` to `None` would change the outside border style and reset its color. The style was being restored after the changes, but not the color.


igitur:
selfie-2
[_GitHub Selfies_](https://github.com/thieman/github-selfies/)

PR #1153, opened 2019-02-27
Correctly throw DivisionByZeroException when denominator is zero.

### LINKED ISSUE #1152: Div by zero in formula results in invalid Excel file

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.94.2

When on a sheet you save via ClosedXML there is a div/0 error and you require that formulae be evaluated, then Excel 2016 16.0.4639.1000 complains upon opening the file:
"We found a problem with some content in <file.xlsx>. Do you want to try to recover as much as we can?"
It is not important that you create the file using ClosedXML. Same thing happens when you create the file with a div/0 error in Excel, then just open and immediately save with no change via ClosedXML.

May be related to #826 ?

I expect a file to be saved, which Excel considers valid.

Didn't test

## Reproducibility
```c#
        {
            var fileName = "divByZeroInvalidFile.xlsx";
            var worksheet = workbook.Worksheets.Add("Sample Sheet");
            worksheet.Cell("A1").SetFormulaA1("0/0");
            workbook.SaveAs(fileName, evaluateFormulae: true, validate: true);
            Process.Start(fileName);
        }
```

igitur:
Fixes #1152

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Expression.cs
+++ b/ClosedXML/Excel/CalcEngine/Expression.cs
@@ -354,12 +354,21 @@ namespace ClosedXML.Excel.CalcEngine
                     return (double)_lft * (double)_rgt;
 
                 case TKID.DIV:
+                    if (Math.Abs((double)_rgt) < double.Epsilon)
+                        throw new DivisionByZeroException();
+
                     return (double)_lft / (double)_rgt;
 
                 case TKID.DIVINT:
+                    if (Math.Abs((double)_rgt) < double.Epsilon)
+                        throw new DivisionByZeroException();
+
                     return (double)(int)((double)_lft / (double)_rgt);
 
                 case TKID.MOD:
+                    if (Math.Abs((double)_rgt) < double.Epsilon)
+                        throw new DivisionByZeroException();
+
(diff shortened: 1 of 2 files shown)


ClosedXML/Excel/CalcEngine/Expression.cs
```diff
@@ -354,12 +354,21 @@ override public object Evaluate()
                     return (double)_lft * (double)_rgt;
 
                 case TKID.DIV:
+                    if (Math.Abs((double)_rgt) < double.Epsilon)
```
Pankraty: Perhaps we should introduce `XLHelper.Epsilon` of approximately `2.2250738585073E-308`, because `double.Epsilon` is 16 decimal orders lower (4.9somethingE **-324**) then Excel supports.

igitur: So you're saying the ClosedXML calculation engine has a higher precision than Excel? If that's true, I wouldn't want to change it. It's a useful feature, not a bug.

Pankraty: Let it be so

PR #1155, opened 2019-03-06
Ignore _X (capital X) as XML escaping char

### LINKED ISSUE #1154: Reading _Xceed_ results in 컭
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.94.2

When having for example "_Xceed_Something" filled out in cell A1 of my excel, the resulting string read from this cell will be "컭Something". 

The resulting string should be "_Xceed_Something", as this is just a string value.

I don't think so, but untested

## Reproducibility

When running the code below, the result will be "컭Something" instead of "_Xceed_Something".

```c#
public void Main()
{
   var stream = new FileStream(@"path\to\your\XceedTest.xlsx", FileMode.Open);
   var workbook = new XLWorkbook(stream);
   var sheet = workbook.Worksheets.First();

    var cell = sheet.RowsUsed().First().Cell("A");
    Console.WriteLine(cell.GetString());
}
```
[XceedTest.xlsx](https://github.com/ClosedXML/ClosedXML/files/2935000/XceedTest.xlsx)

igitur:
Remove own implementation of `IsXmlChar`
Case-insensitive decoding of hex strings
Ignore _X (capital X) as XML escaping char for decoding. Fixes #1154

### DIFF
--- a/ClosedXML/Utils/XmlEncoder.cs
+++ b/ClosedXML/Utils/XmlEncoder.cs
@@ -8,20 +8,6 @@ namespace ClosedXML.Utils
     {
         private static readonly Regex xHHHHRegex = new Regex("_(x[\\dA-F]{4})_", RegexOptions.Compiled);
 
-        /// <summary>
-        /// Checks if a character is not allowed to the XML Spec http://www.w3.org/TR/REC-xml/#charsets
-        /// </summary>
-        /// <param name="ch">Input Character</param>
-        /// <returns>Returns false if the character is invalid according to the XML specification, and will not be
-        /// escaped by an XmlWriter.</returns>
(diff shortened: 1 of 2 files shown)


Pankraty:
Do I understand correctly that lowercase `_x` works fine?

igitur:
> Do I understand correctly that lowercase _x works fine?

Short answer, yes. 

Long answer: Escaping/encoding unsupported characters will generate a string e.g. _x0001_ (represents Unicode character 1). Note the lowercase `_x`. Decoding/unescaping works correctly in recognising the lowercase `_x`.

This change is to avoid recognising `_X` as an escape character too.

PR #1159, opened 2019-03-19
Always treat "A1:B2" range strings as absolute

igitur:
Previously, as commented in https://github.com/ClosedXML/ClosedXML/pull/1159/files#diff-03a6bc67d4fe134e5578c84c9fc4e29fL52 :

```c#
ws.Range("B2:F10").Range("B2").Address.ToString()
```
would return `C3` because the `B2` was treated as relative to the first range.

This PR breaks this and ensures that range strings are always treated as absolute, i.e. relative to the entire worksheet.

Range strings like `"1-3, 4, 7"` are still treated as relative.

The first few commits of this PR does some reorganising of methods: `Intersection()` is consolidated and `Relative` is moved to `IXLRangeBase` and made public.

### DIFF
--- a/ClosedXML/Excel/Ranges/XLRangeAddress.cs
+++ b/ClosedXML/Excel/Ranges/XLRangeAddress.cs
@@ -114,6 +114,9 @@ namespace ClosedXML.Excel
 
         #region Public methods
 
+        public Boolean IsNormalized => LastAddress.RowNumber >= FirstAddress.RowNumber
+                                       && LastAddress.ColumnNumber >= FirstAddress.ColumnNumber;
+
         /// <summary>
         /// Lead a range address to a normal form - when <see cref="FirstAddress"/> points to the top-left address and
         /// <see cref="LastAddress"/> points to the bottom-right address.
--- a/ClosedXML/Excel/Ranges/XLRangeAddress.cs
+++ b/ClosedXML/Excel/Ranges/XLRangeAddress.cs
@@ -367,6 +367,17 @@ namespace ClosedXML.Excel
             return IsValid && IsEntireColumn() && IsEntireRow();
         }
 
+        public IXLRange AsRange()
+        {
+            if (this.Worksheet == null)
+                throw new InvalidOperationException("The worksheet of the current range address has not been set.");
+
+            if (!this.IsValid)
+                return null;
+
+            return this.Worksheet.Range(this);
+        }
+
         #endregion Public methods
 
         #region Operators
--- a/ClosedXML/Excel/Ranges/XLRangeAddress.cs
+++ b/ClosedXML/Excel/Ranges/XLRangeAddress.cs
@@ -24,6 +24,11 @@ namespace ClosedXML.Excel
                 new XLAddress(worksheet, row, XLHelper.MaxColumnNumber, false, false));
         }
 
+        public static readonly XLRangeAddress Invalid = new XLRangeAddress(
+            new XLAddress(-1, -1, fixedRow: true, fixedColumn: true),
+            new XLAddress(-1, -1, fixedRow: true, fixedColumn: true)
+        );
+
         #endregion Static members
 
         #region Constructor
--- a/ClosedXML/Excel/Ranges/XLRangeAddress.cs
+++ b/ClosedXML/Excel/Ranges/XLRangeAddress.cs
@@ -372,6 +372,38 @@ namespace ClosedXML.Excel
             return IsValid && IsEntireColumn() && IsEntireRow();
         }
 
+        public IXLRangeAddress Intersection(IXLRangeAddress otherRangeAddress)
+        {
+            if (otherRangeAddress == null)
+                throw new ArgumentNullException(nameof(otherRangeAddress));
+
+            var xlOtherRangeAddress = (XLRangeAddress)otherRangeAddress;
+            return Intersection(in xlOtherRangeAddress);
+        }
+
+        internal XLRangeAddress Intersection(in XLRangeAddress otherRangeAddress)
+        {
+            if (!this.Worksheet.Equals(otherRangeAddress.Worksheet))
+                throw new ArgumentOutOfRangeException(nameof(otherRangeAddress), "The other range address is on a different worksheet");
+
+            var thisRangeAddressNormalized = this.Normalize();
+            var otherRangeAddressNormalized = otherRangeAddress.Normalize();
+
+            var firstRow = Math.Max(thisRangeAddressNormalized.FirstAddress.RowNumber, otherRangeAddressNormalized.FirstAddress.RowNumber);
+            var firstColumn = Math.Max(thisRangeAddressNormalized.FirstAddress.ColumnNumber, otherRangeAddressNormalized.FirstAddress.ColumnNumber);
+            var lastRow = Math.Min(thisRangeAddressNormalized.LastAddress.RowNumber, otherRangeAddressNormalized.LastAddress.RowNumber);
+            var lastColumn = Math.Min(thisRangeAddressNormalized.LastAddress.ColumnNumber, otherRangeAddressNormalized.LastAddress.ColumnNumber);
+
+            if (lastRow < firstRow || lastColumn < firstColumn)
+                return XLRangeAddress.Invalid;
+
+            return new XLRangeAddress
+            (
+                new XLAddress(this.Worksheet, firstRow, firstColumn, fixedRow: false, fixedColumn: false),
+                new XLAddress(this.Worksheet, lastRow, lastColumn, fixedRow: false, fixedColumn: false)
+            );
+        }
+
         public IXLRange AsRange()
         {
             if (this.Worksheet == null)
--- a/ClosedXML/Excel/Ranges/XLRangeAddress.cs
+++ b/ClosedXML/Excel/Ranges/XLRangeAddress.cs
@@ -372,6 +372,39 @@ namespace ClosedXML.Excel
             return IsValid && IsEntireColumn() && IsEntireRow();
         }
 
+        public IXLRangeAddress Relative(IXLRangeAddress sourceRangeAddress, IXLRangeAddress targetRangeAddress)
+        {
+            var xlSourceRangeAddress = (XLRangeAddress)sourceRangeAddress;
+            var xlTargetRangeAddress = (XLRangeAddress)targetRangeAddress;
+
+            return Relative(in xlSourceRangeAddress, in xlTargetRangeAddress);
+        }
+
+        internal XLRangeAddress Relative(in XLRangeAddress sourceRangeAddress, in XLRangeAddress targetRangeAddress)
+        {
+            var sheet = targetRangeAddress.Worksheet;
+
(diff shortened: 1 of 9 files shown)


ClosedXML/Excel/Ranges/XLRangeAddress.cs
```diff
@@ -106,7 +106,10 @@ IXLAddress IXLRangeAddress.LastAddress
             get { return LastAddress; }
         }
 
-        public bool IsValid => FirstAddress.IsValid && LastAddress.IsValid;
+        public bool IsValid => FirstAddress.IsValid
+                               && LastAddress.IsValid
+                               && FirstAddress.RowNumber <= LastAddress.RowNumber
+                               && FirstAddress.ColumnNumber <= LastAddress.ColumnNumber;
```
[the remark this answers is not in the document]
igitur: When calculating an intersection in the new optimised case, for the `Range()` parameters in https://github.com/ClosedXML/ClosedXML/pull/1159/files#diff-aba9a994b9ee071c751c3455950c7891R2257 to work correctly, the `otherRange` parameter has to be a normalised range. Else the `Math.Max()` and `Math.Min()` will not return the expected row and column numbers.

I suppose I can revert this change and add additional `Math.Min()` and `Math.Min()`'s to do a quick and dirty normalization inside `Intersection()`.

igitur:
@Pankraty Pushed some changes that addresses your one comment. Will rebase later to make commit history cleaner, if you agree with the changes.

ClosedXML_Tests/Excel/Ranges/XLRangeBaseTests.cs
```diff
@@ -463,5 +463,12 @@ public void XLRangesReturnsRangesInDeterministicOrder()
                 Assert.AreEqual(expectedRanges[i], actualRanges[i]);
             }
         }
+
+        [Test]
+        public void RelativeRanges()
+        {
+            var ws = new XLWorkbook().AddWorksheet();
+            Assert.AreEqual("E13:F13", ws.Range("D4:E4").Relative(ws.Range("A1:E4"), ws.Range("B10:F14")).RangeAddress.ToString());
```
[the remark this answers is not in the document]
igitur: If I remember correctly, `XLRange`s cannot be invalid, but `XLRangeAddress`es can be invalid. (right?) So in this case we have a few options:
- return `null`
- throw `ArgumentOutOfRangeException`
- throw new custom ClosedXML exception
- change the method to return `IXLRangeAddress`, which can then be invalid.

ClosedXML/Excel/Ranges/XLRangeBase.cs
```diff
@@ -2249,17 +2249,39 @@ public IXLRangeBase Intersection(IXLRangeBase otherRange, Func<IXLCell, Boolean>
...
+                (
+                    Math.Min(this.RangeAddress.FirstAddress.RowNumber, this.RangeAddress.LastAddress.RowNumber),
+                    Math.Min(otherRange.RangeAddress.FirstAddress.RowNumber, otherRange.RangeAddress.LastAddress.RowNumber)
+                );
+
+                firstColumn = Math.Max
+                (
+                    Math.Min(this.RangeAddress.FirstAddress.ColumnNumber, this.RangeAddress.LastAddress.ColumnNumber),
+                    Math.Min(otherRange.RangeAddress.FirstAddress.ColumnNumber, otherRange.RangeAddress.LastAddress.ColumnNumber)
+                );
```
Pankraty:
```
var thisRangeAddress = this.RangeAddress.Normalize();
var otherRangeAddress = otherRange.RangeAddress().Normalize();
var firstRow = Math.Max(thisRangeAddress.FirstAddress.RowNumber, otherRangeAddress.FirstAddress.RowNumber);
var firstColumn = Math.Max(thisRangeAddress.FirstAddress.ColumnNumber, otherRangeAddress.FirstAddress.ColumnNumber);
var lastRow = Math.Min(thisRangeAddress.LastAddress.RowNumber, otherRangeAddress.FirstAddress.RowNumber);
var lastColumn = Math.Min(thisRangeAddress.LastAddress.ColumnNumber, otherRangeAddress.FirstAddress.ColumnNumber);

if (lastRow < firstRow || lastColumn < firstColumn) return null;

return this.Worksheet.Range(firstRow, firstColumn, lastRow, lastColumn);
```

And we don't need `IsNormilized()` method, I suppose.

ClosedXML_Tests/Excel/Ranges/XLRangeBaseTests.cs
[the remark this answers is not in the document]
Pankraty: You are right about `XLRange`s cannot be invalid.

From all these options I prefer the last one. This way it is easier to reuse this method, say, for shifting formulas, etc. Also if one only needs to get the relative address we will not allocate a new `XLRange` for this.

ClosedXML/Excel/Ranges/XLRangeBase.cs
igitur: I kind of like the idea of keeping `IsNormalized` around. But I will improve its calculation based on your code.

Pankraty: No problem.

ClosedXML_Tests/Excel/Ranges/XLRangeBaseTests.cs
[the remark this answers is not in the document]
igitur: `.Relative()` now returns `IXLRangeAddress`.

igitur:
@Pankraty Completely rewrote the commit history. More commits, but better separated.

ClosedXML/Excel/Ranges/XLRangeAddress.cs
```diff
@@ -367,6 +367,35 @@ public bool IsEntireSheet()
...
+
+            var thisRangeAddressNormalized = this.Normalize();
+            var otherRangeAddressNormalized = otherRangeAddress.Normalize();
+
+            var firstRow = Math.Max(thisRangeAddressNormalized.FirstAddress.RowNumber, otherRangeAddressNormalized.FirstAddress.RowNumber);
+            var firstColumn = Math.Max(thisRangeAddressNormalized.FirstAddress.ColumnNumber, otherRangeAddressNormalized.FirstAddress.ColumnNumber);
+            var lastRow = Math.Min(thisRangeAddressNormalized.LastAddress.RowNumber, otherRangeAddressNormalized.LastAddress.RowNumber);
+            var lastColumn = Math.Min(thisRangeAddressNormalized.LastAddress.ColumnNumber, otherRangeAddressNormalized.LastAddress.ColumnNumber);
+
+            return new XLRangeAddress
```
Pankraty: You missed the validation 
```
if (lastRow < firstRow || lastColumn < firstColumn)
```
This should return an invalid address.

Consider the example:
```
var intersection = ws.Range("A1:C3").Intersection(ws.Range("A5:C8"));
Assert.IsFalse(intersection.IsValid);
```
Now the `intersection` is `A5:C3` which is wrong.

igitur: Ok, I'll have to create an arbitrary invalid range address then. I'll just set all row and column numbers to -1. Agree?

igitur: This is now done after the latest push. Added a static field `XLRangeAddress.Invalid` that we can reuse whenever we want to deliberately return an invalid range address.

Pankraty: This is exactly what I was going to suggest :)

ClosedXML/Excel/Ranges/XLRangeAddress.cs
```diff
@@ -24,6 +24,11 @@ public static XLRangeAddress EntireRow(XLWorksheet worksheet, int row)
                 new XLAddress(worksheet, row, XLHelper.MaxColumnNumber, false, false));
         }
 
+        public static XLRangeAddress Invalid = new XLRangeAddress(
```
Pankraty: That's great, but the field should not be editable. So it should either be `readonly`, or be a property without a setter, or be an expression `public static XLRangeAddress Invalid => new XLRangeAddress(...`.

Pankraty: And I guess we have to do something with the equality of invalid addresses. Comparing two invalid range addresses must always return the same result: either true or false. I am not sure which has more sense but it should definitely be consistent. Now the result may be any. Shall I open a new issue for this?

igitur: This is an internal struct field, so I wasn't too catious, but I've made it read-only anyway.

igitur: > And I guess we have to do something with the equality of invalid addresses. Comparing two invalid range addresses must always return the same result: either true or false. I am not sure which has more sense but it should definitely be consistent. Now the result may be any. Shall I open a new issue for this?

It's OK, I'll fix this quickly.

igitur: I guess we should do the same (new static field and equality operator) for `XLAddress` too?

Pankraty: Agree. I even thought we already have this but it seems we don't

igitur: So we agree that 2 invalid addresses/range addresses are equal if both are invalid (regardless of their actual row/column numbers)?

Do you think 2 invalid addresses are equal is they are both invalid, but on different worksheets? I would argue not equal.

igitur: Also, I see our current implementation of `XLAddress.Equals()` implies that `Sheet1!A1` is equal to `Sheet2!A1`. Worksheet is ignored. Maybe that should change? I can see valid reasons either way.

Pankraty: Addresses on different worksheets are not equal, to my opinion

Pankraty: > Also, I see our current implementation of XLAddress.Equals() implies that Sheet1!A1 is equal to Sheet2!A1. Worksheet is ignored. Maybe that should change?

I don't remember why this was needed to compare addresses from different worksheets, but for me, consider worksheets during comparison also seems reasonable. We also have a method `WithoutWorksheet` that can be utilized to compare addresses regardless of their worksheets - perhaps, we could do it public.

igitur: Ok. I think we must make `XLRangeAddress.IsValid` and `XLAddress.IsValid` public too.

igitur: Maybe this should be a new issue after all, as you suggested. Are you OK with the rest of this PR as is?

Pankraty: Yes, it's fine

PR #1161, opened 2019-03-20
 Avoid creation of new item in Internals.RowsCollection when adding comment

### LINKED ISSUE #1160: Error when try to delete row
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.94.2

If add comment to the cell, then update column width, it throws System.ArgumentException when you try to delete row above the commented cell

System.ArgumentException is not caused

## Reproducibility

```c#
public void Main()
{
    using (var book = new XLWorkbook())
    {
        var ws = book.Worksheets.Add("test");
        ws.Cell(4, 1).Comment.AddText("test");
        ws.Column(1).Width = 100;
        ws.Row(1).Delete();
     }
}
```

igitur:
In the `XLComment` constructor, the previous row's height is being looked up. If the row doesn't exist, we should avoid its creation in `RowsCollection`.

Also refactored `XLRangeBase.Delete` to make it more readable.

Fixes #1160

### DIFF
--- a/ClosedXML/Excel/Ranges/XLRangeBase.cs
+++ b/ClosedXML/Excel/Ranges/XLRangeBase.cs
@@ -1543,11 +1543,15 @@ namespace ClosedXML.Excel
                 RangeAddress.LastAddress.RowNumber,
                 RangeAddress.LastAddress.ColumnNumber);
 
-            foreach (
-                XLCell cell in
-                    Worksheet.Workbook.Worksheets.Cast<XLWorksheet>().SelectMany(
-                        xlWorksheet => (xlWorksheet).Internals.CellsCollection.GetCells(
-                            c => !String.IsNullOrWhiteSpace(c.FormulaA1))))
+            // Shift formulas first
+            foreach (var cell in Worksheet
+                .Workbook
+                .Worksheets
+                .Cast<XLWorksheet>()
+                .SelectMany(ws => ws
+                    .Internals
+                    .CellsCollection
+                    .GetCells(c => c.HasFormula)))
             {
                 if (shiftDeleteCells == XLShiftDeletedCells.ShiftCellsUp)
                     cell.ShiftFormulaRows((XLRange)shiftedRangeFormula, numberOfRows * -1);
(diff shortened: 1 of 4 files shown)


ClosedXML_Tests/Excel/Rows/RowTests.cs
```diff
@@ -263,5 +259,14 @@ public void NegativeRowNumberIsInvalid()
...
             Assert.IsFalse(row.RangeAddress.IsValid);
         }
+
+        [Test]
+        public void DeleteRowOnWorksheetWithComment()
+        {
+            var ws = new XLWorkbook().AddWorksheet();
+            ws.Cell(4, 1).Comment.AddText("test");
+            ws.Column(1).Width = 100;
+            ws.Row(1).Delete();
```
Pankraty: The test does not have as `Assert`. If it just aims to check that the code does not fail I'd prefer to make it explicit. E.g. `Assert.DoesNotThrow(() => ws.Row(1).Delete());`.

ClosedXML/Excel/Ranges/XLRangeBase.cs
```diff
@@ -1557,39 +1561,64 @@ public void Delete(XLShiftDeletedCells shiftDeleteCells)
...
+                                           c.Address.ColumnNumber - columnModifier,
+                                           fixedRow: false,
+                                           fixedColumn: false);
 
-                if (canInsert)
-                    cellsToInsert.Add(newKey, newCell);
+                if (newCellAddress.IsValid)
+                {
+                    var newCell = new XLCell(Worksheet, newCellAddress, c.StyleValue);
+                    newCell.CopyValuesFrom(c);
```
Pankraty: Never thought about how cells are removed during ranges shifting in ClosedXML.

Say, if the user has a reference to the cell `B2` and then removes the column `A`, he'll still have a reference to the cell which a) is not presented on the worksheet, b) has an address `B2`, c) has a value same as the new `A2` cell. And even though this cell is removed from a worksheet I am almost sure that manipulating with it will affect the actual worksheet in an unpredictable manner. This seems wrong to me. Besides, this causes an unnecessary allocation of `XLCells`, so I see the possibility for improvement here.

For now, let's leave it as it is, your refactoring made the code easier to read indeed. We can make it better some time later.

ClosedXML_Tests/Excel/Rows/RowTests.cs
igitur: Of course. I forget that NUnit has this assertion.

ClosedXML/Excel/Ranges/XLRangeBase.cs
igitur: Not sure I follow you - maybe you can illustrate with an example. But what happens currently, references to cells doesn't necessarily cause a new entry in `XLWorksheet.Internals.CellsCollection`. Only if the reference is evaluated in some way, e.g. a value is requested, or if the target cell has a style applied (or other conditions) does an entry get added. 

When column `A` is removed, cell `B2` will be removed from `CellsCollection`. A new entry with address `A2` will be created and the old `B2`'s properties will be copied to `A2`. See the `newCell.CopyValuesFrom(c);` line. So cells here are never shifted. They are explicitly removed and readded with a new address. I do see, however, how this is not optimal in terms of performance, though.

Pankraty: Created #1162 to address this.

PR #1164, opened 2019-03-26
Address equality

igitur:
Consider addresses `Sheet1:A1` and `Sheet2!A1` and then we delete sheets `Sheet1` and `Sheet2`. Should the 2 addresses be treated as equal?

### DIFF
--- a/ClosedXML/Excel/Coordinates/IXLAddress.cs
+++ b/ClosedXML/Excel/Coordinates/IXLAddress.cs
@@ -1,3 +1,4 @@
+// Keep this file CodeMaid organised and cleaned
 using System;
 using System.Collections.Generic;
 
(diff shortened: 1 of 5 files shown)


Pankraty:
> Consider addresses Sheet1:A1 and Sheet2!A1 and then we delete sheets Sheet1 and Sheet2. Should the 2 addresses be treated as equal?

Yes, I think. Consider two formulas: `=Sheet1!A1` and `=Sheet2!A1`.
If we delete column A on both sheets formulas will become `=Sheet1!#REF` and `=Sheet2!#REF` - thus we can treat such addresses as different. If we remove both sheets then, the formulas will both become `=#REF!#REF`, and I think we can safely assume that addresses are equal as well. Does it look logical for you?

Pankraty (approved):

igitur:
I started this almost 2 years ago! I can't even remember anymore why exactly I wanted to implement this.

jahav:
Closing as obsolete.

It's six years old, the original reason for change unclear, plus I am a big proponent of always-valid model and added `IsValid` property of `IXLAddress` makes me unhappy (though yes, it can happen). I would much rather replace the class altogether with struct + return null on invalid (yeah, not going to happen).

The added `[DebuggerStepThrough]` is not useful anymore, there is an option in Visual Studio to set that.

PR #1166, opened 2019-03-28
Ignore default Excel function namespace and implement CONCAT function

### LINKED ISSUE #1165: Request support for CONCAT() function

- [] Bug
- [x] Feature

**Version of ClosedXML**

0.94 & develop branch

concat() is not supported

concat() is the replacement for concatenate() in Excel per Microsoft: https://support.office.com/en-us/article/concat-function-9b1a9a3f-94ff-41af-9736-694cbd6b4ca2
It accepts the same parameters and has the same functionality.

No

## Reproducibility

Attempt to evaluate formula using CONCAT()

igitur:
Fixes #1165

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Functions/Text.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/Text.cs
@@ -1,6 +1,7 @@
 using ClosedXML.Excel.CalcEngine.Exceptions;
 using ExcelNumberFormat;
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
@@ -18,6 +19,7 @@ namespace ClosedXML.Excel.CalcEngine
             ce.RegisterFunction("CHAR", 1, _Char); // Returns the character specified by the code number
             ce.RegisterFunction("CLEAN", 1, Clean); //	Removes all nonprintable characters from text
             ce.RegisterFunction("CODE", 1, Code); // Returns a numeric code for the first character in a text string
+            ce.RegisterFunction("CONCAT", 1, int.MaxValue, Concat); //	Joins several text items into one text item
             ce.RegisterFunction("CONCATENATE", 1, int.MaxValue, Concatenate); //	Joins several text items into one text item
             ce.RegisterFunction("DOLLAR", 1, 2, Dollar); // Converts a number to text, using the $ (dollar) currency format
             ce.RegisterFunction("EXACT", 2, Exact); // Checks to see if two text values are identical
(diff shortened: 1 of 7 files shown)


ClosedXML/Excel/CalcEngine/Functions/Text.cs
```diff
@@ -60,6 +62,22 @@ private static object Code(List<Expression> p)
...
+            {
+                if (x is IEnumerable enumerable)
+                {
+                    foreach (var i in enumerable)
+                        sb.Append((string)(new Expression(i)));
+                }
+                else
+                    sb.Append((string)x);
+            }
+            return sb.ToString();
```
Pankraty: > If the resulting string exceeds 32767 characters (cell limit), CONCAT returns the #VALUE! error.

Don't we have to follow the spec?

ClosedXML_Tests/Excel/CalcEngine/TextTests.cs
```diff
@@ -73,6 +73,26 @@ public void Code_Value()
...
+
+            var ws = new XLWorkbook().AddWorksheet();
+
+            ws.FirstCell().SetValue(20)
+                .CellBelow().SetValue("AB")
+                .CellBelow().SetFormulaA1("=DATE(2019,1,1)")
+                .CellBelow().SetFormulaA1("=CONCAT(A1:A3)");
+
+            actual = ws.Cell("A4").Value;
+            Assert.AreEqual("20AB43466", actual);
```
Pankraty: Could you please add a similar test case for Concatenate function with the expected result "20"? It works this way, but I think it would be nice to have it covered with a test.

igitur: OK, but that exact same test for `CONCATENATE` would and should throw an exception.

ClosedXML/Excel/CalcEngine/Functions/Text.cs
igitur: Erm, I'm in 2 minds about this... what does LibreOffice do?

ClosedXML_Tests/Excel/CalcEngine/TextTests.cs
Pankraty: Huh. I misinterpreted Excel's behavior. Turned out, it treats such expressions like array functions:
image
(all B-cells have the same formula here)

ClosedXML/Excel/CalcEngine/Functions/Text.cs
igitur: In LibreOffice Calc I just did a `CONCAT()` that return 64000 characters. So I would say this decision is implementation-specific.

ClosedXML_Tests/Excel/CalcEngine/TextTests.cs
igitur: Added test for this.

PR #1134, opened 2019-01-15
Correctly propagate worksheet style to cells from a worksheet

Pankraty:
As @b0bi79 has discovered, ClosedXML does not assign the worksheet style to empty cells that belong to columns or rows which had been initialized before the worksheet style was set.

```
        [TestCase("A1", TestName = "First cell")] // OK
        [TestCase("A2", TestName = "Cell from initialized row")]  // Fails
        [TestCase("B1", TestName = "Cell from initialized column")]  // Fails
        [TestCase("D4", TestName = "Initialized cell")]  // OK
        [TestCase("F6", TestName = "Non-initialized cell")]  // OK
        public void CellTakesWorksheetStyle(string cellAddress)
        {
            using (var wb = new XLWorkbook())
            {
                var ws = wb.AddWorksheet("Sheet1");
                ws.Column(2);
                ws.Row(2);
                ws.Cell("D4").Value = "Non empty";
                ws.Style.Font.SetFontName("Arial");
                ws.Style.Font.SetFontSize(9);

                var cell = ws.Cell(cellAddress);
                Assert.AreEqual("Arial", cell.Style.Font.FontName);
                Assert.AreEqual(9, cell.Style.Font.FontSize);
            }
        }
```

PR #1168, opened 2019-03-30
Fix initialization of default styles for XLCells, XLRows, XLColumns and XLRanges

Pankraty:
#1134 did not fix the issue with the default cell styles completely. Many ClosedXML.Report tests were failing because the worksheet in the template had a non-standard style (Arial 8pt, to be specific) and some of the empty cells were initialized with the default value (Calibri 11 pt). In some cases, this affected the heights of the rows so it could not be just ignored.


Pankraty:
cc @b0bi79

ClosedXML/Excel/Rows/XLRows.cs
```diff
@@ -11,7 +11,15 @@ internal class XLRows : XLStylizedBase, IXLRows, IXLStylized
...
-        public XLRows(XLWorksheet worksheet) : base(XLWorkbook.DefaultStyleValue)
+
+        /// <summary>
+        /// Create a new instance of <see cref="XLRows"/>.
+        /// </summary>
+        /// <param name="worksheet">If worksheet is specified it means that the created instance represents
+        /// all rows on a worksheet so changing its height will affect all rows.</param>
+        /// <param name="defaultStyle">Default style to use when initializing child entries.</param>
+        public XLRows(XLWorksheet worksheet, XLStyleValue defaultStyle)
+            : base(defaultStyle ?? XLWorkbook.DefaultStyleValue)
```
igitur: Instead of falling back to `XLWorkbook.DefaultStyleValue` here, should we not modify `XLStylizedBase`'s constructor to be
```c#
        protected XLStylizedBase(XLStyleValue styleValue)
        {
            StyleValue = styleValue ?? XLWorkbook.DefaultStyleValue;
        }
```

ClosedXML/Excel/XLWorkbook.cs
```diff
@@ -640,7 +640,7 @@ public IXLCells FindCells(Func<IXLCell, Boolean> predicate)
 
         public IXLRows FindRows(Func<IXLRow, Boolean> predicate)
         {
-            var rows = new XLRows(null);
+            var rows = new XLRows(null, null);
```
igitur: Please use named parameters in these cases. I realise it was like that already, but since we're touching lines, we might as well fix it.

ClosedXML/Excel/Rows/XLRows.cs
igitur: And if so, maybe `defaultStyle` should be optional with default of `null`?

ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -1850,13 +1850,13 @@ private void LoadFont(OpenXmlElement fontSource, IXLFontBase fontBase)
             if (row.CustomFormat != null)
             {
                 Int32 styleIndex = row.StyleIndex != null ? Int32.Parse(row.StyleIndex.InnerText) : -1;
-                if (styleIndex > 0)
+                if (styleIndex >= 0)
```
igitur: Why `>=`? As far as I know, `0` is reserved for the default style.

Pankraty: There were problems with many ClosedXML.Report tests because of this. Style 0 is the default style of a workbook but it does not necessarily equal to `XLStyle.Default` and it is stored in the file just like any other style.

See the example:
image

ClosedXML/Excel/Rows/XLRows.cs
Pankraty: Good point.

PR #1172, opened 2019-04-04
Clear all RelId's if workbook is loaded from template.

igitur:
If a workbook is loaded with `XLWorkbook.OpenFromTemplate()`, it should be treated as brand new, i.e. no RelId's should be prepopulated.


Pankraty:
Why not 0.95?

igitur:
Ok, I guess we can squeeze it in.

PR #1174, opened 2019-04-04
Correctly dispose temporary memorystream

igitur:


Pankraty:
Did you have issues with this? Because as far as I remember `MemoryStream.Dispose()` method does nothing.

igitur:
No, no issues. I'm just used to closing all streams properly.

PR #1178, opened 2019-04-09
Fixes in GetHashCode functions (1073)

### LINKED ISSUE #1073: Extreme performance hit in several GetHashCode implementations
- [x] Bug
- [ ] Feature

**Version of ClosedXML**
0.93.1

Saving the workbook takes longer than on previous versions. Profiler tells us that 50% of the time is spent in `GetHashCode()` where most of the time is spent in `ToUpperInvariant()` (causing a subsequent GC):

https://github.com/ClosedXML/ClosedXML/blob/develop/ClosedXML/Excel/Style/XLFontKey.cs#L63
https://github.com/ClosedXML/ClosedXML/blob/develop/ClosedXML/Excel/Style/XLNumberFormatKey.cs#L15

**There are a few more places.** I think all `GetHashCode()` implementations needs an overhaul. In general, you must **never** allocate heap objects in GetHashCode! This is the perfect performance killer (especially if your application has high memory pressure).

It worked a bit better in 0.90.0. (But there we had crazy delays in IsEmpty() method instead.)

Pankraty:
This PR fixes the bug in `XLNumberFormatKey` equality check: the comparison was case-insensitive while it should be case-sensitive as, for example, `mm` stands for minutes while `MM` - for months.

Besides, it improves `XLFontKey.GetHashCode` in order to reduce memory allocations.

Closes #1073

### DIFF
--- a/ClosedXML/Excel/Style/XLFontKey.cs
+++ b/ClosedXML/Excel/Style/XLFontKey.cs
@@ -60,7 +60,7 @@ namespace ClosedXML.Excel
             hashCode = hashCode * -1521134295 + Shadow.GetHashCode();
             hashCode = hashCode * -1521134295 + FontSize.GetHashCode();
             hashCode = hashCode * -1521134295 + FontColor.GetHashCode();
-            hashCode = hashCode * -1521134295 + FontName.ToUpperInvariant().GetHashCode();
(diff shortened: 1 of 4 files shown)


igitur:
How much of a performance gain is there?

Pankraty:
Not much, though it may depend on a testing scenario. But the bug is nasty.

znakeeye:
Actually quite much. The performance gain will be significant when under heavy memory pressure.

PR #1179, opened 2019-04-09
Enhancement - Added Color as a filter. 

pranav700:
#### Questions:
- Is there a blog post?
- Does the knowledge base need an update? -Yes


igitur:
I noticed that you did all your work on your `develop` branch. It's better to branch of a new branch with a distinct name. That way, if you want to work on multiple pull requests, you have that option.

So far this PR looks good, but I notice that it doesn't take conditional formatting into account, and to do that would anyway be quite difficult. Isn't the use case of colour autofilters indeed to be combined with with conditional formats? Or, in your use case, do you actually want to filter on manually set background colours?

pranav700:
Thanks for the review.
Yes, there is no conditional formatting involved in this use case. This implementation is just for filtering manually set background colors.

igitur:
Ok, good work so far. You still need to implement loading and saving of the color autofilter to and from the file.

pranav700:
If I am getting it correctly, then I think saving has been done in ClosedXML_Examples/AutoFilters/ColouredAutoFilter.cs and loading was done inside   ClosedXML_Tests/Examples/AutoFilterTests.cs

pranav700:
@igitur - Did you get a chance to look into my comments?

Pankraty:
@pranav700 could you please mark this checkbox in this PR:
image

I will push some formatting changes and then it will be ready to be included in 0.96.

pranav700:
@Pankraty - Done, thanks

PR #1183, opened 2019-04-15
Improve Merge operation performance

Pankraty:
As @b0bi79 has discovered, `XLRangeBase.Merge()` is not quite efficient: `cellsUsed` is enumerated multiple times.

In addition, `LastCellUsed()` method is called in `CellRangeReference` twice so this part was also improved.

In my testing example, the execution time changes from 17 sec to 11 sec (>130K random merges).


igitur:
Nice find.

PR #1185, opened 2019-04-17
Avoid auto creation of cell hyperlink

### LINKED ISSUE #1184: Better treatment of half-baked properties (eg RichText, Hyperlink etc)

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

**0.94.2**

in Powershell 5.1, Windows 10 x64

If I enter a cell object in PowerShell command line to inspect its properties, the cell will turn into a hyperlink cell, and prevent the workbook to save. Save the workbook cause error below:
```
Exception calling "SaveAs" with "1" argument(s): "Value cannot be null.
Parameter name: source”
At line:4 char: 1
+ $wb.SaveAs("Desktop\Test.xlsx")
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : ArgumentNullException
```
This is the InnerException stacktrace (from $error[0].Exception.InnerException.Stacktrace):
```
   at System.Linq.Enumerable.Contains[TSource](IEnumerable`1 source, TSource value, IEqualityComparer`1 comparer)
   at ClosedXML.Excel.XLHyperlink.get_InternalAddress()
   at ClosedXML.Excel.XLWorkbook.GenerateWorksheetPartContent(WorksheetPart worksheetPart, XLWorksheet xlWorksheet, Boolean evaluateFormulae, SaveContext context)
   at ClosedXML.Excel.XLWorkbook.CreateParts(SpreadsheetDocument document, SaveOptions options)
   at ClosedXML.Excel.XLWorkbook.CreatePackage(String filePath, SpreadsheetDocumentType spreadsheetDocumentType, SaveOptions options)
   at ClosedXML.Excel.XLWorkbook.SaveAs(String file, SaveOptions options)
   at CallSite.Target(Closure , CallSite , Object , String )
```

Should save with no problem.

same problem in 0.93.1

## Reproducibility

```powershell
add-type -Path .nuget\packages\fastmember\1.3.0\lib\net45\FastMember.dll
add-type -path .nuget\packages\excelnumberformat\1.0.3\lib\net20\ExcelNumberFormat.dll
add-type -path .nuget\packages\documentformat.openxml\2.7.2\lib\net46\DocumentFormat.OpenXml.dll
add-type -path .nuget\packages\closedxml\0.94.2\lib\net46\ClosedXML.dll

$wb = New-Object ClosedXML.Excel.XLWorkbook
$sheet = $wb.AddWorksheet("Test")
$sheet.Cell("A1")  # This line will cause the following line throw error
$wb.SaveAs("Desktop\Test.xlsx")
$sheet.Cell("A1").Hyperlink = $null  # This line will somehow fix the error
$wb.SaveAs("Desktop\Test.xlsx")  # It's ok this time.
```

smilinger:
#### What's this PR do?
Fix #1184 
Test it in vs or powershell
#### Questions:
- Is there a blog post? 
No
No
No

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -1270,9 +1270,6 @@ namespace ClosedXML.Excel
         {
             get
             {
-                if (_hyperlink == null)
-                    Hyperlink = new XLHyperlink();
-
                 return _hyperlink;
             }
 
--- a/ClosedXML/Excel/Hyperlinks/XLHyperlink_Internal.cs
+++ b/ClosedXML/Excel/Hyperlinks/XLHyperlink_Internal.cs
@@ -4,11 +4,6 @@ namespace ClosedXML.Excel
 {
     public partial class XLHyperlink
     {
-        internal XLHyperlink()
-        {
-
-        }
-
         internal XLHyperlink(XLHyperlink hyperlink)
         {
             _externalAddress = hyperlink._externalAddress;
--- a/ClosedXML/Excel/Hyperlinks/XLHyperlink_public.cs
+++ b/ClosedXML/Excel/Hyperlinks/XLHyperlink_public.cs
@@ -72,6 +72,8 @@ namespace ClosedXML.Excel
             {
                 if (IsExternal)
                     return null;
+                if (_internalAddress == null)
+                    return null;
                 if (_internalAddress.Contains('!'))
                 {
                     return _internalAddress[0] != '\''


igitur:
Thanks for the PR. We appreciate all contributions. Notice that it breaks some of the unit tests, so please run the unit tests and ensure that all existing unit tests pass. You should also add your own unit test to test your own improvement. 

I'll add individual code comments in a review.

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -1270,9 +1270,6 @@ public XLHyperlink Hyperlink
         {
             get
             {
-                if (_hyperlink == null)
```
igitur: We definitely want to keep this. Removing this will break unit tests and people's code.

igitur (changes requested):

igitur:
I think the correct solution for this issue is to change a bit of the code in `XLWorkbook_Save.cs`. I think you should check whether a cell's hyperlink is valid (i.e. its internal address is not `null`) and only then generate the necessary XML.

@Pankraty Have you been following the issue ( #1184 )? I'm starting to think all of those "self-instantiating" properties should come with fail-safe mechanisms. Maybe a private boolean field to indicate whether it truly has been fully "instantiated" with subproperties. Understand what I mean?

smilinger:
Yes, you are right. I realized that right after those failed tests, then I had a look at the save functions and decided it's a little complicated for me to modify it, and like you said it's not only about hyperlinks now, I will close my PR and leave it to you guys.


-------- 原始邮件 --------
主题：Re: [ClosedXML/ClosedXML] Avoid auto creation of cell hyperlink (#1185)
发件人：Francois Botha
收件人：ClosedXML/ClosedXML
抄送：smilinger ,Author


I think the correct solution for this issue is to change a bit of the code in XLWorkbook_Save.cs. I think you should check whether a cell's hyperlink is valid (i.e. its internal address is not null) and only then generate the necessary XML.

@Pankraty<https://github.com/Pankraty> Have you been following the issue ( #1184<https://github.com/ClosedXML/ClosedXML/issues/1184> )? I'm starting to think all of those "self-instantiating" properties should come with fail-safe mechanisms. Maybe a private boolean field to indicate whether it truly has been fully "instantiated" with subproperties. Understand what I mean?

―

Pankraty:
@igitur, this makes a perfect sense. I can't check the code right now but I got your point and agree with you.

igitur:
Properties that need private fields to indicate full initialisation:
- Hyperlink
- Comment
- RichText
- DataValidation

igitur:
Fixed in #1396

Issue #1187, opened 2019-04-17
ClosedXML AddConditionalFormat() Adds Quotes To Value In Excel (French)


CaptBli:
## Read and complete the full issue template

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

If you are logging a possible bug or feature request, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.94.2

**What is the current behavior?**

I am using conditional formatting and it works great with my U.S. settings. When ClosedXML is used on European PCs, the **conditional formatting** when using numbers in the translated spreadsheets works great for Italian and German, but in French the conditional formatting is incorrect.  The numbers get converted to strings which when opened in Excel doesn't work.

The AddConditionalFormat().WhenEqualOrLessThan() methods do not work correctly when the locale is French. It adds quotes to the conditional statement. When I open it in Excel the Rule looks like:

Cell Value <= "-98.00002"

It is supposed to be:

Cell Value <= -98.00002

In the case of French localization, the extraneous quotes are added, but not the the Italian and German cases.

Following is my actual code:

```
ws.Range(r, 1, r, 2).Style.Alignment.Horizontal = XLAlignmentHorizontalValues.General; 

__ws.Cell(r, 2).AddConditionalFormat().WhenEqualOrLessThan(-98.00002f). Font.SetFontColor(XLColor.LightGray);__

ws.SheetView.FreezeRows(2);
ws.Columns().AdjustToContents();
```

We posted this problem a long time ago and thought it was fixed.  It is definitely in the current builds.

**What is the expected behavior or new feature?**

We expect that the Conditional Format to be: Cell Value <= -98.00002

**Did this work in previous versions of our tool?  Which versions?**
The old code: 

```
private String GetQuoted(XLFormula formula)
{
String value = formula.Value;
Double num;
if ((!Double.TryParse(value, out num) && !formula.IsFormula) && value[0] != '\"' && !value.EndsWith("\""))
return String.Format("\"{0}\"", value.Replace("\"", "\"\""));

return value;
}
```

We made a modification to the code to make this work in the past.  The following code will fix the issue in older code. 

```
private String GetQuoted(XLFormula formula)
{
String value = formula.Value;
Double num;
[...]

PR #1190, opened 2019-04-18
Fix for issue #1187

TheonTech:
#### What's this PR do?
Fixes issue #1187 
line 34 of XLCFCellIsConverter.cs
-set OS Formats and Location to French (France) and France respectively.
-using ClosedXML, create a spreadsheet that creates a conditional format using a number. 
   Specific example is: 
`ws.Cell(r,2).AddConditionalFormat().WhenEqualOrLessThan(-98.00002f).Font.SetFontColor(XLColor.LightGray);`
-observe that the conditional format is inserted with quotes and has other side effects (ref FrenchSpreadsheet.xlsx attached to issue #1187)
-apply this fix
-observe that the conditional format is now created correctly.

We actually discovered this bug about 5 years ago. At the time, we submitted a bug report but we also downloaded the source, fixed the bug, and have been using our custom version of ClosedXML ever since. I have recently been refactoring our shared libraries to .netstandard and decided to try using the updated version of ClosedXML from NuGet. However, our testing has revealed that this bug still exists which is why I am submitting this pull request.

#### Questions:
- Is there a blog post? No
- Does the knowledge base need an update? No
- Does this add new (C#) dependencies? No


igitur:
Thanks for the contribution so far. Can you olease add some unit tests? You can have a look at the other conditional format unit tests for a good starting point.

TheonTech:
I would like to do that but I am not sure how I would emulate the OS being set to French.
Also, I am currently in crunch mode for a May 10 release of our product so it will be after that before I can look into adding any unit test(s).

igitur:
There are several tests that first set the current culture before continuing. You can use ` Thread.CurrentThread.CurrentCulture = CultureInfo.CreateSpecificCulture("en-US");` to do this, but adjust for the required culture name. (French, you said?)

Pankraty:
Replaced with #1245

Issue #893, opened 2018-06-01
Formulas handling redesign


Pankraty:
After Styles and Ranges handling has been improved it's time to revise formulas.

**Goals:**

1. Decrease memory usage by formulas stored in `XLCell` s by implementing the formula sharing.
2. Improve the performance by reducing the amount of formula parsing work (#815).
3. Unify the approach to storing the formulas among different entities (cells, named ranges, conditional formats, etc.)
4. Unify the approach to storing formula computation results. Now implemented for cells only.
5. Unify the approach to formula shifting (#686, #880). This may be achieved if formulas refer to cells and ranges by reference rather than by addresses in string format. Thus, when the related range address changes the formula changes accordingly automatically; no need to shift them one-by-one separately for cells, conditional formats, named ranges, data validations, etc.
6. Support table column references, e.g. [Value] (#271)

**Affected entities**
1. `XLCell`
2. `XLConditionalFormat`
3. `XLNamedRange`
4. `XLDataValidation`
5. `XLTableField`
6. Charts - somewhere far in the future

**Basic concepts**
1. Different contexts require different formulas. E.g. for cell, formula result must be a single value, while for named range it can be either a single value, or a range reference, or a collection of range references. Technically, they may be implemented as different subclasses of a single class, or by using an `enum` reflecting the formula type.
2. For the purpose of sharing formulas between entities the new repository has to be introduced (`XLExpressionRepository`?). It will store expression trees without relation to concrete cells or ranges. For example, two cells having formulas `=SUM(A1:A2)` and `=SUM(D2:E18)` will internally refer to the same expression `SUM(*range reference nr. 1*)`; another cell with the formula `=SUM(A1:A2)+SUM(D2:E18)` will refer to another expression (`SUM(*range reference nr. 1*) + SUM(*range reference nr.2*)`) having no relation to the first one. *(Textual representation of the expressions serves to explanation only; they must be stored in a compiled form)*
3. Concrete instances that may use formulas (see **Affected entities**) hold separate instances of `XLFormula` (?). It includes
* A reference to the shared expression (`XLExpression` ?)
* A collection of concrete range references (`A1:A2` for the cell 1, `D2:E18` for the cell 2, `{A1:A2, D2:E18}` for the cell 3). Perhaps, the same collection must store cell references and table column references as well.
* Cached computation results, a stamp of the last computation, flag `NeedsRecalculation`, etc. - the existing staff implemented in `XLCell` for caching formula values.
[...]

PR #1201, opened 2019-05-08
[WIP] Implement formula caching

### LINKED ISSUE #882: FormulaWithCircularReferenceFails2 test fails
`XLCellTests.FormulaWithCircularReferenceFails2` was never marked with the `Test` attribute and was skipped. It also fails with a `FormulaWithCircularReferenceFails2`

Pankraty:
Relates to #893 

1. I created an `XLFormulaDefinition` class that contains chunks of formula and a collection of `IXLReference`-s. Those `IXLReference`-s can be easily formatted in A1 or R1C1 style, making the conversion of `XLFormulaDefinition` quite easy.
2. To store shared `XLFormulaDefinition`s I added a dedicated repository, defined on a workbook level.
3. To store a concrete formula for a concrete cell I added a class `XLCellFormula`. Right now it may seem unnecessary, but I plan to move the fields related to calculations there (such as `EvaluatedAtVersion`, `NeedsRecalculationEvaluatedAtVersion`, maybe `CachedValue`, etc.). This will make `XLCell` class a bit simpler and also will make cells _without_ formulas lighter.
4. I changed the regexes used for searching references in formulas (see e588b87) so the formula would not require to be placed between `><` chars before parsing. This alone reduced the number of allocations and GC cycles (on my test example - from 43 GC cycles to 29).
5. I found that previous implementation would not process RC references for a single row or column (`R[1]`, `C1`,...) correctly. In A1-style they are always defined as ranges (e.g. `A:A`), but it RC it is not always true. Now this is fixed.
6. Finally, I found that very long evaluation chains may cause `StackOverflowException`. To avoid this I perform an evaluation in a loop rather than via recursion (see e5ebc49). In fact, this commit is independent and may be merged separately, if you wish.

What I would also like to do:
1. To make `XLCalclEngine` working with `XLContextualFormula`. The calc engine would parse the formula and store the parsed and compiled `Expression` in `XLFormulaDefinition`. On evaluation, it would only provide the appropriate values from the corresponding `IXLReference`s. This has to reduce the time consumed by formula parsing greatly.
2. Modify `GetAffectingCells` method for it analyzed `IXLReference`s instead of parsing the `A1` formula. To make this possible I have to include a worksheet into `IXLReference`.
3. And, as described before, I plan to move certain properties of `XLCell` to `XLCellFormula`.

Probably these changes have to be made in future PRs.

In the meantime, @igitur you can start a review. I suggest checking separate commits - I think it is easier to understand the logic this way.

PS. Fixes #882 too.


ClosedXML/Excel/Formula/XLFormulaDefinitionParserBase.cs
```diff
@@ -0,0 +1,130 @@
...
+        /// <summary>
+        /// Split a formula into string parts (which do not change on formula shifting)
+        /// and a series of <see cref="IXLReference"/>-s.
+        /// </summary>
+        /// <param name="formula">A formula to parse.</param>
+        /// <param name="baseAddress">A base address used for parsing relative references.</param>
+        /// <returns>A tuple where Item1 is a collection of strings (is never empty) and
+        /// Item2 is a collection of references. Item2 may be empty and always contains one element
+        /// fewer than Item1, so they can be joined together in order to get the original formula this way:
+        /// Item1[0] + Item2[0] + Item1[1] + Item2[2] + ... + Item1[n] + Item2[n] + Item1[n+1]</returns>
```
Pankraty: ` + Item2[1]`, of course

ClosedXML/Excel/Formula/XLFormulaDefinition.cs
```diff
@@ -0,0 +1,77 @@
+﻿// Keep this file CodeMaid organised and cleaned
+using System.Text;
+
+namespace ClosedXML.Excel
+{
+    /// <summary>
+    /// Formula defined in a context-independent way. For example, is cell A1 contains formula '=B1'
```
Pankraty: `...if cell A1 contains...`

igitur:
Too big to include in 0.95. I'll start looking at this soon. In the meantime, please rebase and fix conflicts.

Pankraty:
Returning back to this PR after a long pause I realized that I have re-implement some parts different way. So make it WIP.

jahav:
I am closing this, because I think it is an incorrect way forward and it doesn't look like it's worth it.

The PR is trying to decrease memory usage by referencing the same formula in some form. That should help with the problems of repeating formulas in tables, e.g. each row contains a formula that is different in A1, but same in R1C1:

|     |           A        |         B          |     B represented in R1C1             |
| -- | ------------- | ------------- | ------------- |
| 1  | "Cake"         | =LENGTH(A1)  | =LENGTH(R[0]C[-1]) |
| 2  | "Pie"           | =LENGTH(A2) | =LENGTH(R[0]C[-1]) |
| 3  | "Waffle"           | =LENGTH(A3) | =LENGTH(R[0]C[-1]) |

That is done through indirection from `XLCellFormula` (specific to a cell) reference shared `XLFormulaDefinition`.

This is from a paper that analyzed spreadsheets from real world usage: Enron's Spreadsheets and Related Emails: A Dataset and Analysis:
> The 15,770 spreadsheets together have 97,636,511 nonempty cells of which 20,277,835 are formulas. This is an
average of 6,191 cells and 1,286 formulas per spreadsheet. Not all spreadsheets in the Enron set contain formulas:
there are 9,120 spreadsheets containing formulas, which is about half of all spreadsheets in the set.
> **Unique formulas** are defined in terms of having an **equal relative R1C1** notation
> The average spreadsheet with formulas contains **2,223 formulas of which 100 are unique**. In the EUSES set, spreadsheets have an **average of 372 formulas and 33 unique formulas**, which is substantially lower.

So the real examples don't have that many non-unique elements.

Next thing is that we are trying to save memory, the tradeoff is keeping them as a string.

The average length of a formula is ~21 characters, so string object takes up 8 bytes for a reference and 8(synblk)+8(type header)+4(length)+21*2(data)+2 (zero at end) = 64 bytes per formula. Even if we had 500'000 in a sheet, that is 32 MB and that is for a massive amount of formulas. I don't know how much memory object in PR take up, but I doubt it actually decreases memory consumption.

The CLosedParser also parses average formula in ~2us, so on one thread, we can parse 500'000 in 1 second.

Basically, I think it tries to solve a problem that doesn't really exist and it's uncertain that it would actually be improvement.

The way forward is to keep formulas as a strings and parse them each time without actually constructing AST in memory. Just evaluate formula through visitor during parsing and parse for each formula evaluation. I would bet, that most use cases of ClosedXML also parse exactly once, to evaulate formulas. Also use more threads (hopefully.... some day...).

PR #1208, opened 2019-05-13
optional range consolidation via save options

### LINKED ISSUE #989: XLConditionalFormat formats independent columns as one range for ColorScale

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.94

When trying to format several independent columns using ColorScale, all of them are formating in one range.

Each column is formatting separately.

No

## Reproducibility

```c#
public void Main()
{
    using (var output = new MemoryStream())
            {
                {
                    var sheet = wb.Worksheets.Add("test");
                    sheet.Column(1).AddConditionalFormat().ColorScale()
                    .LowestValue(XLColor.FromHtml("#f56a6d"))
                    .Midpoint(XLCFContentType.Percent, 50, XLColor.FromHtml("#fdec84"))
                    .HighestValue(XLColor.FromHtml("#63be7b"));

                    sheet.Column(2).AddConditionalFormat().ColorScale()
                    .LowestValue(XLColor.FromHtml("#f56a6d"))
                    .Midpoint(XLCFContentType.Percent, 50, XLColor.FromHtml("#fdec84"))
                    .HighestValue(XLColor.FromHtml("#63be7b"));

                    wb.SaveAs(output);
                }
            }
}
```

Current result:
default

Expected result:

### LINKED ISSUE #1113: add Databars (conditional formatting) without format consolidation

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.94.2

I want to add a databar as conditional formatting rules for several columns of a table. Unfortunately, the ranges for all columns are consolidated. Therefore the size of the databars are depicted based on the lowest/ highest value in all ranges together.

I want to add a databar for each column separately (as separate conditional formatting rule) resp. supress conditional formatting consolidation, so that the size of a data bar is calculated based on the lowest/ highest value in each column separately.

No (not sure)

## Reproducibility

```c#
 foreach (var column in listOfColumnNames)
            {             resultTable.DataRange.Column(column).AddConditionalFormat().DataBar(numberFormattingColor).LowestValue().HighestValue();
            }
```
currentresult
expectedresult

ladida28:
#### What's this PR do?
this PR adds two more options to save_options
- Boolean ConsolidateConditionalFormatRanges
- Boolean ConsolidateDataValidationRanges

Therewith it can be configured, whether ranges are consolidated for conditional formats and data validation when the workbook is saved. Per default both are set to true

two tests are added:
- OptionalRangeConsolidationWithoutConsolidation
- OptionalRangeConsolidationWithConsolidation

from my perspective, manual tests are not necessary, test are available

Closes #1113 
Closes #989

#### Questions:
- Is there a blog post?
no
yes: the two options should be explained
no


ClosedXML_Tests/Excel/ConditionalFormats/ConditionalFormatTests.cs
```diff
@@ -21,5 +21,41 @@ public void MaintainConditionalFormattingOrder()
...
+            var options = new SaveOptions
+            {
+                ConsolidateConditionalFormatRanges = false,
+                ConsolidateDataValidationRanges = false
+            };
+
+            using (var stream = TestHelper.GetStreamFromResource(TestHelper.GetResourcePath(@"Other\StyleReferenceFiles\ConditionalFormattingOptional\inputfile_multiple_ranges.xlsx")))
+            using (var ms = new MemoryStream())
+            {
+                TestHelper.CreateAndCompare(() =>
```
Pankraty: These CreateAndCompare tests bring a lot of headaches because every time the saving logic changes many of such tests start to fail even if the changes have no relation to a feature under testing.

This is why I recommend avoiding such kinds of tests if possible and leave them for complicated cases only.

So, if you are intended to check that `ConsolidateConditionalFormatRanges = false` turns the consolidation off, then just save the file to a memory stream, load it from there and ensure a number of conditional formats is 2. Same for data validations. This way the test will be much easier to support.

ClosedXML/Excel/SaveOptions.cs
```diff
@@ -17,5 +17,7 @@ public SaveOptions()
         public Boolean EvaluateFormulasBeforeSaving { get; set; } = false;
         public Boolean GenerateCalculationChain { get; set; } = true;
         public Boolean ValidatePackage { get; set; }
+        public Boolean ConsolidateConditionalFormatRanges { get; set; } = true;
+        public Boolean ConsolidateDataValidationRanges { get; set; } = true;
```
Pankraty: I believe the properties are ordered alphabetically. Put these two accordingly.

ClosedXML/Excel/XLWorkbook_Save.cs
```diff
@@ -304,7 +304,7 @@ private void CreateParts(SpreadsheetDocument document, SaveOptions options)
                     GenerateVmlDrawingPartContent(vmlDrawingPart, worksheet, context);
                 }
 
-                GenerateWorksheetPartContent(worksheetPart, worksheet, options.EvaluateFormulasBeforeSaving, context);
+                GenerateWorksheetPartContent(worksheetPart, worksheet, options.EvaluateFormulasBeforeSaving, options.ConsolidateConditionalFormatRanges, options.ConsolidateDataValidationRanges, context);
```
Pankraty: Maybe simply pass `SaveOptions` to the method? For one flag it was OK to pass a boolean, but for three it looks weird.

ClosedXML_Tests/TestHelper.cs
```diff
@@ -101,7 +101,7 @@ public static void RunTestExample<T>(string filePartName, bool evaluateFormulae
             }
         }
 
-        public static void CreateAndCompare(Func<IXLWorkbook> workbookGenerator, string referenceResource, bool evaluateFormulae = false)
+        public static void CreateAndCompare(Func<IXLWorkbook> workbookGenerator, string referenceResource, bool evaluateFormulae = false, SaveOptions options = null)
```
Pankraty: Once `options` has been added there is no sense in `evaluateFormulae` argument. Remove it.

Pankraty: On the other hand, if you get rid of CreateAndCompare tests you added, this method may be reverted.

ClosedXML/Excel/XLWorkbook_Save.cs
```diff
@@ -4473,9 +4473,13 @@ private static bool NumberFormatsAreEqual(NumberingFormat nf, XLNumberFormatValu
         #region GenerateWorksheetPartContent
 
         private static void GenerateWorksheetPartContent(
-            WorksheetPart worksheetPart, XLWorksheet xlWorksheet, bool evaluateFormulae, SaveContext context)
+            WorksheetPart worksheetPart, XLWorksheet xlWorksheet, bool evaluateFormulae, bool consolidateConditionalFormatRanges, bool consolidateDataValidationRanges, SaveContext context)
         {
-            ((XLConditionalFormats)xlWorksheet.ConditionalFormats).Consolidate();
+            if(consolidateConditionalFormatRanges)
```
Pankraty: Add a space after `if`

ClosedXML/Excel/XLWorkbook_Save.cs
```diff
@@ -5359,7 +5363,12 @@ private static bool NumberFormatsAreEqual(NumberingFormat nf, XLNumberFormatValu
                 var dataValidations = worksheetPart.Worksheet.Elements<DataValidations>().First();
                 cm.SetElement(XLWorksheetContents.DataValidations, dataValidations);
                 dataValidations.RemoveAllChildren<DataValidation>();
-                xlWorksheet.DataValidations.Consolidate();
+
+                if(consolidateDataValidationRanges)
```
Pankraty: Add a space after `if`

ClosedXML/Excel/XLWorkbook_Save.cs
igitur: To add to this, install the CodeMaid Visual Studio plugin and do a CodeMaid Clean on each file that you have altered. It will ensure the whitespace is in line with our convention.

Pankraty:
Besides making a consolidation optional (which may benefit performance), I would fix the `XLConditionalFormats.Consolidate` method in order to exclude certain types of conditional formats (i.e. `ColorScale`, `DataBar`, something else?) from a consolidation. They must not be screwed up even if the user enforces a consolidation.

The fix should be trivial, I expect, but it will require adding a couple more tests.

igitur:
@ladida28 FYI, I just added some developer guidelines. Hopefully this helps. https://github.com/ClosedXML/ClosedXML/blob/develop/CONTRIBUTING.md

ClosedXML_Tests/Excel/ConditionalFormats/ConditionalFormatTests.cs
ladida28: test implementation updated

ClosedXML/Excel/SaveOptions.cs
ladida28: codemaid applied in new version

ClosedXML/Excel/XLWorkbook_Save.cs
ladida28: adjusted according to your comment

ClosedXML_Tests/TestHelper.cs
ladida28: test implementation adjusted - no createAndCompare used anymore

ClosedXML/Excel/XLWorkbook_Save.cs
ladida28: codemaid applied

ClosedXML/Excel/XLWorkbook_Save.cs
ladida28: codemaid applied

ladida28:
I can implement this, but I am not sure were to store the list of conditionalFormatTypes, that should be excluded.
Option 1: directly in XLConditionalFormats
Option 2: configurable via save_options
Option 3: somewhere else?

What would you suggest?

> Besides making a consolidation optional (which may benefit performance), I would fix the `XLConditionalFormats.Consolidate` method in order to exclude certain types of conditional formats (i.e. `ColorScale`, `DataBar`, something else?) from a consolidation. They must not be screwed up even if the user enforces a consolidation.
> 
> The fix should be trivial, I expect, but it will require adding a couple more tests.

Pankraty:
>Option 1: directly in XLConditionalFormats
>Option 2: configurable via save_options
>Option 3: somewhere else?

It's definitely not a save option, such formats should be excluded from a consolidation at all. So I suggest putting an additional check right in XLConditionalFormats.Consolidate().

ClosedXML_Tests/Excel/ConditionalFormats/ConditionalFormatTests.cs
```diff
@@ -21,5 +22,51 @@ public void MaintainConditionalFormattingOrder()
...
         }
+
+        [Test]
+        public void OptionalRangeConsolidationWithoutConsolidation()
+        {
+            var options = new SaveOptions
+            {
+                ConsolidateConditionalFormatRanges = false,
+                ConsolidateDataValidationRanges = false
+            };
```
Pankraty: Now it is much better. But still, this test checks too many things at once. 

So,
1. Join `OptionalRangeConsolidationWithoutConsolidation` and `OptionalRangeConsolidationWithConsolidation` in one test `ConsolidationDoesNotAffectDataBarAndColorScale`. There you don't even have to save a workbook. You can call `((XLConditionalFormats)(ws.ConditionalFormats)).Consolidate()` directly.
2. Add another test to check that `ConsolidateConditionalFormatRanges` has the effect. For this, you will create some "ordinary" conditional formats that are supposed to be consolidated with the option turned on. To avoid code duplication, create a single test with two test cases (e.g. 
```
[TestCase(true, 1)]
[TestCase(false, 2)]
public void SaveOptionAffectsConsolidation(bool consolidateConditionalFormatRanges, int expectedCount)
```
3. In `DataValidationTests` create a similar test with two test cases for conditional formats consolidation.

Pankraty: Besides, once you removed those CreateAndCompare tests you also have to remove XLSX files (ConditionalFormattingOptional_consolidated.xlsx, ConditionalFormattingOptional_not_consolidated.xlsx, 
inputfile_multiple_ranges.xlsx)

ladida28: files removed

ladida28: test case implementation adjusted

ladida28:
> > Option 1: directly in XLConditionalFormats
> > Option 2: configurable via save_options
> > Option 3: somewhere else?
> 
> It's definitely not a save option, such formats should be excluded from a consolidation at all. So I suggest putting an additional check right in XLConditionalFormats.Consolidate().

I adjusted XLConditionalFormats.Consolidate() according to your comments

Pankraty:
Good work! Thanks for the contribution and do not stop ;) There is a whole lot of tasks to address.

ladida28:
> Good work! Thanks for the contribution and do not stop ;) There is a whole lot of tasks to address.

Thanks. Do you already know when it will be merged? Will it be part of v0.95. I am desperately waiting for this change :)

Besides, I am thinking about adding support for data slicers, but I actually do not really know, where to start. Any recommendations are welcome

Pankraty:
> Thanks. Do you already know when it will be merged? Will it be part of v0.95. I am desperately waiting for this change :)

This is up to @igitur to decide.

> Besides, I am thinking about adding support for data slicers, but I actually do not really know, where to start. Any recommendations are welcome

I would start from opening a new issue since it has no relation to this PR :)

ClosedXML/Excel/ConditionalFormats/XLConditionalFormats.cs
```diff
@@ -1,13 +1,24 @@
...
 using System.Linq;
-using System.Text;
 
 namespace ClosedXML.Excel
 {
     internal class XLConditionalFormats : IXLConditionalFormats
     {
         private readonly List<IXLConditionalFormat> _conditionalFormats = new List<IXLConditionalFormat>();
+
+        private readonly List<XLConditionalFormatType> _conditionalFormatTypesExcludedFromConsolidation = new List<XLConditionalFormatType>()
```
igitur: This could be `static`. Maybe better to use `HashSet<>`, @Pankraty ?

igitur: @ladida28 Please make this static.

igitur:
@ladida28 Are you comfortable with rewriting git history? Ideally I want to squash this into 2 commits. I'll give more info later, if you can rewrite git history.

ladida28:
> @ladida28 Are you comfortable with rewriting git history? Ideally I want to squash this into 2 commits. I'll give more info later, if you can rewrite git history.

@igitur Never did that before but I can give it a try using smartgit. Which commits should be squashed together?

igitur:
Ok, nevermind, it's OK to leave it as is. I'll squash it all into 1 commit. Just see the last review comment, please.

ClosedXML/Excel/ConditionalFormats/XLConditionalFormats.cs
ladida28: static is set

ladida28:
> > Thanks. Do you already know when it will be merged? Will it be part of v0.95. I am desperately waiting for this change :)
> 
> This is up to @igitur to decide.
> 
> > Besides, I am thinking about adding support for data slicers, but I actually do not really know, where to start. Any recommendations are welcome
> 
> I would start from opening a new issue since it has no relation to this PR :)

comming soon :)

igitur:
Thanks for the contribution

PR #1209, opened 2019-05-14
Add developer guidelines

igitur:
First stab at adding developer guidelines.


igitur:
I should add some comments about CodeMaid.

Pankraty:
Looks great!

PR #1212, opened 2019-05-20
Fix opening of file with pivot table created from same-named columns

### LINKED ISSUE #1211: Cannot open file with pivot table created from same-named columns
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.95.0-beta1 (tested also with develop)

`System.ArgumentOutOfRangeException : The column 'Résultat ToxB2' does not appear in the source range.
Parameter name: sourceName` 

ClosedXML should be able to cope with pivoted source columns that have the same name. It seems that MS Excel adds running number to them to distinguish them apart.

Yes, at least with 0.88 which I'm trying to upgrade from.

## Reproducibility

```c#
public void Main()
{
new XLWorkbook(File.OpenRead(@"c:\temp\test.xlsx"), XLEventTracking.Disabled);
}
```

[test.xlsx](https://github.com/ClosedXML/ClosedXML/files/3197052/test.xlsx)

lahma:
#### What's this PR do?

fixes #1211 

Automated test case should suffice.

#1211 

#### Questions:
- Is there a blog post? *no*
- Does the knowledge base need an update? *no*
- Does this add new (C#) dependencies? *no*


igitur:
Thanks for the contribution. I think for this change you would specifically have to add tests to check whether the file also saves correctly. Pivot tables are finicky and a wrong field name will corrupt the file.

lahma:
I tried to test saving with following test case, and it indeed did fail.

```
[Test]
public void LoadingAndSavingPivotTableUsingDuplicateNames()
{
    using (var stream = TestHelper.GetStreamFromResource(TestHelper.GetResourcePath(@"Misc\PivotTableWithDuplicateSourceNames.xlsx")))
    {
        {
            wb.SaveAs(ms);
        }

        ms.Seek(0, SeekOrigin.Begin);

        using (var wb = new XLWorkbook(ms))
        {
        }
    }
}
```
Unfortunately I'm not familiar with the pivot logic enough to create required fix, I tried and mostly got errors about invalid document parts. Can anyone help out with this? I'm only trying to read files with pivot tables (I'm interested in the data). I can also close this PR if you want to derive your own PR.

lahma:
I've rebased this PR and still hoping that it could land in the main repository.

lahma:
This doesn't seem to be an issue with the latest NuGet release anymore.

PR #1219, opened 2019-05-28
Implement MATCH and INDEX functions

### LINKED ISSUE #1218: INDEX function not recognized when access from another cell

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.95.0-beta1

when another cell in a workwook access a cell with INDEX function , the INDEX function is not recognized and error is returned : "Unknown function: INDEX"

without that cell access , the INDEX function is correctly recognized and return correct value

should return the correct value even when the cell is accessed somewhere else

in 0.94.2, i believe the same problem occurs but a different error is thrown : 
'Unknown function: ClosedXML.Excel.CalcEngine.CellRangeReference'

maybe i'm mistaken but a simple sum with a cell range work

## Reproducibility

```c#
public void Main()
{
  var wb = new XLWorkbook(@"bug_ClosedXML.xlsx");
  var ws = wb.Worksheet(1);
  var cellA = ws.Cell("L10");
  var cellB = ws.Cell("L11");
  Console.WriteLine(cellA.Value);
  cellB.FormulaA1 = "L10";
  Console.WriteLine(cellA.Value);
}
```
- [ [bug_ClosedXML.xlsx](https://github.com/ClosedXML/ClosedXML/files/3223670/bug_ClosedXML.xlsx)] I attached a sample spreadsheet

igitur:
Fixes #1218 

Did some CodeMaid reorganising too.

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Functions/Lookup.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/Lookup.cs
@@ -19,7 +19,7 @@ namespace ClosedXML.Excel.CalcEngine.Functions
             //ce.RegisterFunction("GETPIVOTDATA", , Getpivotdata); // Returns data stored in a PivotTable report
             ce.RegisterFunction("HLOOKUP", 3, 4, Hlookup); // Looks in the top row of an array and returns the value of the indicated cell
             ce.RegisterFunction("HYPERLINK", 1, 2, Hyperlink); // Creates a shortcut or jump that opens a document stored on a network server, an intranet, or the Internet
-            //ce.RegisterFunction("INDEX", , Index); // Uses an index to choose a value from a reference or array
+            ce.RegisterFunction("INDEX", 2, 4, Index); // Uses an index to choose a value from a reference or array
             //ce.RegisterFunction("INDIRECT", , Indirect); // Returns a reference indicated by a text value
             //ce.RegisterFunction("LOOKUP", , Lookup); // Looks up values in a vector or array
             ce.RegisterFunction("MATCH", 2, 3, Match); // Looks up values in a reference or array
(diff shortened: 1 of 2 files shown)


ClosedXML/Excel/CalcEngine/Functions/Lookup.cs
```diff
@@ -87,6 +88,84 @@ private static object Hyperlink(List<Expression> p)
...
+
+                default:
+                    throw new NoValueAvailableException("Invalid match_type");
+            }
+
+            IXLCell foundCell = null;
+
+            if (match_type == 0)
+                foundCell = range
+                    .CellsUsed(XLCellsUsedOptions.Contents, c => !c.IsEmpty() && lookupPredicate.Invoke(new Expression(c.Value).CompareTo(lookup_value)))
```
Pankraty: ` .CellsUsed(XLCellsUsedOptions.Contents, c => !c.IsEmpty() && lookupPredicate.Invoke(new Expression(c.Value).CompareTo(lookup_value)))`

Does it really have to be this complicated? Does `.CellsUsed(XLCellsUsedOptions.Contents)` not return non-empty cells already? Can't we use `c => c.Value == lookup_value`? This way we'd lower the number of allocations (`new Expression(...)` for every analyzed cell).

ClosedXML/Excel/CalcEngine/Functions/Lookup.cs
```diff
@@ -87,6 +88,84 @@ private static object Hyperlink(List<Expression> p)
...
+                        {
+                            // When match_type != 0, we have to assume that the order of the items being search is ascending or descending
+                            var previousValueExpression = new Expression(previousValue);
+                            if (!lookupPredicate.Invoke(previousValueExpression.CompareTo(currentCellExpression)))
+                                return false;
+                        }
+
+                        previousValue = c.Value;
+
+                        return lookupPredicate.Invoke(currentCellExpression.CompareTo(lookup_value));
```
Pankraty: I think, instead of defining `lookup_predicate` we can preserve a simple int (-1, 0 or 1) and compare the result with it (`return currentCellExpression.CompareTo(lookup_value) == expectedComparisonResult`).

ClosedXML/Excel/CalcEngine/Functions/Lookup.cs
igitur: You're right about hte `IsEmpty()`, but the `Expression` way is necessary to correctly do comparison of different types and also case insensitive string comparison. Lessons I learnt from implementing `VLOOKUP`s. Maybe we should factor out `Expression.CompareTo()` in order instantiate fewer `Expression`s. I'd leave that for a separate PR though.

ClosedXML/Excel/CalcEngine/Functions/Lookup.cs
igitur: But when `match_type != 0`, then `lookup_predicate` uses '<=', not '<', so a single `int` won't be enough.

Pankraty: Ah, I missed that.

igitur: Also, the result can be other values too, e.g. `StringComparer.OrdinalIgnoreCase.Compare("asdf", "qwer") == -16`

Issue #1186, opened 2019-04-17
Can't create pivot styles for multiple value columns.


b0bi79:
**Do you want to request a *feature* or report a *bug*?**
- [x] Feature

**Version of ClosedXML**
0.95.0-beta1

**What is the current behavior?**
In the API pivotstyles there is a lack. We cannot assign different styles for different data fields.
```
nameField.StyleFormats.DataValuesFormat
                        .ForValueField(ordersField)
                        .Style.Fill.BackgroundColor = XLColor.Apricot;
nameField.StyleFormats.DataValuesFormat
                        .ForValueField(qtyField)
                        .Style.Fill.BackgroundColor = XLColor.ArmyGreen;
```
two styles should be created here, but it turns out that the same style will be changed

**What is the expected behavior or new feature?**
I suggest doing this:
```
nameField.StyleFormats.AddValuesFormat()
                        .ForValueField(ordersField)
                        .Style.Fill.BackgroundColor = XLColor.Apricot;
nameField.StyleFormats.AddValuesFormat()
                        .ForValueField(qtyField)
                        .Style.Fill.BackgroundColor = XLColor.ArmyGreen;
```

**Did this work in previous versions of our tool?  Which versions?**
No

PR #1222, opened 2019-05-29
Second revision of pivot styles API #1186

b0bi79:
#### What's this PR do?
The refined API now allows to assign styles for:
- each individual value column;
- the totals of each individual value column;
- labels of the totals;
- any kind of totals, not just for DefaultSubtotal.
- правильно загружать стили файлов, созданных в MS Excel

Contains fix for #1186


b0bi79:
This PR is needed for the release of ClosedXML.Report

igitur:
I'm finally looking at this PR. But it seems like you have undone many of the changes I made when we last worked on the pivot table formats. For example, now we're back to using `public interface IFieldRef` again, which is not a good public name for an interface. Also, some of the old patterns are back too.

b0bi79:
I had to go back to the old models because the models you entered could not describe all the options for using Pivot styles. And for this reason, I had to make many interfaces open.

b0bi79:
Sorry, Francois. I again did not a beautiful interface. If I understood what you don't like, maybe I could have done better. And it seemed to me that we agreed with you that I would make a public interface like yours, and an additional "low-level" API with lambda functions, to which you agreed.

ClosedXML_Examples/PivotTables/PivotTableStyles.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            var amountPaidField = pt.Values.Add("AmountPaid", "Amount paid").SetSummaryFormula(XLPivotSummary.Sum);
+            var itemsTotalField = pt.Values.Add("ItemsTotal", "Items total").SetSummaryFormula(XLPivotSummary.Sum);
+
+            // styles
+            var purpleBlue = XLColor.FromArgb(102, 102, 153);
+            var brightYellow = XLColor.FromArgb(255, 255, 153);
+            var gray = XLColor.FromArgb(192, 192, 192);
+
+            companyField.AddHeaderStyle().Style.Font.SetBold();
```
igitur: I don't see any fields in the output file that display as bold.

igitur:
Is it possible that you can generate a pivot table that illustrates every kind of pivot table format? If I understand the limitations of the current implementation on the `develop` branch, then I will understand better what this PR intends.

b0bi79:
At first it seemed to me that the options that you described should describe everything possible. But when I began to open and save different documents with pivot tables, I realized that it was impossible to describe everything (or it was very difficult). Therefore, I decided that it is better to return to the API that you do not like.

ClosedXML_Examples/PivotTables/PivotTableStyles.cs
b0bi79: Fixed

ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotStyleFormatEnums.cs
```diff
@@ -5,7 +5,6 @@ namespace ClosedXML.Excel
     [Flags]
     public enum XLPivotStyleFormatElement
     {
-        None = 0,
```
Pankraty: Why remove the default value? Since `enum` is a value type, the variable of this type always has value, even before it is assigned explicitly. I rather have `None` than `0` in that case.

ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotStyleFormatEnums.cs
```diff
@@ -21,4 +20,24 @@ internal enum XLPivotStyleFormatTarget
...
+        All = 3,
+        Origin = 4,
+        Button = 5,
+        TopRight = 6,
+        TopEnd = 7
+    }
+
+    public enum XLPivotTableAxisValues
+    {
+        AxisRow = 0,
```
Pankraty: Same here. Unless the particular numbers have some meaning (i.e. they coincide to OpenXML values), it is better to reserve `0` to `NotSet` than use an arbitrary value as a default for non-assigned variables. It protects from rare unpleasant bugs.

ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotValueStyleFormat.cs
```diff
@@ -1,30 +1,40 @@
...
 {
-    internal class XLPivotValueStyleFormat : XLPivotStyleFormat, IXLPivotValueStyleFormat
+    internal class XLPivotValueStyleFormat : IXLPivotValueStyleFormat
     {
-        public XLPivotValueStyleFormat(IXLPivotField field = null, IXLStyle style = null)
-            : base(field, style)
-        { }
+        public XLPivotValueStyleFormat(IXLPivotFormat pivotFormat)
+        {
+            PivotFormat = pivotFormat;
```
Pankraty: Better check the input for null here than have NRE in `Style` getter.

ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotValueStyleFormat.cs
```diff
@@ -1,30 +1,40 @@
...
+
         public IXLPivotValueStyleFormat AndWith(IXLPivotField field)
         {
             return AndWith(field, null);
         }
 
         public IXLPivotValueStyleFormat AndWith(IXLPivotField field, Predicate<Object> predicate)
         {
-            FieldReferences.Add(new PivotLabelFieldReference(field, predicate));
+            ((List<IFieldRef>)PivotFormat.FieldReferences).Add(FieldRef.ForField(field.SourceName, predicate));
```
Pankraty: Better avoid typecast here. For example, in the constructor, you can accept a concrete implementation of `IXLPivotFormat` that allows modification of `FieldReferences` collection. Or define an internal interface for that. I can provide a more detailed explanation if needed.

ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotStyleFormatEnums.cs
b0bi79: Because there are only two possible options - Label or Data. The third option can not be processed. If leave None, will need to check and throw an Exception in the case of None. I removed it to eliminate possible errors.

b0bi79: Default here should be Data.

Pankraty: Dealing with `enum`s you always have to keep in mind that there can be other values, not listed in the enum because somebody can cast an arbitrary value to it (`(XLPivotStyleFormatElement)999`) intentionally or by mistake. And if the code does not accept `None` this should be stated explicitly; removing valud from enum is not enough.

Consider this:
```
struct Test
{
  public XLPivotStyleFormatElement StyleFormatElement {get; set;}
}

var t = new Test();
//t.StyleFormatElement == (XLPivotStyleFormatElement)0 at this point, but XLPivotStyleFormatElement.None or XLPivotStyleFormatElement.NotSet looks nicer.
```

b0bi79: If you leave `None`, users will make such calls `AddSubtotalStyle(XLPivotStyleFormatElement.None)`, which will lead to Exception.
I understand that you are trying to adhere to the rules/patterns, but will it be correct to give users a choice of pre-disabled options?
The only place where the property XLPivotStyleFormatElement is declared [here](https://github.com/ClosedXML/ClosedXML/blob/74e5c1bdb1e4263d158e6766ab49e49072578ecb/ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotFormat.cs#L14). In this case, the default is Data.

Pankraty: Fair point. Accepted

PR #1225, opened 2019-06-01
POC: Add performance tests

Pankraty:
I have been thinking about implementing the performance tests that would allow the developer to notice the performance degradation early. Those tests must be environment independent, and they should not require too much effort to run. Thus, we cannot just save the results of the benchmark at a given point in time and compare to them - because the next run may happen on another machine. And we cannot force the developer to checkout master, run the benchmarks, then checkout the feature branch, run the benchmarks again and compare results manually.

I managed to implement a simple benchmark using [BenchmarkDotNet](https://github.com/dotnet/BenchmarkDotNet/) that compares the current state of ClosedXML against the latest released one. To be able to do this I cheated a little: the "baseline" run uses a signed version of the library, while the "current" run works with the unsigned. This, I expect, does not influence the results of the benchmark but will, I am afraid, make it harder to upgrade the dependencies.

The open question is whether or not to include the performance tests in the AppVeyor pipeline. They are going to be very slow, but on the other hand, they will prevent us from accepting the PRs killing the performance.


b0bi79:
Alexey, a great contribution. Very useful.

Now I see one test there. This one test takes a long time to run or do you mean that there will be some tests added that will take a long time to run?

Pankraty:
Thanks Alexey! (я выспался днем, вот и прогал допоздна ))

For now, there is a single very simple test which takes around 40 seconds to run because BenchmarkDotNet does many runs in order to determine the mean value more precisely. But of course, there will be much more tests, and more complex too. For example, add 1000 merges, or open a large file, or insert 100K rows, and so on. So I will not be surprised if entire test suite will be taking 30-60 minutes.

b0bi79:
Then I think it makes sense to write one test (or 2-3 tests), which includes the main functions of ClosedXML and run this test in the AppVeyor pipeline. Other tests should be runned from time to time (for example, before building a release).

Pankraty:
I am going to try this lib https://www.nuget.org/packages?q=CodeJam.PerfTests. It uses BenchmarkDotNet too but is designed for performance tests to run quicker.

igitur:
I think these tests should be performed on AppVeyor only when building a tag.

PR #1228, opened 2019-06-05
Rename Misc folder to TryToLoad

igitur:
Any objections against this? @Pankraty @b0bi79 

I think this will make the folder name more descriptive for the intent.


b0bi79:
Agree

PR #1231, opened 2019-06-12
Add support for Emoji via Unicode.net library

### LINKED ISSUE #1226: Composited unicode characters not displayed correctly.

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.94.2

When I fill a cell with value: `C0  👍 0/2 ✖0 ⚠️0`

It is displayed like this: `C0  _xD83D__xDC4D_ 0/2 ✖0 ⚠️0` in LibreOffice 6.2.3.2

Unicode emoji should display correctly.

Now, I don't know if this is a ClosedXML bug, or a LibreOffice bug. It seems 👍 is a composited unicode character... dunno which are the rules to display this in excel and libreoffice.

## Reproducibility

```c#
public void Main()
{
    var book = new XLWorkbook();
    var sheet = book .AddWorksheet("sheet");
    var cell = sheet.Cell(1,1);
    cell.Value ="C0  👍 0/2 ✖0 ⚠️0";
    book.SaveAs("unicode.xlsx");
}
```

igitur:
Adds better support for writing emoji to file. Uses https://github.com/neosmart/unicode.net library. 

We should check performance impact before merging.

Fixes #1226


ClosedXML_Tests/Resource/Other/Emoji/let_it_go_in_emoji.txt
```diff
@@ -0,0 +1 @@
+The ❄️ 🌟 🔦 ⚪ on the mountain 🌙 🌠. 🙅🏻 a👣 to 🐝  👀. A 🏰 of 😢, and it 👀  like☝️️ the 👑. The 💨 is 🐺 like this 🌀 ❄️ ☔️ 🏠. 🙅🏻 keep it in, ☁️ 💡 ☝️️ tried. 🙅🏻 let 👬👫 in,🙅🏻 let 👬👫 👀. 🐝 the 👍  👧  👇 always have to 🐝. 🙅🏻, don't 👐, 🚫 let 👬👫💡. Well now 👬👫 💡. 👐 it 🚗,, 👐 it 🚗,,🙅🏻 ✊ it back anymore. 👐 it 🚗,, 👐 it 🚗,  turn ✈️ and 🔨 the 🚪. ☝️️ 🚫 care, what 👬👫  going to 👄, let the ☔️ ⚡ ❄️  😡 on, the ❄️ ⛄️ 🙅🏻 bothered ☝️️ anyway. It's 😜😂  how some ✈️ 🚆 makes everything 😳  🐜. And the 😱  that once 👮  me, 🙅🏻 get to☝️️ at all. It's 🕓  to 👀  what☝️️ can do. To 📝 the 📊 and 🔨 through. 🚫 👍 , 🚫  👎, 🚫  👮 for ☝️️. ☝️️ 🏃. 👐 it 🚗,, 👐 it 🚗., ☝️️ am ☝️ with the 🌀  and 🌌. 👐 it 🚗,, 👐 it 🚗..👇 🙅🏻 👀  ☝️️  😭 .  👉 ☝️️ 🚶, and 👉 ☝️️ stay. Let the⚡ ❄️ 😡 on. ☝️️ 💪 ❄️ through the 🌀 into the 🌎.☝️️ 👤 is 🌀 in ❄️ ⛄️ fractals all 🔁. And 1️⃣💡 💎 like an ❄️ 📢.  ☝️️ 🙅🏻 🏃 back, the past is in the past. 👐 it 🚗,,👐 it 🚗,. And ☝️️ 🚀 like the 💔 of 🌌. 👐 it 🚗,, 👐 it 🚗.. That 💁 is 🚫. Here ☝️️ 🚶, in the 🔦 of ☀️. Let the ⚡ ❄️  😡  on, the ❄️ ⛄️ 🙅🏻 bothered ☝️️ anyway.
```
Pankraty: 🤣

igitur:
Also waiting for a signed version of Unicode.net to be released. Looks like a commit has already been merged for this: https://github.com/neosmart/unicode.net/commit/f4b00be5cf95f8488a8b3627f6333b73a223d378

igitur:
@Pankraty Do you think the benefit of this PR worth the extra dependency on the 3rd party library?

igitur:
> Also waiting for a signed version of Unicode.net to be released. Looks like a commit has already been merged for this: [neosmart/unicode.net@f4b00be](https://github.com/neosmart/unicode.net/commit/f4b00be5cf95f8488a8b3627f6333b73a223d378)

This has been resolved.

Pankraty:
I am surprised with the size of that library - around 562 KB for each platform. Considering ClosedXML itself is about 1120 KB, adding half a megabyte only to support emojis may be an overkill.

igitur:
That's because the library lists all known emoji (actually duplicates them)
https://github.com/neosmart/unicode.net/blob/master/unicode/Emoji-All.cs
https://github.com/neosmart/unicode.net/blob/master/unicode/Emoji-Basic.cs

And we don't need that. If I remove those classes' entries, the library shrinks to 37kB. 

So maybe we should compile the classes that we use INTO ClosedXML?

Pankraty:
Yes, it makes sense. And the library code does not seem to change often so it won't be too hard to keep our copy in sync, if we like to.

jahav:
I am closing this PR in favor of #1978. 

Reasons:
* As far as I can tell, it can be solved in BCL 
* dependency on another library (850KiB at this time) is not acceptable. Maintaining fork is also not ideal solution.
* the proposed version didn't work correctly for all emoji and likely wouldn't work for all hieroglyphs and all other fun new characters in astral planes. See the xD83C encoded sequence in the image:

image

PR #1232, opened 2019-06-12
 Set versioning number to allow enhanced pivot table features

### LINKED ISSUE #1182: Pivot Table range limitation
- [x] Bug
- [ ] Feature

Version 0.94.2

I have a simple application that creates an Excel file, 2 sheets, a table in 1 sheet and then a pivot table in the second sheet using the table as the data.
The table has a static column count of 10. The row count varies but is generally between 35,000 and 40,000. As seen in the code below, I use a row count to define the range. But each time I ran it, the Excel file generated the old error:  _"We found a problem with some content in [file name]. Do you want to recover..."_ 
The repair log:
_Removed Feature: PivotTable report from /pivotCache/pivotCacheDefinition1.xml part (PivotTable cache)
Removed Feature: PivotTable report from /xl/pivotTables/pivotTable.xml part (PivotTable view)
Removed Records: Workbook properties from /xl/workbook.xml part (Workbook)._

After a lot of trial and error by manually setting the range, I was able to determine that if the rows were less than or equal to 32,895, the pivot table was created correctly. Anything above that row count and the Excel file generated the errors.
I expected any field limitations in ClosedXML to match Excel’s limitations.
This is the first project in which I have attempted to create pivot tables so I cannot say if this worked under older version.
Code to reproduce:
```c#
        {
            String fileName = "DeliveryCompliace " + DateTime.Today.ToString("MMddyyyy") + ".xlsx";
            String runTime = DateTime.Now.ToString("MM/dd/yyyy HH:mm");
            String reportDate = DateTime.Today.ToLongDateString();
            
            // Create Workbook
            // Create Worksheet
            IXLWorksheet ws = wb.Worksheets.Add("Data");

            
            // Fill table from stored procedure
            SqlConnection sqlConnection = new SqlConnection(Properties.Settings.Default.conString);
            SqlCommand sqlCommand = new SqlCommand("dbo.cex_DScanCompData", sqlConnection);
            sqlCommand.CommandType = CommandType.StoredProcedure;
            sqlCommand.Parameters.Add("@dDate", SqlDbType.DateTime).Value = reportDate;
            SqlDataAdapter sqlDataAdapter = new SqlDataAdapter(sqlCommand);
            DataTable dataTable = new DataTable();
            sqlDataAdapter.Fill(dataTable);

            // Insert Datatable data
            IXLTable table = ws.Cell(1, 1).InsertTable(dataTable, "Data", true);

            // Fix OrderTrackingID format in ws
            ws.Column(5).CellsUsed().Style.NumberFormat.Format = "#.000000";
            // Adjust column widths in ws
            ws.Columns().AdjustToContents();

            // Pivot Table
            // Add new worksheet
            IXLWorksheet ps = wb.Worksheets.Add("Report");

            // Create pivot table
            var totalRows = ws.LastRowUsed().RowNumber() - 5;
            var totalColumns = ws.LastColumnUsed().ColumnNumber();
            IXLPivotTable pt = ps.PivotTables.Add("PivotTable", ps.Cell(3, 1), ws.Range(1,1,totalRows,totalColumns)); // Max rows 32,895!!

            pt.RowLabels.Add("TerminalName").SetCollapsed();
            pt.RowLabels.Add("Driver").SetCollapsed();
            pt.RowLabels.Add("ConsCompany", "Client").SetCollapsed();

            pt.Values.Add("PkgCount").SetSummaryFormula(XLPivotSummary.Sum);
            pt.Values.Add("DScan").SetSummaryFormula(XLPivotSummary.Sum);
            pt.Values.Add("MScan").SetSummaryFormula(XLPivotSummary.Sum);
            pt.Values.Add("NScan").SetSummaryFormula(XLPivotSummary.Sum);

            //For testing purposes
            IXLWorksheet ws3 = wb.Worksheets.Add("Test");
            ws3.Cell("A1").Value = totalRows;
            ws3.Cell("A2").Value = totalColumns;

            // Save workbook
            wb.SaveAs(@"C:\Users\CEsolutions\Documents\Test2.xlsx");
        }
```
- [X] I attached a sample spreadsheet.  
[Test2.xlsx](https://github.com/ClosedXML/ClosedXML/files/3074698/Test2.xlsx)

igitur:
Pivot tables were hindered because the correct versioning attributes were not set. This is why the pivot tables in #1182 were restricted to 32500 cached values.

Fixes #1182


ClosedXML/Excel/XLWorkbook_Save.cs
```diff
@@ -2076,6 +2076,37 @@ private static void GenerateTableDefinitionPartContent(TableDefinitionPart table
...
             }
 
+            #region CreatedVersion
+
+            byte createdVersion = 5;
+
+            if (pivotCacheDefinition.CreatedVersion?.HasValue ?? false)
+                pivotCacheDefinition.CreatedVersion = Math.Max(createdVersion, pivotCacheDefinition.CreatedVersion.Value);
+            else
+                pivotCacheDefinition.CreatedVersion = createdVersion;
```
Pankraty: Maybe something like this?

```
byte createdVersionClosedXml = 5;
var createdVersionFromFile = pivotCacheDefinition.CreatedVersion.GetValueOrDefault();
pivotCacheDefinition.CreatedVersion = Math.Max(createdVersionFromFile, createdVersionClosedXml);
```

Or even extract `createdVersion`, `refreshedVersion` and `minRefreshableVersion` to some class with constants?

igitur: The existing pivot table reference files implicitly test this, but yes, I'll add an explicit test.

igitur:
> Is it possible to also add a unit test to #1182?

The existing pivot table reference files implicitly test this, but yes, I'll add an explicit test.

ClosedXML/Excel/XLWorkbook_Save.cs
igitur: `pivotCacheDefinition.CreatedVersion` is of type `ByteValue`, not `byte?` and hence doesn't have a `GetValueOrDefault()` method.

ClosedXML/Excel/XLConstants.cs
```diff
@@ -4,13 +4,18 @@ namespace ClosedXML.Excel
...
-        public const byte PivotTableCreatedVersion = 5;
-        public const byte PivotTableRefreshedVersion = 5;
-        public const string PivotTableValuesSentinalLabel = "{{Values}}";
+        public static class PivotTable
+        {
+            internal const byte CreatedVersion = 5;
+            internal const byte RefreshedVersion = 5;
+
+            //TODO: Needs to be refactored to be more user-friendly.
+            public const string ValuesSentinalLabel = "{{Values}}";
```
Pankraty: Does it have to be `public`? Maybe the whole `PivotTable` class can be made internal?

igitur: Once I refactor the ugly public use of `ValuesSentinalLabel`, then this can be made internal.

bassvel:
So what exactly user can do with the hindered pivots? It keep on returning me an err _we found a problem with some content_

PR #1233, opened 2019-06-13
WIP Refactor value comparison into its own IEqualityComparer

igitur:
Depends on #1237

As mentioned before to @pankraty.

Now we don't have to create a new `Expression` when comparing cell values in e.g. `VLOOKUP`.

The comparer can also be reused to calculate the `BaseItem` value for pivot tables.

@Pankraty , you're the master of computing patterns, so let me know if this can be improved.

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Functions/Lookup.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/Lookup.cs
@@ -1,5 +1,6 @@
 // Keep this file CodeMaid organised and cleaned
 using ClosedXML.Excel.CalcEngine.Exceptions;
+using ClosedXML.Excel.Patterns;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -59,7 +60,8 @@ namespace ClosedXML.Excel.CalcEngine.Functions
                 throw new CellReferenceException("Row index has to be positive");
 
             IXLRangeColumn matching_column;
-            matching_column = range.FindColumn(c => !c.Cell(1).IsEmpty() && new Expression(c.Cell(1).Value).CompareTo(lookup_value) == 0);
+            matching_column = range.FindColumn(c => !c.Cell(1).IsEmpty()
+                                                    && ClosedXMLValueComparer.DefaultComparer.Compare(c.Cell(1).Value, lookup_value) == 0);
             if (range_lookup && matching_column == null)
             {
                 var first_column = range.FirstColumn().ColumnNumber();
@@ -68,9 +70,15 @@ namespace ClosedXML.Excel.CalcEngine.Functions
                 matching_column = range.FindColumn(c =>
                 {
                     var column_index_in_range = c.ColumnNumber() - first_column + 1;
-                    if (column_index_in_range < number_of_columns_in_range && !c.Cell(1).IsEmpty() && new Expression(c.Cell(1).Value).CompareTo(lookup_value) <= 0 && !c.ColumnRight().Cell(1).IsEmpty() && new Expression(c.ColumnRight().Cell(1).Value).CompareTo(lookup_value) > 0)
+                    if (column_index_in_range < number_of_columns_in_range
+                        && !c.Cell(1).IsEmpty()
+                        && ClosedXMLValueComparer.DefaultComparer.Compare(c.Cell(1).Value, lookup_value) <= 0
+                        && !c.ColumnRight().Cell(1).IsEmpty()
+                        && ClosedXMLValueComparer.DefaultComparer.Compare(c.ColumnRight().Cell(1).Value, lookup_value) > 0)
                         return true;
-                    else if (column_index_in_range == number_of_columns_in_range && !c.Cell(1).IsEmpty() && new Expression(c.Cell(1).Value).CompareTo(lookup_value) <= 0)
+                    else if (column_index_in_range == number_of_columns_in_range
+                             && !c.Cell(1).IsEmpty()
+                             && ClosedXMLValueComparer.DefaultComparer.Compare(c.Cell(1).Value, lookup_value) <= 0)
                         return true;
                     else
                         return false;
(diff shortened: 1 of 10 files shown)


ClosedXML/Excel/CalcEngine/Functions/Lookup.cs
```diff
@@ -242,9 +250,15 @@ private static object Vlookup(List<Expression> p)
                 matching_row = range.FindRow(r =>
                 {
                     var row_index_in_range = r.RowNumber() - first_row + 1;
-                    if (row_index_in_range < range.RowsUsed().Count() && !r.Cell(1).IsEmpty() && new Expression(r.Cell(1).Value).CompareTo(lookup_value) <= 0 && !r.RowBelow().Cell(1).IsEmpty() && new Expression(r.RowBelow().Cell(1).Value).CompareTo(lookup_value) > 0)
+                    if (row_index_in_range < range.RowsUsed().Count()
```
Pankraty: I have a strong suspicion, `range.RowsUsed().Count()` will be called on every iteration in `FindRow` which is not very efficient. Better calculate the value outside the lambda expression.

ClosedXML/Excel/CalcEngine/Functions/Lookup.cs
```diff
@@ -242,9 +250,15 @@ private static object Vlookup(List<Expression> p)
...
+                        && !r.Cell(1).IsEmpty()
+                        && ClosedXMLValueComparer.DefaultComparer.Compare(r.Cell(1).Value, lookup_value) <= 0
+                        && !r.RowBelow().Cell(1).IsEmpty()
+                        && ClosedXMLValueComparer.DefaultComparer.Compare(r.RowBelow().Cell(1).Value, lookup_value) > 0)
                         return true;
-                    else if (row_index_in_range == range.RowsUsed().Count() && !r.Cell(1).IsEmpty() && new Expression(r.Cell(1).Value).CompareTo(lookup_value) <= 0)
+                    else if (row_index_in_range == range.RowsUsed().Count()
```
Pankraty: And here it is called again

ClosedXML/Excel/Patterns/ClosedXMLValueComparer.cs
```diff
@@ -0,0 +1,108 @@
...
+                return true;
+
+            if (y == null || x == null)
+                return false;
+
+            if (ReferenceEquals(x, y))
+                return true;
+
+            if (!x.GetType().Equals(x.GetType()))
+                return false;
```
Pankraty: Is `(byte)1` not equal to `(int)1`?

Pankraty (commented):
Sorry, I don't have enough time for conscious review ATM, will get back to it later (this weekend most likely).

ClosedXML/Excel/Patterns/ClosedXMLValueComparer.cs
igitur: `IXLCell.Value` can only one of these types:

- Double
- DateTime
- TimeSpan
- String

igitur:
> Sorry, I don't have enough time for conscious review ATM, will get back to it later (this weekend most likely).

No rush. This is targeted for a post 0.95 release.

ClosedXML/Excel/Patterns/ClosedXMLValueComparer.cs
igitur: But I do see that this differs to the `Compare()` method. This should be fixed, definitely.

ClosedXML/Excel/Patterns/ClosedXMLValueComparer.cs
```diff
@@ -0,0 +1,108 @@
...
+                }
+                catch (InvalidCastException) { return -1; }
+                catch (FormatException) { return -1; }
+                catch (OverflowException) { return -1; }
+                catch (ArgumentNullException) { return -1; }
+            }
+
+            // String comparisons should be case insensitive
+            if (c1 is string s1 && c2 is string s2)
+                return StringComparer.OrdinalIgnoreCase.Compare(s1, s2);
```
Pankraty: `_stringComparer.Compare`

ClosedXML/Excel/Patterns/ClosedXMLValueComparer.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            if (x is double dbl1 && y is double dbl2)
+                return Math.Abs(dbl1 - dbl2) < double.Epsilon;
+
+            return x.Equals(y);
+        }
+
+        public int GetHashCode(object obj)
+        {
+            return obj.GetHashCode();
```
Pankraty: Check for null?

ClosedXML/Excel/Patterns/ClosedXMLValueComparer.cs
```diff
@@ -0,0 +1,108 @@
...
+                return false;
+
+            if (ReferenceEquals(x, y))
+                return true;
+
+            if (!x.GetType().Equals(x.GetType()))
+                return false;
+
+            if (x is string s1 && y is string s2)
+                return _stringComparer.Equals(s1, s2);
```
Pankraty: String comparison also differs: in `Compare()` method it is case-insensitive, here it relies on `_stringComparer`.

Pankraty (commented):
Is this me obsessed with unit tests, or are they really begging to be written? 😆

igitur:
Yup, consider this PR still work in progress.

ClosedXML/Excel/CalcEngine/Functions/Lookup.cs
igitur: Fixed in #1237

ClosedXML/Excel/CalcEngine/Functions/Lookup.cs
igitur: Fixed in #1237

jahav:
I am closing this PR in favor of #1238. Cell value is now represented through `XLCellValue` and thus most of the code it no longer relevant and the rest is part of the rebased #1238 anyway.

PR #1234, opened 2019-06-13
Merged ranges cell values should not be set, except for top left cell of merged range

### LINKED ISSUE #1229: Contents of merged cells are not cleared if they are filled after the merge
Huh. I've rebased my sparklines PR and several tests fail. One of them fails due to this bug

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

`develop`, after #1224 is merged

```
ws1.Range("A2:A4").Merge().SetValue("Linear, Colorful 1, All markers, SameForAll scale");
```

This piece of code assigns a value to cells A2, A3 and A4. Previously, we would clear cells A3 and A4 during the loading. We could also do the merge _after_ the filling:
```
ws1.Range("A2:A4").SetValue("Linear, Colorful 1, All markers, SameForAll scale").Merge();
```
In both cases, A3 and A4 would end up empty.

But now, with the optimized loading, they have the value. We can even unmerge them in Excel:
image

Merged cells (except the first one) must remain empty no matter if we merge them after filling the value or the opposite - fill after they've been merged.
This is true for styling too: we should exclude merged cells from style propagation unless the entire merged range is affected. _(I have to provide more sufficient code sample for this, I know)_

Yes, between #1031 and #1224 

## Reproducibility
```c#
public void Main()
{
            var ws1 = workbook.AddWorksheet();
            ws1.Range("A2:A4").Merge().SetValue("Test");
           Console.WriteLine(ws1.Cell("A2").Value);// Should be Test
           Console.WriteLine(ws1.Cell("A3").Value);// Should be empty. Actual: Test
}
```

[SampleSparklines.xlsx](https://github.com/ClosedXML/ClosedXML/files/3258715/SampleSparklines.xlsx)
[SampleSparklines_Before1224.xlsx](https://github.com/ClosedXML/ClosedXML/files/3258748/SampleSparklines_Before1224.xlsx)

igitur:
Fixes #1229

### DIFF
--- a/ClosedXML_Tests/Excel/Ranges/MergedRangesTests.cs
+++ b/ClosedXML_Tests/Excel/Ranges/MergedRangesTests.cs
@@ -324,7 +324,6 @@ namespace ClosedXML_Tests
                 Assert.AreEqual(XLBorderStyleValues.None, ws.Cell("D2").Style.Border.BottomBorder);
                 Assert.AreEqual(XLBorderStyleValues.None, ws.Cell("D2").Style.Border.LeftBorder);
 
-
                 Assert.AreEqual(XLBorderStyleValues.None, ws.Cell("B3").Style.Border.TopBorder);
                 Assert.AreEqual(XLBorderStyleValues.None, ws.Cell("B3").Style.Border.RightBorder);
                 Assert.AreEqual(XLBorderStyleValues.None, ws.Cell("B3").Style.Border.BottomBorder);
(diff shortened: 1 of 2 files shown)


ClosedXML_Tests/Excel/Ranges/MergedRangesTests.cs
```diff
@@ -357,5 +355,17 @@ public void UnmergedCellsPreserveStyle()
...
+        [Test]
+        public void MergedRangesCellValuesShouldNotBeSet()
+        {
+            using (var workbook = new XLWorkbook())
+            {
+                var ws = workbook.AddWorksheet();
+                ws.Range("A2:A4").Merge().SetValue(1);
+                ws.Cell("B1").FormulaA1 = "SUM(A:A)";
+                Assert.AreEqual(1, ws.Cell("B1").Value);
+            }
```
Pankraty:
```
ws.Range("A2:A4").Merge();
ws.Cell("A2").FormulaA1 = "=1";
ws.Cell("A3").FormulaA1 = "=1";
ws.Cell("A4").FormulaA1 = "=1";
ws.Cell("B1").FormulaA1 = "SUM(A:A)";
Assert.AreEqual(1, ws.Cell("B1").Value);
```
image

Pankraty:
Beside to cell value and formulas, I think we have to also process styles, comments, sparklines (once they're implemented), maybe something else. But this, of course, has a lower priority.

igitur:
Urgh, how did I forget about formulas! Of course!

I'm not convinced about the others though. Would keeping styles, etc have any effect on other dependent cells? Values and formulas affect dependent cells, so we definitely need exclusion for them.

ClosedXML_Tests/Excel/Ranges/MergedRangesTests.cs
igitur: Fixed.

PR #1250, opened 2019-07-13
WIP Pivot table formats improved

igitur:
@b0bi79 I'm trying to see whether I can enhance the current pivot table formats implementation to address your needs. #1222 reverses many of the changes we did when we initially implemented pivot table formats, and I would like to avoid so many reversals. Please attach more files with other pivot table formats so that we can identify all the possible problems.

This PR is still WIP and some tests still fail.


b0bi79:
For now, try with these three files, if you succeed, I will give more.
[tPivot11.xlsx](https://github.com/ClosedXML/ClosedXML/files/3389178/tPivot11.xlsx)
[tPivot31.xlsx](https://github.com/ClosedXML/ClosedXML/files/3389179/tPivot31.xlsx)
[tPivot51.xlsx](https://github.com/ClosedXML/ClosedXML/files/3389180/tPivot51.xlsx)

lionsoft:
[CustomPivotTheme.xlsx](https://github.com/lionsoft/ClosedXML/raw/lion_pivot_tables_styling/ClosedXML_Tests/Resource/Other/Lion/PivotTables/CustomPivotTheme.xlsx)

[PivotWithStyles.xlsx](https://github.com/lionsoft/ClosedXML/raw/lion_pivot_tables_styling/ClosedXML_Tests/Resource/Other/Lion/PivotTables/PivotWithStyles.xlsx)

jahav:
@igitur I have been thinking about how to deal with the pivot table styles. I took a loot at Aspose.Cells and EPPlus capabilities of pivot table format and neither have full expressivness of pivot style.

I want to achieve two goals:
1. Read and write arbitrary pivot styles. That means I have to store full selection rule (`PivotArea` + `References` elements). This is basically required to be able to read and write pivot table styles from other producers. The common pattern of use is that there is a template with pivot table and ClosedXML fills data to another sheet. To keep styles, we need full expressivness of the format. 
2. I want to keep current API. The current API of ClosedXML is of course also rather limited, but I have no desire to do API design at this time. That is just scope creep and thus not aligned with my target of 1.0. Minimal changes to current API. I made a wiki documentaiton page about pivot styles a year ago about styles and although the API has limitations, it's sufficient for reasonable work https://github.com/ClosedXML/ClosedXML/wiki/Pivot-Table-Styles

I think best way forward is to store `Selection Rule -> Style` directly in the `XLPivotTable` and use it across `IXLPivotStyleFormat`s in the API. The current approach is limited, but if properly separated, can work without much brekage.

By separation I mean that individual styles will expose only subset of all `Selection Rule -> Style`. E.g. `IXLPivotTables.StyleFormats.ColumnGrandTotalFormats` (IXLPivotStyleFormats) will only contain format that aren't attached to a field and are used for column grand total.

What do you think?

jahav:
Closing this PR. Recent pivot rework (e.g. #2386) reimplemented the pivot structure and styles and this is no longer applicable.

As far as I can see, this improves load/save of pivot areas with subtotals (mostly). The pivot areas should be full representable by `XLPivotArea` and subtotals in particular by `XLPivotReference.Subtotals`.

igitur:
Agree. Thanks again for the good work. Pivot tables were indeed a mess.

PR #1258, opened 2019-07-24
ReadTheDocs Proof of Concept

KnapSac:
For now I have created a very simple getting started page, as well as a Hello World tutorial as an introduction to ClosedXML.


KnapSac:
As you can see, the build is failing because of the example files. I suggest we find a way to either exclude them from the build, or find another way to fix this issue.

The reason I organised the example files this way, is because with ReadTheDocs/Sphinx it is possible to automatically show the difference between two files, and I believe this is useful in the tutorial.

Pankraty:
@KnapSac thanks for the contribution!

In order to exclude some of the *.cs files from a build process, you can specify Build Action = None in properties:
image

Or even better, you can specify excluded files using a wild card in .csproj, like this
image

KnapSac:
Allright, thanks, I didn't know that. I'll commit a fix.

KnapSac:
My apologies, I forgot to build the project before committing, but now it is fixed.

igitur:
Just noticed that you're using reStructuredText instead of MarkDown. Didn't know readthedocs support both. Anyone got a summary of the benefits of each?

igitur:
@KnapSac Please create a new branch and push to it. Don't work on `develop` directly. It will just possibly create problems in the future. I'll try to link our readthedocs instance to your new branch, once set up, so that you can see you changes online. Thanks for your help.

igitur:
@KnapSac What is your readthedocs.io username? Also KnapSac?

KnapSac:
Yes, and I'll push to the other branch.

KnapSac:
Here is a good article I read about the differences between MarkDown and reStructuredText: https://www.ericholscher.com/blog/2016/mar/15/dont-use-markdown-for-technical-docs/

What it comes down to is that MarkDown has a lot of different versions, which aren't compatible. Also, reStructuredText has both inline and block markup, as well as offering semantic meaning. This is something which is lacking in MarkDown. Another very convenient feature is the option to automatically diff two files, and you can also include files, which means you don't have to write your example code in your doc pages.

PR #1259, opened 2019-07-26
WIP Readthedocs POC Doxygen integration

KnapSac:
When looking for an easy way to create the documentation, I stumbled upon Doxygen. In combination with Breathe, it is quite doable to create documentation from the existing XML comments in the source files. I think this could be very useful as a reference (see the cell page as an example, although we should probably not show everything at once) in addition to handmade tutorials to really familiarize the user with ClosedXML.


igitur:
Until this PR is merged, I would like the rtd to integrate with your repo, i.e. when you push a change, that rtd automatically builds the documentation. 

Please read the [rtd webhooks documentation](https://docs.readthedocs.io/en/stable/webhooks.html) and use this URL to enable the hook: https://readthedocs.org/api/v2/webhook/closedxml/66647/

igitur:
At the moment, the doc building is failing with this output:
image

KnapSac:
Locally I had to pip install sphinx_copybutton, but I don't know how that works on Readthedocs. I'll look into it and get back to you.

KnapSac:
I looked around, it seems possible by using a requirements file which pip checks for additional modules to install, I'll add it tonight.

KnapSac:
Everything is setup now, and the docs automatically rebuild when I push to my repo.

igitur:
@KnapSac You must let me know when you think the docs are ready for their first release. I think at least everything in the current wiki should be present on the new RTD site, agree? I haven't checked whether that is the case.

KnapSac:
I agree with you that everything from the current wiki should be present, although the layout and structure won't be the same because I have moved a few things around. If you would prefer to move to RTD soon, I could always temporarily make a page with all the old tutorials. If we were to go that route, I would just remove everything I have created a new tutorial for from the page with the old stuff, until there isn't any old tutorials left. That might be a good idea, but let me know what you think.

igitur:
@KnapSac There's no rush. And I think a restructure is a good idea. Just let me know when you think that your branch is ready for final review. I don't really monitor the commits individually as they come in.

igitur:
@KnapSac And I'll reward you with a postcard of my hometown. I actually owe the other contributors one too still! Sorry, @Pankraty , I'll send yours soon. ;-)

KnapSac:
All right, once I feel like a final review is due, I'll remove the WIP from the title.

KnapSac:
#1393 

After I submitted the issue, I realized this would probably have been a better place to leave that message. But anyway, feel free to either close or merge this PR. I won't be continuing with it.

PR #1260, opened 2019-07-27
WIP Word functionality

KnapSac:
This is a small step towards implementing Word functionality, but I would love to hear your thoughts about it.


igitur:
If we start doing Word support, then I think the library must be split up into `ClosedXML.Core` for shared functions, `ClosedXML.Excel` and `ClosedXML.Word`. @Pankraty, agree?

Pankraty:
I agree, in general. But what functionality can be shared between those two? They seem so different...

KnapSac:
I'm also not sure what functions can be used by both, but one possible issue I was thinking about was styling. If we want to implement styling for Word, I don't think we can use IXLStyle and related things. Should I name the styling differently, or do you have other ideas about that? Naming it something along the lines of IXLDocStyle is possible, but I don't know if that is the best way to go.

igitur:
> I agree, in general. But what functionality can be shared between those two? They seem so different...

Mostly the helper functions, e.g. dealing with reading and writing OpenXML will be shared. If we did it correctly, it should be everything outside the `ClosedXML.Excel` namespace in the current codebas.e

igitur:
@KnapSac Please have a look at https://github.com/ClosedXML/ClosedXML/blob/develop/CONTRIBUTING.md for style guides, etc.

Pankraty:
There is also a drawback. While we have a single library we can define most of the helpers as internal classes and be more reluctant in their API and backward compatibility. Once we extract this in Core assembly we have to make it all public increasing the amount of APIs to support.

BTW, I've been working on a read-only API for faster reading. It is going to read just a necessary data on demand. If I succeed this part can also be shared as I am trying to do it quite generic. But I am on the very beginning of the long path.

KnapSac:
Are you using the default CodeMaid settings? Because I use CodeMaid, but I may have changed a couple settings, which I can change back ofcourse.

igitur:
Yes, I use the default settings.

KnapSac:
Oke, will do.

igitur:
Maybe we should create a solution-specific CodeMaid  config file.

igitur:
> There is also a drawback. While we have a single library we can define most of the helpers as internal classes and be more reluctant in their API and backward compatibility. Once we extract this in Core assembly we have to make it all public increasing the amount of APIs to support.
> 
> BTW, I've been working on a read-only API for faster reading. It is going to read just a necessary data on demand. If I succeed this part can also be shared as I am trying to do it quite generic. But I am on the very beginning of the long path.

What about the 'make internals visible to' approach that we use for the tests project?

ClosedXML/Word/XLDocStyle.cs
```diff
@@ -0,0 +1,62 @@
+﻿using DocumentFormat.OpenXml.Packaging;
+using DocumentFormat.OpenXml.Wordprocessing;
+
+namespace ClosedXML.Word
+{
+    public class XLDocStyle : IXLDocStyle
```
igitur: In general, try not to abbreviate, i.e. use `XLDocumentStyle` instead of `XLDocStyle`.

igitur:
@KnapSac A tip... use the [OpenXML Productivity Tools](https://www.microsoft.com/en-us/download/details.aspx?id=30425) to reflect the code from a given Word document. You can then use this general reflected code for you saving the document to file. This was the approach taken by the Excel side originally.

ClosedXML/Word/XLDocStyle.cs
KnapSac: Okay, will do.

b0bi79:
I believe that it will be correct to separate ClosedXML.Excel and ClosedXML.Word into different libraries. The reasons for this:
- these libraries do not have common code;
- ClosedXML is already more than 1 Mb in size, and it will double with Word, which will lead to a needless increase in the size of client product distributions and an increase in the load of the application as a whole;
- When using Excel functionality, Word functionality is not often needed.

KnapSac:
> I believe that it will be correct to separate ClosedXML.Excel and ClosedXML.Word into different libraries. The reasons for this:
> 
> * these libraries do not have common code;
> * ClosedXML is already more than 1 Mb in size, and it will double with Word, which will lead to a needless increase in the size of client product distributions and an increase in the load of the application as a whole;
> * When using Excel functionality, Word functionality is not often needed.

I agree. If we are going to use this approach, I would propose moving the Word code to a new library, ClosedXML.Word. In my opinion we shouldn't change anything about the existing ClosedXML library, to avoid creating any confusion for users.

KnapSac:
I won't be continuing with this PR. You can close it if you want.

igitur:
I'll rather leave this open, even if indefinitely.

jahav:
As a new maintainer, I am closing this PR. Resources are very scare and focus should be on fixing/improving the Excel functionality. @KnapSac thanks for the work even though it won't be used.

PR #1266, opened 2019-08-05
Fix CONCATENATE for cell references

### LINKED ISSUE #1264: concatenate() regression
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

develop branch, broken in #1166

CONCATENATE function throws CellValueException when passed a list of single cells (e.g. CONCATENATE(A1, B1)

Concatenation of cells should evaluate as it does in Excel.

In Excel, concatenate() on a range will sometimes return a value in the range, but I don't understand the logic. For instance, concatenate(A1:C1) returns B1 value, but A1:B2 throws  #VALUE! (see attached)

Yes, 0.94.2

## Reproducibility

```c#
public void Main()
{
     	using (var xls = new XLWorkbook())
	{
		var sheet = xls.AddWorksheet();
		sheet.Cell("A1").Value = "Hello";
		sheet.Cell("B1").Value = "World";
		
		Console.WriteLine(sheet.Evaluate("=CONCATENATE(\"Hello\", \"World\")"));
		Console.WriteLine(sheet.Evaluate("=CONCATENATE(A1, B1)")); // throws, incorrectly
		Console.WriteLine(sheet.Evaluate("=CONCATENATE(A1:B1)")); // throws, correctly
		
		Console.WriteLine(sheet.Evaluate("=CONCAT(\"Hello\", \"World\")"));
		Console.WriteLine(sheet.Evaluate("=CONCAT(A1, B1)"));
		Console.WriteLine(sheet.Evaluate("=CONCAT(A1:B1)"));
	
	}

}
```
- [ x ] I attached a sample spreadsheet.  (You can drag files on to this issue)

[concat examples.xlsx](https://github.com/ClosedXML/ClosedXML/files/3453973/concat.examples.xlsx)

igitur:
CONCATENATE may accept cell range references as parameters, but its implementation is tricky and we can't support the full scope that Excel supports. See added code comments for clarification. This improves it though.

Fixes #1264

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Functions/Text.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/Text.cs
@@ -84,7 +84,20 @@ namespace ClosedXML.Excel.CalcEngine
             foreach (var x in p)
             {
                 if (x is XObjectExpression objectExpression)
-                    throw new CellValueException("This function does not accept cell ranges as parameters.");
+                {
+                    if (objectExpression.Value is CellRangeReference cellRangeReference)
+                    {
+                        if (!cellRangeReference.Range.RangeAddress.IsValid)
+                            throw new CellReferenceException();
+
+                        // Only single cell range references allows at this stage. See unit test for more details
+                        if (cellRangeReference.Range.RangeAddress.NumberOfCells > 1)
+                            throw new CellValueException("This function does not accept cell ranges as parameters.");
+                    }
+                    else
+                        // I'm unsure about what else objectExpression.Value could be, but let's throw CellReferenceException
+                        throw new CellReferenceException();
+                }
 
                 sb.Append((string)x);
             }
@@ -316,8 +329,8 @@ namespace ClosedXML.Excel.CalcEngine
             bool ignoreEmptyStrings;
             try
             {
-                delimiter = (string) p[0];
-                ignoreEmptyStrings = (bool) p[1];
+                delimiter = (string)p[0];
+                ignoreEmptyStrings = (bool)p[1];
             }
             catch (Exception e)
             {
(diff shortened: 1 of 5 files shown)


ClosedXML/Excel/CalcEngine/Functions/Text.cs
```diff
@@ -84,7 +84,22 @@ private static object Concatenate(List<Expression> p)
...
+                    if (objectExpression.Value is CellRangeReference cellRangeReference)
+                    {
+                        if (!cellRangeReference.Range.RangeAddress.IsValid)
+                            throw new CellReferenceException();
+
+                        // Only 1 column per parameter allowed.
+                        var columnCount = cellRangeReference.Range.RangeAddress.LastAddress.ColumnNumber - cellRangeReference.Range.RangeAddress.FirstAddress.ColumnNumber + 1;
+
+                        if (columnCount > 1)
+                            throw new CellValueException("This function does not accept cell ranges as parameters.");
```
Pankraty: image

igitur: Oh, so the transpose is also valid. 

In the screenshot you pasted, only E3 and F3 would currently be possble. So what should we do? Either return wrong values or throw an exception?

Pankraty: Maybe it is time to implement context-dependent functions? Like `ROW()`, `COLUMN()`, etc.

igitur: Yes, we should. But for now we need to fix this regression. I think we just return `1 text` or `5 text` for each value in the column/row.

Pankraty: I am not convinced that this is a regression as this function never worked with range references as it should. I would prefer the function screaming out loud that it is not fully implemented (=throw an exception) rather than silently returning an _incorrect_ result that might be extremely hard to trace should it cause the error far down the stream.

igitur: The above use-case isn't a regression, but the `CONCATENATE(A1, B1)` case is a regression.

Yes, I now agree with you. Let's throw an exception. I doubt many people use that edge case anyway.

igitur:
Added some `IXLRangeAddress` helper functions too.

ClosedXML_Tests/Excel/Ranges/XLRangeAddressTests.cs
```diff
@@ -289,6 +289,28 @@ public void RelativeRanges()
             Assert.AreEqual("#REF!", rangeAddress.ToString());
         }
 
+        [Test]
+        public void TestSpanProperties()
+        {
+            var ws = new XLWorkbook().AddWorksheet() as XLWorksheet;
+
+            var range = ws.Range("B3:E5");
```
Pankraty: Try also `B5:E3` and `E5:B3`. The results must be the same (now they are not). I suggest using `Normalize()` in the implementation.

igitur:
@Pankraty Happy to merge?

PR #1281, opened 2019-08-28
Update OpenXmlHelper.cs

behardikjoshi:
use of dynamic gives RuntimeInternal compiler exception while binding dynamic operation as c# is type safe language so use of dynamic must be exceptional.

### DIFF
--- a/ClosedXML/Utils/OpenXmlHelper.cs
+++ b/ClosedXML/Utils/OpenXmlHelper.cs
@@ -88,7 +88,47 @@ namespace ClosedXML.Utils
         /// Since these types do not implement a common interface we use dynamic.</param>
         /// <param name="colorCache">The dictionary containing parsed colors to optimize performance.</param>
         /// <returns>The color in ClosedXML format.</returns>
-        private static XLColor ConvertToClosedXMLColor(dynamic openXMLColor, IDictionary<string, Drawing.Color> colorCache )
+        private static XLColor ConvertToClosedXMLColor(this X14.ColorType openXMLColor, IDictionary<string, Drawing.Color> colorCache )
+        {
+            XLColor retVal = null;
+            if (openXMLColor != null)
+            {
+                if (openXMLColor.Rgb != null)
+                {
+                    String htmlColor = "#" + openXMLColor.Rgb.Value;
+                    Drawing.Color thisColor;
+                    if (colorCache?.ContainsKey(htmlColor) ?? false)
+                    {
+                        thisColor = colorCache[htmlColor];
+                    }
+                    else
+                    {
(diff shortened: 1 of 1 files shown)


Pankraty:
Did you experience any issues with this or you just do not like dynamics? This is a private method called from public methods having strongly-typed signatures which makes it quite hard to misuse the method.
Out of two evils - a risk of incorrect usage of the private method or a code duplication - I chose the first. You propose the opposite. So I wonder why.

06Hardik:
It gives Runtime internal compiler exception "unexpected exception while binding dynamic operation" during compilation.

IDE: Microsoft visual studio code insider.
Version: 1.38.0

.net core version 2.1.501

Pankraty:
This is strange. It works fine for me and all the other developers, as well as for the AppVeyor. Try to upgrade .net core SDK and see if the problem has gone.

igitur:
Yes, I think an issue needs to be opened first. Please do so and complete
the full template so that we can reproduce the problem. Thanks.

igitur:
@06Hardik Please provide a reproducible test case that illustrates the compilation problem. Else we'll unfortunately have to close this.

justcode0012:
The usage of dynamics objects is a problem for me too.
I want to use ClosedXML in Microsoft SharePoint. Because of a legacy switch in the web.config the following exception occurs when opening an excel-file: "Dynamic operations can only be performed in homogenous AppDomain."
I use a simple excel file with some tables and colored cells.

See the following stackoverflow entry which targets the same problem:
https://stackoverflow.com/questions/5541143/legacycasmodel-true-and-dynamic-data-operations

igitur:
@justcode0012 Are you saying that you have no ability to remove the `<trust legacyCasModel="true" />` line in your web.config?

justcode0012:
@igitur Not really. Even if it does not breaks other components of Micorosoft SharePoint (which i don't know because SharePoint has a lot of feautures), it negatively affects the official vendor support.
Quote from an blog post:

> You shouldn’t change it because of backward compatibility and dependancy reasons we need that. Disabling this property cause anomalies.
Please be aware meaning of NOT SUPPORTED ,does not mean your SharePoint not work. It may work but  we may not give support unless you rollback your changes or can not forseen any other anomalies .

For the people how are more familiar with SharePoint: This problem affects only classic "farm solutions". Because i'am maintaining/extending an legacy solution i'am affected.

igitur:
@Pankraty In light of this, I think we should revert the use of `dynamic`. This seems to be a genuine case where this is a problem.

Pankraty:
Agree

behardikjoshi:
Thanks for considering the suggestion

Issue #1280, opened 2019-08-28
Memory usage - opening/saving files


clumba:
## Read and complete the full issue template

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.94.2

**What is the current behavior?**

During opening and especially saving files memory usage spikes high even if nothing was edited. Is this normal? My work files is somewhat big - 20k rows, but not that big for a 200/300+ MB RAM for saving process.

[upd] I don't know why, but I can't upload my file, so I used random big xlsx from data.gov: https://data.gov.au/data/dataset/historical-australian-government-contract-data/resource/c7983dca-5286-4e88-8de4-bc421dd156fe

It bigger than my example, but there is same picture.

Also as you can see .Dispose() not freeing memory like wb = null. Seems like a bug.

**What is the expected behavior or new feature?**

Can it be lower? 

**Did this work in previous versions of our tool?  Which versions?**

Yes

## Reproducibility

**Code to reproduce problem:**
```c#
using ClosedXML.Excel;
using System;
using System.IO;
using System.Collections.Generic;
using System.Diagnostics;
using System.Linq;
using System.Text;
using System.Threading.Tasks;

namespace closedxml_memory_usage
{
    class Program
    {
        static void Main(string[] args)
        {
            Process proc = Process.GetCurrentProcess();
            Console.WriteLine("Program started.");
            Console.WriteLine("Memory usage, MB: " + proc.PrivateMemorySize64 / (1024*1024));
            Console.WriteLine();

            var wb = new XLWorkbook(@"d:\20142015fy (1).xlsx");
            proc.Refresh();
            Console.WriteLine("XLWorkbook opened file 12.7MB, 69k rows");
            Console.WriteLine("Memory usage, MB: " + proc.PrivateMemorySize64 / (1024 * 1024));
            Console.WriteLine();


            wb.SaveAs(@"d:\file1.xlsx");
            Console.WriteLine("XLWorkbook file 12.7MB, 69k rows");
            proc.Refresh();
            Console.WriteLine("Memory usage, MB: " + proc.PrivateMemorySize64 / (1024 * 1024));
            Console.WriteLine();


            GC.Collect();
            GC.WaitForPendingFinalizers();
            Console.WriteLine("Running GC");
            proc.Refresh();
            Console.WriteLine("Memory usage, MB: " + proc.PrivateMemorySize64 / (1024 * 1024));
[...]

PR #1284, opened 2019-08-31
Improvements for workbook saving (#1280)

Pankraty:
Here I remove filling `cellsWithoutFormulas` collection which has not been used for last two years (from 56db35624eeaae2ec794e6b4b131f0a22f838a4c). This lowers the memory usage (~147 MB on the test file from #1280) and should speed up saving thanks to removing multiple `c.Address.ToStringRelative()` calls.

Better see the diff in "ignore spaces" mode.

**UPDATE**
Being unable to significantly reduce memory usage I managed to slightly speed up saving. Before the changes it took 63-64 seconds to open and save the workbook; now it takes 55-57 seconds.

### DIFF
--- a/ClosedXML/Excel/XLWorkbook_Save.cs
+++ b/ClosedXML/Excel/XLWorkbook_Save.cs
@@ -956,7 +956,7 @@ namespace ClosedXML.Excel
             bool hasSharedString(IXLCell c)
             {
                 if (c.DataType == XLDataType.Text && c.ShareString || c.HasRichText)
-                    return c.Style.IncludeQuotePrefix || String.IsNullOrWhiteSpace(c.FormulaA1) && (c as XLCell).InnerText.Length > 0;
+                    return (c as XLCell).StyleValue.IncludeQuotePrefix || String.IsNullOrWhiteSpace(c.FormulaA1) && (c as XLCell).InnerText.Length > 0;
                 else
                     return false;
             }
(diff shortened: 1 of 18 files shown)


InusualZ:
Is this PR it's in a merge-able state? I would like to use it on my local copy. Not saying this to rush things or anything.

Another question. Do you guys have a communication method? I would like to ask some trivial question, not worthy of a github issue, just like this one :)

igitur:
@InusualZ It's mergeable, but hasn't been reviewed.

Please log an issue for any trivial question. Others might benefit from it.

ClosedXML/Excel/XLWorkbook_Save.cs
```diff
@@ -6030,7 +6030,7 @@ private static void SetCellValue(XLCell xlCell, XLTableField field, Cell openXml
 
             if (dataType == XLDataType.Text)
             {
-                if (!xlCell.Style.IncludeQuotePrefix && xlCell.InnerText.Length == 0)
+                if (!xlCell.StyleValue.IncludeQuotePrefix && xlCell.InnerText.Length == 0)
```
igitur: Looks like there are some other places too, e.g. https://github.com/ClosedXML/ClosedXML/blob/fb542bfa6d7a8a65a616a11c8b8e1df384ff6316/ClosedXML/Excel/XLWorkbook_Save.cs#L959 , right?

igitur (changes requested):
Let me know if you agree with just that one extra adjustment.

ClosedXML/Excel/XLWorkbook_Save.cs
Pankraty: Agree. Fixed.

igitur (approved):

PR #1289, opened 2019-10-03
Remove unused code and fix ATAN2 tests

Pankraty:
I accidentally found that some of the classes and methods are not used; also, we had two tests without `[Test]` attribute and bodies not fully matching their names.


igitur:
Do you perhaps know at what stage these classes become unused? Or were they always unused?

Pankraty:
`XLColumnParameters` and `XLRowParameters` were used for creating XLColumn and XLRow instance before range repository was introduced. `XLStylizedContainer` aimed to be some kind of generic entity having style but apparently was not of any use.

ClosedXML/Excel/XLWorkbook.cs
```diff
@@ -288,45 +288,6 @@ private void InitializeTheme()
             };
         }
 
-        internal XLColor GetXLColor(XLThemeColor themeColor)
```
igitur: Instead of removing this, isn't it useful and should be made public?

igitur: Although I would maybe refactor it to support for future additions:
```c#
public XLColor GetXLColor(XLThemeColor themeColor)
{
    var propertyName = Enum.GetName(typeof(XLThemeColor), themeColor);
    var xlColor = typeof(XLTheme).GetProperty(propertyName)?.GetValue(this.Theme) as XLColor;
    return xlColor ?? throw new ArgumentException("Invalid theme color");
}
```

igitur: Nevermind, I see there is a `XLTheme.ResolveThemeColor` that already does this.

PR #1290, opened 2019-10-03
WIP: Named styles handling

### LINKED ISSUE #1269: Is it possible to see if cell is based on named style?
In a spreadsheet, I want to allow users to mark certain cell with the built-in "Bad" cell style. When I process the spreadsheet with ClosedXML, I want to be able to detect whether a cell is using this "Bad" style and then take a specific action.

Is it possible to check whether a cell is using a built-in or named style?

Pankraty:
Fixes #1269 

We did not have support for reading or changing named styles. With this PR I added basic operations for managing named styles (reading, changing, copying).

Reading works fine already but I had issues with saving. Once a new named style is created it is not displayed in Excel unless I exclude `BuiltinId = 0U`. But even with this the style is displayed in a list but does not have formatting. 

I am going to get back to this later, but maybe someone has ideas about how to fix this.


Pankraty:
Ah, only after I submitted a pull request I realized that the property was not `BuildIn = 0` (= **not** built-in) but `BuildInId` (=ID of the built-in style). This explains the first part of the "weird" behavior XD

TimeJunkie:
The ability to add named styles is something I have been looking all over the internet to see how to add in ClosedXML. This feature would be nice to have for sure as I have been working with a project that uses Excel interop COM libraries that does just this; however, I would like to migrate all that over to using ClosedXML instead for faster reads and writes and less memory overhead.

PR #1194, opened 2019-04-29
Fix for issue 1192

UnlawfulCitizen:
#### What's this PR do?
Fix for issue ClosedXML#1192
#### Where should the reviewer start?
ClosedXML/Excel/Cells/XLCell.cs Line 679
#### How should this be manually tested?
By following the example in the Issue
#### Questions:
- Is there a blog post? No 
- Does the knowledge base need an update? Maybe?
- Does this add new (C#) dependencies? No 

- [ ] C# Code Review: @csreviewer
- [ ] Test Automation Review: @csreviewer

Closes #1192

PR #1292, opened 2019-10-04
Redefine rules for sheet name escaping (#1287)

### LINKED ISSUE #1287: SetRowsToRepeatAtTop does not work on worksheets which names require escaping
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.94.2

Using:

`ws.PageSetup.SetRowsToRepeatAtTop(1, row);`

The values that are expect to be populated in the Page Setup screen are not being populated:

image

And the behaviour is not as expected. 

1) The fields in the highlighted field above should be populated.
2) In print view, the data headers should be repeated like so:
image

Not sure- first time using ClosedXML

## Reproducibility

```c#
{
    File.WriteAllBytes("Document.xlsx", GetReport_Excel());
}

static DataTable GetTable()
{
    // Here we create a DataTable with four columns.
    DataTable table = new DataTable();
    table.Columns.Add("Dosage", typeof(int));
    table.Columns.Add("Drug", typeof(string));
    table.Columns.Add("Patient", typeof(string));
    table.Columns.Add("Date", typeof(DateTime));

    for (int i = 0; i < 400; i++)
    {
        table.Rows.Add(25, "Indocin", "David", DateTime.Now);
        table.Rows.Add(50, "Enebrel", "Sam", DateTime.Now);
        table.Rows.Add(10, "Hydralazine", "Christoff", DateTime.Now);
        table.Rows.Add(21, "Combivent", "Janet", DateTime.Now);
        table.Rows.Add(100, "Dilantin", "Melanie", DateTime.Now);
    }
    return table;
}


public static byte[] GetReport_Excel()
{
    //Generated Datatable
    DataTable dt = GetTable();

    #region Excel Stuff
    var columns = dt.Columns.Count;
    DateTime date = DateTime.Now;

    List<string> locations = new List<string> { "A", "B", "C" };
    
    var ws = wb.Worksheets.Add(date.ToString("MM-dd-yyyy"));
    int row = 1;
    //Build titles
    ws.Cell(row, 1).Value = "All Patient Visits on " + date.ToString("MM-dd-yyyy");
    IXLRange headerRange = ws.Range(row, 1, row, columns).Merge().AddToNamed("Title");
    var headerHight = FormatTitle(ref headerRange, 16);
    row++;

    ws.Cell(row, 1).Value = "Resources: All resources";
    headerRange = ws.Range(row, 1, row, columns).Merge().AddToNamed("Resources");
    FormatTitle(ref headerRange);
    row++;

    ws.Cell(row, 1).Value = "Location: " + string.Join(", ", locations);
    headerRange = ws.Range(row, 1, row, columns).Merge().AddToNamed("Location");
    FormatTitle(ref headerRange);
    row++;

    //Insert data
    ws.Cell(row, 1).InsertTable(dt);


    //Format data table
    var firstTable = ws.Tables.FirstOrDefault();
    if (firstTable != null)
    {
        firstTable.Theme = XLTableTheme.TableStyleMedium18;
        firstTable.CellsUsed().Style.Alignment.WrapText = true;
        //firstTable.Sort("Time ASC, Location ASC");
    }

    //For some reason, these adjust actions do not account for visible overflows with larger text, sow e have to set the height again after the fact.
    //ws.Columns().AdjustToContents();
    //ws.Rows().AdjustToContents();
    ws.Row(1).Height = headerHight;

    ws.PageSetup.SetRowsToRepeatAtTop(1, row);
    ws.PageSetup.PagesWide = 1;
    #endregion

    //We export it as a byte array for users to download.
    using (MemoryStream fs = new MemoryStream())
    {
        wb.SaveAs(fs);
        fs.Position = 0;
        return fs.ToArray();
    }
}

public static int FormatTitle(ref IXLRange cells, int size = 12)
{
    cells.Style.Font.Bold = true;
    cells.Style.Font.FontSize = size;
    double idealHeight = size * 1.3125;
    return (int)Math.Floor(idealHeight);
}
```

I am aware of the weird conversion to byte[] then back to file process I am using. I ported over code that usually calls for a byte[]. Not the point anyways lol. 

- [ ] I attached a sample spreadsheet.  

No files to be included. Use the above code to generate one.

Pankraty:
Fixes #1287 
Depends on #1194 ~(will do a rebase after it is merged)~

Here I joined together two classes both called `StringExtensions`, made all their methods internal (classes were internal anyway), rewrote the rules for sheet name escaping and provided a bunch of tests.


ClosedXML/Extensions.cs
```diff
@@ -132,6 +132,40 @@ public static String ToProper(this String value)
 
             return value.Substring(0, 1).ToUpper() + value.Substring(1);
         }
+
```
igitur: I agree the 2 `StringExtensions` classes should be merged, but I still prefer to have them in a separate file. Don't you agree?

igitur:
How does this depend on #1294 ? Did you perhaps mean another PR?

Pankraty:
Ah, it's #1194, apparently. This PR was based on commits from that one.

ClosedXML/Extensions.cs
Pankraty: Yeah, that makes sense.

igitur:
#1194 is merged. Please rebase.

ClosedXML/Extensions.cs
Pankraty: I've extracted all extension classes in separate files and put them in the Extensions folder.

ClosedXML_Tests/Excel/Misc/ExtensionsTests.cs
```diff
@@ -33,5 +33,55 @@ public void DoubleValueSaveRound()
...
+        [TestCase("R1C1", ExpectedResult = "'R1C1'")]
+        [TestCase("A{", ExpectedResult = "'A{'")]
+        [TestCase("A}", ExpectedResult = "'A}'")]
+        [TestCase("A`", ExpectedResult = "'A`'")]
+        [TestCase("Русский", ExpectedResult = "Русский")]
+        [TestCase("日本語", ExpectedResult = "日本語")]
+        [TestCase("한국어", ExpectedResult = "한국어")]
+        [TestCase("Slovenščina", ExpectedResult = "Slovenščina")]
+        [TestCase("", ExpectedResult = "")]
+        [TestCase(null, ExpectedResult = null)]
```
igitur: Very thorough!

PR #1300, opened 2019-10-26
Grow method now uses default value

carlosproiete:
I removed the Grow Method with no parameters, that returned a call to an overload of Grow with a Int parameter. Now there is only one grow method that takes an optional parameter.


Pankraty:
Thanks for the contribution! 

But if you want my opinion, there is no very much use in removing the existing method since it is a breaking change. The code will still compile and the behavior will remain the same but the binary compatibility will be broken for no benefit. Here are more details on the subject https://codeblog.jonskeet.uk/2018/04/13/backward-compatibility-and-overloading/

igitur:
Going to close this, but thanks for your contribution anyway.

PR #1301, opened 2019-10-28
Added GrowRow()

carlosproiete:
Hello! I added a new method called GrowRow(int), to grow the selection not only in both ways, but downwards.

### DIFF
--- a/ClosedXML/Excel/Ranges/IXLRangeBase.cs
+++ b/ClosedXML/Excel/Ranges/IXLRangeBase.cs
@@ -321,6 +321,13 @@ namespace ClosedXML.Excel
         IXLRangeBase Grow(Int32 growCount);
 
         /// <summary>
+        /// Grows this the current range by the specified number of rows.
+        /// </summary>
+        /// <param name="growRowCount">The grow count.</param>
+        /// <returns></returns>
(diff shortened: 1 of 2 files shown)


igitur:
Please add some unit tests. Look for the unit tests for `Grow()` for an example.

ClosedXML/Excel/Ranges/IXLRangeBase.cs
```diff
@@ -320,6 +320,13 @@ public interface IXLRangeBase : IXLAddressable
         /// <returns></returns>
         IXLRangeBase Grow(Int32 growCount);
 
+        /// <summary>
+        /// Grows this the current range by the specified number of rows.
+        /// </summary>
+        /// <param name="growRowCount">The grow count.</param>
+        /// <returns></returns>
+        IXLRangeBase GrowRow(Int32 growRowCount);
```
igitur, line 328: I think `GrowDown` is a better name. Agree?

igitur (changes requested):
Please also add unit tests.

jahav:
Closing without merge.

I don't think it's worth expanding API surface for this. The proposed method could easily be implemented in client code as an extension method. Also, there are no unit tests and semantic is unclear, doc says that it expands the current range, but method returns a new one without expanding the original one.

PR #1305, opened 2019-11-05
Added support for DateTimeOffset to SetKnownTypedValue

mathlang:
Added support for DateTimeOffset to XLCell.SetKnownTypedValue. The DateTimeOffset is converted to a DateTime in the local time zone.


igitur:
Thanks for the contribution. But can you please explain why this is required? Currently the user can just do the conversion to `DateTime` self and set the resulting value.

mathlang:
I encountered this when trying to write a DataTable from a database using InsertTable. I figured that since DateTime is supported it would make sense to support DateTimeOffset.

I understand that I could write it manually and do the conversion to DateTime but it defeats the purpose of InsertTable in that use case.

igitur:
Ah ok, that makes more sense. For `InsertTable` we want to support all data types that are natively supported by SQL. Is `DateTimeOffset` supported by SQL? And if so, please add unit tests for `InsertTable`, which is the root issue here.

CoenraadS:
> Ah ok, that makes more sense. For `InsertTable` we want to support all data types that are natively supported by SQL. Is `DateTimeOffset` supported by SQL? And if so, please add unit tests for `InsertTable`, which is the root issue here.

`DateTimeOffset` is supported by SQL: https://www.w3schools.com/sql/sql_datatypes.asp

igitur:
Fixed in #1705

PR #1311, opened 2019-11-21
SUBTOTAL formula calculation fix

jbzorg:
**What is it:**
Fix of the problem with calculation of the SUBTOTAL formula.

**Description of the problem:**
When calculating the SUBTOTAL formula, ALL cells of the range were used. However, cells with the SUBTOTAL formulas must be excluded from the range.

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Functions/MathTrig.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/MathTrig.cs
@@ -5,6 +5,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 
 namespace ClosedXML.Excel.CalcEngine
 {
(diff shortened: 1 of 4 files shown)


ClosedXML_Tests/Excel/Loading/LoadingWithSubtotalTests.cs
```diff
@@ -0,0 +1,40 @@
...
+                Assert.AreEqual(24, wb.Worksheet(1).Cell(6, "H").Value);
+                Assert.AreEqual(254, wb.Worksheet(1).Cell(8, "H").Value);
+                Assert.AreEqual(192, wb.Worksheet(1).Cell(9, "H").Value);
+
+                Assert.AreEqual(6, wb.Worksheet(1).Cell(3, "J").Value);
+                Assert.AreEqual(24, wb.Worksheet(1).Cell(6, "J").Value);
+                Assert.AreEqual(254, wb.Worksheet(1).Cell(8, "J").Value);
+                Assert.AreEqual(192, wb.Worksheet(1).Cell(9, "J").Value);
+                Assert.AreEqual(510, wb.Worksheet(1).Cell(13, "J").Value);
+            }
```
Pankraty: The test may be improved greatly if fill cells with formulas explicitly rather than open the file. In that case, it would be easier to see what cases are covered.

E.g.
```
var wb = new XLWorkbook();
var ws = wb.AddWorksheet();
ws.Cell("A1").Value = 1;
ws.Cell("B1").Value = 2;
ws.Cell("A2").Formula = "SUBTOTAL(9, A1:B1)";
ws.Cell("B2").Formula = "SUBTOTAL(9, A1:A2)";

Assert.AreEqual(3, ws.Cell("A2").Value);
Assert.AreEqual(1, ws.Cell("B2").Value);
// etc.

File: ClosedXML/Excel/CalcEngine/Functions/MathTrig.cs [hunk 2]
```diff
@@ -927,8 +927,17 @@ private static object SqrtPi(List<Expression> p)
 
         private static object Subtotal(List<Expression> p)
         {
+            var expressions = p.Skip(1)
+                .SelectMany(e =>
+                    e is XObjectExpression xoe && xoe.Value is CellRangeReference crr
+                        ? crr.Range.CellsUsed()
+                            .Where(z => !z.FormulaA1?.StartsWith("SUBTOTAL(") ?? true) // except cells with the formula "SUBTOTAL"
```
Pankraty: It is not that easy, unfortunately. 
1. String comparison must be case insensitive.
2. A formula may contain `SUBTOTAL` not only in the beginning but anywhere, for a cell to be excluded from totals. But `string.Contains` would also exclude cells having `SUBTOTALS` included as text (e,g, `=IF(A1="SUBTOTAL()", 10, 50)`). I admit this is an extremely rare case though.

ClosedXML_Tests/Excel/Loading/LoadingWithSubtotalTests.cs
jbzorg: Ok, the xlsx file has been deleted. Now the workbook, data and formulas are created directly in the test.

ClosedXML/Excel/CalcEngine/Functions/MathTrig.cs
jbzorg: Now `SUBTOTAL` is searched using regex anywhere in the formula with case-insensitive. If a substring is found in the formula's text part, it will be ignored.

ClosedXML/Excel/CalcEngine/Functions/MathTrig.cs
```diff
@@ -927,8 +928,22 @@ private static object SqrtPi(List<Expression> p)
...
         private static object Subtotal(List<Expression> p)
         {
+            var expressions = p.Skip(1)
+                .SelectMany(e =>
+                    e is XObjectExpression xoe && xoe.Value is CellRangeReference crr
+                        ? crr.Range.CellsUsed()
+                            .Where(z =>
+                            {
+                                if (!z.HasFormula) return true;
+                                var skipedText = Regex.Replace(z.FormulaA1, @"(""([^""]|"""")*"")", "", RegexOptions.Compiled); // skip text in formula
```
igitur: To avoid instantiation on every iteration, make this regex a `private static readonly` field, similar to the `Regex` fields in `XLCell`.

@Pankraty Agree?

Issue #1310, opened 2019-11-15
ClosedXML has a hard reference to OpenXML. If added to an existing project that doesn't have AutoGenerateBindingRedirects set to true results in a runtime error if OpenXML is updated.


MarkStega:
## Read and complete the full issue template

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug (Documentation lacking)
- [ ] Feature

If you are logging a possible bug or feature request, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.94.2

**What is the current behavior?**

Adding ClosedXML to an existing project that doesn't have
```
<autogeneratebindingredirects>true</autogeneratebindingredirects>
```
will result in a runtime failure if OpenXML has been upgraded to a newer version than 2.7.2

**What is the expected behavior or new feature?**

The need to have autogeneratebindingredirects set to true to allow the explicit reference to OpenXML 2.7.2 to be resolved by later versions needs to be documented.

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

Without a code sample, it is unlikely that your issue will get attention. Don't be lazy. Do the effort and assist the developers to reproduce your problem. Code samples should be [minimal complete and verifiable](https://stackoverflow.com/help/mcve). Sample spreadsheets should be attached whenever applicable. Remove sensitive information.

**Code to reproduce problem:**
```c#
//
// This project demonstrates the hard reference to Open XML2.7.2
// ClosedXML was added to an existing project that did not have generated binding redirects
// Build, run, and fail due to the lack of OXML 2.7.2
//
//  To fix:
//  1) Add the setting <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
//  2) Delete bin directory
//  3) Build, run, succeed due to the binding redirects that get added to the .exe.config
//

namespace BrokenClosedXML
{
    class Program
    {
        private static void Main(string[] args)
        {
            using (var wb = new ClosedXML.Excel.XLWorkbook())
            {
                var ws = wb.AddWorksheet("sheet1");
                ws.Cell("A1").Value = 100;

                wb.SaveAs(@"c:\temp\test.xlsx");
            }
        }
    }
}
```
The project file
```
<?xml version="1.0" encoding="utf-8"?>
[...]

PR #1313, opened 2019-12-02
Update ClosedXML.csproj

jeromep01:
Forced to DocumentFormat.OpenXML 2.9.1 !

### DIFF
--- a/ClosedXML/ClosedXML.csproj
+++ b/ClosedXML/ClosedXML.csproj
@@ -67,7 +67,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="DocumentFormat.OpenXml" Version="2.7.2" />
+    <PackageReference Include="DocumentFormat.OpenXml" Version="2.9.1" />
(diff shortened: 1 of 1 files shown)


Pankraty:
Any particular reasons?

jeromep01:
Yes,

i use this package in many of our librairies (domain+dataaccess) and the
ClosedXML.
I want to follow all packages to the newer version all times, i needed to
upgrade OpenXML.
When i run, i encounter an assembly loader exception due to mismatch
between CloseXML and OpenXML versions (both 2.7.2 and 2.9.1).

Best regards

igitur:
Please see the discussion at #1310 . You should be using the correct binding redirects.

Pankraty:
I am afraid, with this fix many thousand other developers will be forced to upgrade their dependencies too. And I bet they won't be happy with this.

Try to configure binding redirects in your project instead (https://docs.microsoft.com/en-US/dotnet/framework/configure-apps/redirect-assembly-versions).

Pankraty:
@igitur was faster than me :)

PR #1316, opened 2019-12-02
Fix exception when loading rows with no row index populated (#1314)

### LINKED ISSUE #1314: Worksheets where rows do not have the row number attribute set fail to load correctly.

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.95.0-beta2

OpenXML worksheets that have been created without the row index attribute populated on the row elements and without the reference attributes populated on cells fail to load correctly.

Where these attributes (that are optional in the specification) are not populated, the code that loads the spreadsheet fails to set the correct row number on all but the first row. As a result, exceptions are thrown during the load routine and prevent the data from being accessed. There are several third party components that generate files in this way (including SpreadsheetGear).

The issue appears to be due to the lastRow variable being initialised to zero when every single row is read. This means that the index reverts back to zero when the second row in a sheet is loaded and as a result multiple cells are loaded with the same address. This then creates problems when handling the data type of the cell.

The solution appears to be to move the initialisation of the lastRow variable outside of the main loading loop within the Worksheet.

The file should load. Where the reference is missing, ClosedXML should use the row number immediately following that of the last row processed.

No

## Reproducibility

The attached workbook can be loaded to illustrate the problem:

```c#
public void Main()
{
var xl = new XLWorkbook("NoReferencesOnRowsOrCells.xlsx");
}
```
- [x ] I attached a sample spreadsheet.  (You can drag files on to this issue)

[NoReferencesOnRowsOrCells.xlsx](https://github.com/ClosedXML/ClosedXML/files/3912501/NoReferencesOnRowsOrCells.xlsx)

TimGreaves:
I have moved the initialisation of the lastRow outside of the loading loop to ensure that it increments correctly as subsequent rows are loaded in.

Fixes #1314


Pankraty:
Thank you for the contribution!

PR #1319, opened 2019-12-05
Fix number of columns when loading a LibreOffice file

jbzorg:
**What is it:**
Fix of the problem with incorrect number of columns when loading a LibreOffice file.

**Description of the problem:**
LibreOffice and possibly OpenOffice supports a maximum of 1024 columns. Because the algorithm supported only 16,384 columns, ALL 1024 columns were loaded for LibreOffice files. This is the reason for the decline in performance.

### DIFF
--- a/ClosedXML/Excel/XLWorkbook_Load.cs
+++ b/ClosedXML/Excel/XLWorkbook_Load.cs
@@ -2100,8 +2100,10 @@ namespace ClosedXML.Excel
         {
             if (columns == null) return;
 
-            var wsDefaultColumn =
-                columns.Elements<Column>().FirstOrDefault(c => c.Max == XLHelper.MaxColumnNumber);
+            var wsDefaultColumn = columns.Elements<Column>()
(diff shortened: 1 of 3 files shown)


igitur:
This is a good catch, but I dislike having to add a new constant like that. Are there no alternative solutions, possibly deducing the maximum column number from the file elsewhere?

Pankraty:
Huh, I never knew LibreOffice support such few columns.

This is a good catch indeed, but the solution is too particularistic. The file you refer too has a width specified for first 1025 columns. Should we have a file with column width assigned to, say, 16300 columns we will have the same problem.

We have to find a way to avoid the instantiation of columns having nothing but width. For example, store default widths for ranges of columns on a worksheet level (`[1..1025] => 11.52, [1026..16384] => 8.5`) and assign it to a column when it is created. These widths will have to be adjusted on column inserts/deleting, of course, but it looks doable.

Pankraty:
I am closing this in preference to #1373. 

@jbzorg thanks for trying but we'll have to find a better way.

PR #1321, opened 2019-12-15
replaced usage of invariant culture with current culture

lvrush:
Different cultures have different formats for numbers/dates and etc. E.g. thousand separator in english ',' but in russian it's ' '.
I need covert excel file to xml as it is (with culture specified on user machine).

### DIFF
--- a/ClosedXML/Extensions/FormatExtensions.cs
+++ b/ClosedXML/Extensions/FormatExtensions.cs
@@ -11,7 +11,7 @@ namespace ClosedXML.Extensions
             if (!nf.IsValid)
                 return format;
 
-            return nf.Format(o, CultureInfo.InvariantCulture);
+            return nf.Format(o, CultureInfo.CurrentCulture);
         }
     }
 }


Pankraty:
1. I do not understand how ToExcelFormat influences conversion from Excel to XML.
2. Why do you even need XML formatted using current locale? XML is supposed to be machine-readable format on the first place, not human-readable. In order to make it possible to exchange data via XML you'd better make it independent from current user's settings. Otherwise, how would you determine if 1,234 is "one thousand something" or "one something"?

igitur:
I think the better solution is to allow a `CultureInfo` instance to optionally be passed into the `XLWorkbook` constructor, and where applicable, that instance is used.

jahav:
Considering unclear requirements for the change (the ToExcelFormat is an internal method, it mostly is used to display cell content, but also for other things), nonresponse to suggested alternatives and the age, I am closing the issue.

igitur:
Here is a test case. This became relevant last week when I upgraded the dependency on ExcelNumberFormat to v1.1.0. I tend to agree that `cell.GetFormattedString()` should a value formatted in local/current culture. But I also think my comment above about allowing a constructor parameter to override the culture is relevant.

```c#
var ws = wb.AddWorksheet();
var cell = ws.FirstCell();
cell.Value = new DateTime(2019, 4, 30, 17, 45, 0);
Console.WriteLine(cell.GetFormattedString());
```

Currently outputs: `4/30/2019 17:45` and I'm quite averse to that US date format.

Your thoughts, @Pankraty @jahav ?

jahav:
In this case, I think it's better to specify a different culture for whole workbook in ctor (as you suggest) than use CultureInfo.CurrentCulture. 

There are some features that rely on current culture (string to number conversion comes to mind) and having an option to specify it could be nice (e.g workbook saved in one culture and opened in another).

igitur:
If we touch any of this, we should relook `XLPredefinedFormat.FormatCodes` too. Currently that is just a static list of format strings, but it should be made dependent on the culture that we pass in in the XLWorkbook constructor.

@lvrush The easy workaround is of course to set an explicit format, e.g.  `cell.Style.NumberFormat.Format = " # ##0.00";`

Issue #1342, opened 2020-01-29
Invalid customFormat flag on row with customHeight


mathlang:
## Read and complete the full issue template

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

If you are logging a possible bug or feature request, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.95.0-beta2

**What is the current behavior?**

On save, the customFormat="1" attribute is set on <row> elements when the customHeight="1" is set, even if the row does not have a custom format. 

When reading the resulting file, this can cause discrepancies in behavior between ClosedXML and Excel application.

Excel will apply the default worksheet style to the entire row, ignoring column style. 
ClosedXML will ignore the row style and will apply the column style.

**What is the expected behavior or new feature?**

The expected behavior is:

- Properly set the customFormat="1" attribute only when there is actually a custom style applied to the row
- Correctly interpret the style information to match the Excel behavior

**Did this work in previous versions of our tool?  Which versions?**

No

## Reproducibility

1. Run the attached code
2. Open resulting file in Excel
3. Check bold status of cell B2 in Excel

**Code to reproduce problem:**
```c#
public void Main()
{
    using (var wb = new ClosedXML.Excel.XLWorkbook())
            {
                var ws = wb.AddWorksheet("Test");
                ws.Column(2).Style.Font.Bold = true;
                ws.Row(2).Height = 20;
                wb.SaveAs(@"C:\styletest.xlsx");
            }

            using (var wb = new ClosedXML.Excel.XLWorkbook(@"C:\styletest.xlsx"))
            {
                // Closed XML will return true. Cell B2 in Excel will not be bold
                var bold = wb.Worksheet(1).Cell("B2").Style.Font.Bold;
            }
}
```
- [x] I attached a sample spreadsheet.  (You can drag files on to this issue)
[styletest.xlsx](https://github.com/ClosedXML/ClosedXML/files/4129569/styletest.xlsx)

PR #1344, opened 2020-01-30
Changed the cell ping behavior in the Row method to match the Column …

### LINKED ISSUE #1343: Style can be lost when applied to column
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.95.0-beta2

If a style is applied to a column first, and then to a row, the cell at the intersection of the column and the row will not have the column style applied to it.

Applying the style to the row first and then to a column will yield the expected result

Both styles applied to the row and the column should be preserved no matter what order it is applied.

No

## Reproducibility

1. Execute attached code and observe console output

```c#
public void Main()
{
    using (var wb = new ClosedXML.Excel.XLWorkbook())
            {
                var ws1 = wb.AddWorksheet("ColumnFirst");

                ws1.Column(2).Style.Font.SetBold(true);
                ws1.Row(2).Style.Font.SetItalic(true);

                var ws2 = wb.AddWorksheet("RowFirst");

                ws2.Row(2).Style.Font.SetItalic(true);
                ws2.Column(2).Style.Font.SetBold(true);

                // Bold = false and italic = true when style is applied on column first
                Console.WriteLine("When applying style on column first:");
                Console.WriteLine("Bold = {0}", ws1.Cell("B2").Style.Font.Bold);
                Console.WriteLine("Italic = {0}", ws1.Cell("B2").Style.Font.Italic);

                Console.WriteLine();

                // Bold and italic = true when style is applied on row first
                Console.WriteLine("When applying style on row first:");
                Console.WriteLine("Bold = {0}", ws2.Cell("B2").Style.Font.Bold);
                Console.WriteLine("Italic = {0}", ws2.Cell("B2").Style.Font.Italic);

                Console.WriteLine();
                Console.WriteLine("Press any key to continue");
                Console.ReadKey();
            }
}
```
[styletest.xlsx](https://github.com/ClosedXML/ClosedXML/files/4135220/styletest.xlsx)

mathlang:
Fixes #1343 by properly pinging cells of all referenced columns. This matches the way the Column method pings cells of referenced rows.


mathlang:
The failing tests are differences to the <dimension> node of <worksheet>. I'll look into it.

Pankraty:
`HideUnhide` and `Outline` were failing due to the difference in a worksheet dimensions. The newer version is on the right and seems more correct:

image

So I choose to update reference files.

In the `DeleteRanges` test, cells `A1`, `A2`, `B1` now have explicitly defined style. The appearance has not changed, and I am updating the reference file either.

In addition, I've split the `ChangeColumnStyleFirst` test in two, as it was checking two scenarios in fact.

Pankraty:
This PR fixes #1342 as well.

PR #1352, opened 2020-02-28
Fix evaluation of cells referring to themselves

### LINKED ISSUE #1347: How to handle StackOverflowException in ClosedXML

- [x] Bug

**Version of ClosedXML**
0.94.2.0

I have written code to check cell formula of excel cell in C# ClosedXML. Everything works fine, but only problem is, if the checked cell formula contains self cell then it gives StackOverflowException and this exception stops my service.

Please Handle StackOverflowException for self formulated cell.

No

## Reproducibility

```c#
var _xlWorkBook = new ClosedXML.Excel.XLWorkbook(fileStream);
var workSheet = xlWorkBook.Worksheet(1);
var row = workSheet.Row(2);
string cellValue = row.Cell(2).Value.ToString();
```
-
[Test File.xlsx](https://github.com/ClosedXML/ClosedXML/files/4179529/Test.File.xlsx)
 [ ] I attached a sample spreadsheet.  (You can drag files on to this issue)

Pankraty:
Fixes #1347

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -446,11 +446,19 @@ namespace ClosedXML.Excel
             if (Worksheet.Workbook.Worksheets.Contains(sName)
                 && XLHelper.IsValidA1Address(cAddress))
             {
-                var referenceCell = Worksheet.Workbook.Worksheet(sName).Cell(cAddress);
-                if (referenceCell.IsEmpty(XLCellsUsedOptions.AllContents))
-                    return 0;
-                else
-                    return referenceCell.Value;
+                try
+                {
+                    IsEvaluating = true;
+                    var referenceCell = Worksheet.Workbook.Worksheet(sName).Cell(cAddress);
+                    if (referenceCell.IsEmpty(XLCellsUsedOptions.AllContents))
+                        return 0;
+                    else
+                        return referenceCell.Value;
+                }
+                finally
+                {
+                    IsEvaluating = false;
+                }
             }
 
             object retVal;
--- a/ClosedXML_Tests/Excel/Cells/XLCellTests.cs
+++ b/ClosedXML_Tests/Excel/Cells/XLCellTests.cs
@@ -974,6 +974,7 @@ namespace ClosedXML_Tests
             }
         }
 
+        [Test]
         public void FormulaWithCircularReferenceFails2()
         {
             var cell = new XLWorkbook().Worksheets.Add("Sheet1").FirstCell();


ClosedXML_Tests/Excel/Cells/XLCellTests.cs
```diff
@@ -974,6 +974,7 @@ public void InvalidFormulaShiftProducesREF()
             }
         }
 
+        [Test]
```
igitur: 😩

igitur: This was my fault.

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -441,16 +441,22 @@ private object RecalculateFormula(string fA1)
...
-            {
-                var referenceCell = Worksheet.Workbook.Worksheet(sName).Cell(cAddress);
-                if (referenceCell.IsEmpty(XLCellsUsedOptions.AllContents))
-                    return 0;
-                else
-                    return referenceCell.Value;
+            if (Worksheet.Workbook.Worksheets.Contains(sName)
+                && XLHelper.IsValidA1Address(cAddress))
+            {
+                try
```
igitur: I know that `try/catch` blocks severely affect performance. But does `try/finally` without `catch` perform well?

Pankraty: From what I know, the most performance hit is caused by unwinding stack trace so without the exception there should be no difference. But you've raised an interesting question, i'll try to do a little of benchmarking.

Pankraty: https://stackoverflow.com/a/4107320

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -441,16 +441,22 @@ private object RecalculateFormula(string fA1)
...
-                w => String.Compare(w.Name, sName, true) == 0)
-                && XLHelper.IsValidA1Address(cAddress)
-                )
-            {
-                var referenceCell = Worksheet.Workbook.Worksheet(sName).Cell(cAddress);
-                if (referenceCell.IsEmpty(XLCellsUsedOptions.AllContents))
-                    return 0;
-                else
-                    return referenceCell.Value;
+            if (Worksheet.Workbook.Worksheets.Contains(sName)
```
igitur: Please make this change (to use `.Contains` at line 470 too). Bonus points if you can put this change in a separate commit.

igitur: Hmm, I see wee have many cases where we use `String.Compare(String, String, Boolean)` (not only for worksheet names, but also for named ranges). Maybe this warrants a new PR?

Pankraty: I've extracted these changes in a separate commit and replaced all `.Any(x => String.Compare(...))` with the direct access, where applicable.

igitur:
Is it ok if we base this on #1357  ?

Pankraty:
> Is it ok if we base this on #1357 ?

Sure!

igitur:
#1357 has been merged.  Please rebase this on `develop`.

Pankraty:
Done

PR #1353, opened 2020-02-29
Implement % operator and correctly parse expressions with multiple unary operators (--, ++)

### LINKED ISSUE #1324: Unable to use % in formula
- [x] Bug
- [ ] Feature

Version of ClosedXML: 0.94.2.0

A formula like "=A1*A2%" in an Excel cell results in an exception: "ClosedXML.Excel.CalcEngine.ExpressionParseException: Identifier expected. at ClosedXML.Excel.CalcEngine.CalcEngine.Throw(String msg)".

I'm using a code similar to the following, where rawData is of type byte[] containing the Excel file:

```C#
using (var s = new MemoryStream(rawData))
{
    XLWorkbook wb = new XLWorkbook(s);
    var sheet = wb.Worksheet(sheetName);
    var table = sheet.RangeUsed(XLCellsUsedOptions.AllContents).AsTable();
    var dt = table.AsNativeDataTable();

    foreach (DataRow row in dt.Rows)
        Console.WriteLine($"{string.Join(", ", row.ItemArray)}");
}
```

As far as I understand the percentage character (%) can be used in normal cells, so it should also work within formulars, shouldn't it?

Attached file can be used to reproduce the behaviour.

[formular_percent_crash.xlsx](https://github.com/ClosedXML/ClosedXML/files/3972930/formular_percent_crash.xlsx)

Pankraty:
Fixes #1324 

Previously, we parsed % sign only as a part of a numeric literal, but in fact, % in formulas is equivalent to /100 and such expressions as `(100+300)%` or `123%%` are perfectly valid.

I've also noticed that we do not parse expressions with multiple unary operators (e.g. `=---12`) correctly. This is also fixed.

### DIFF
--- a/ClosedXML/Excel/CalcEngine/CalcEngine.cs
+++ b/ClosedXML/Excel/CalcEngine/CalcEngine.cs
@@ -403,11 +403,19 @@ namespace ClosedXML.Excel.CalcEngine
         private Expression ParseUnary()
         {
             // unary plus and minus
-            if (_token.ID == TKID.ADD || _token.ID == TKID.SUB)
+            if (_token.Type == TKTYPE.ADDSUB)
             {
-                var t = _token;
-                GetToken();
+                var sign = 1;
+                do
+                {
+                    if (_token.ID == TKID.SUB)
+                        sign = -sign;
+                    GetToken();
+                } while (_token.Type == TKTYPE.ADDSUB);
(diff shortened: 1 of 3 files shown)


igitur:
Lucky guy! I enjoy working on the parser/tokenizer. But you beat me to it.

Issue #1358, opened 2020-03-02
Expression not converting to double from string sub-expression


mathlang:
**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

If you are logging a possible bug or feature request, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

Latest commit (47f3da53de5241fbb5c7c3095ce5a8f4cecf3940)

**What is the current behavior?**

A sub-expression returning a string (like LEFT or RIGHT) will not convert to a double in an outer expression.

Example:
A1 = "2020"
B1 = RIGHT(A1, 2) + 1

ClosedXML will return B1 = 1. Excel will display B1 = 21.

**What is the expected behavior or new feature?**

ClosedXML should match the Excel behavior and properly handle sub expressions.

Complete this.

**Did this work in previous versions of our tool?  Which versions?**

No

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

Without a code sample, it is unlikely that your issue will get attention. Don't be lazy. Do the effort and assist the developers to reproduce your problem. Code samples should be [minimal complete and verifiable](https://stackoverflow.com/help/mcve). Sample spreadsheets should be attached whenever applicable. Remove sensitive information.

**Code to reproduce problem:**
```c#
public void Main()
{
    // Code standards:
    // - Fully runnable. I should be able to copy and paste this code into a 
    //   console application and run it without having to edit it much.
    // - Declare all your variables (this follows from the previous point)
    // - The code should be a minimal code sample to illustrate issue. The code 
    //   samples on the wiki are good examples of the terseness that I want. Don't
    //   post your full application.

    var workBook = new ClosedXML.Excel.XLWorkbook();
    var workSheet = workBook.AddWorksheet("Test");
    var row = workSheet.Row(1);
    row.Cell(1).Value = "2020";
    var cell = row.Cell(2);
    cell.FormulaA1 = "=RIGHT(A1, 2) + 1";
    var cellValue = cell.Value; // cellValue = 1

    workBook.SaveAs(@"C:\SubExpressionTest.xlsx");
}
```
- [x] I attached a sample spreadsheet.  (You can drag files on to this issue)
[SubExpressionTest.xlsx](https://github.com/ClosedXML/ClosedXML/files/4277739/SubExpressionTest.xlsx)

[...]

PR #1359, opened 2020-03-02
Fixed the Expression double conversion operator to handle string values

mathlang:
Fixes issue #1358

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Expression.cs
+++ b/ClosedXML/Excel/CalcEngine/Expression.cs
@@ -125,6 +125,12 @@ namespace ClosedXML.Excel.CalcEngine
                 return ts.TotalDays;
             }
 
+            // handle string
+            if (v is string s && double.TryParse(s, out var doubleValue))
+            {
+                return doubleValue;
+            }
+
             // handle nulls
             if (v == null || v is string)
             {
--- a/ClosedXML_Tests/Excel/CalcEngine/FunctionsTests.cs
+++ b/ClosedXML_Tests/Excel/CalcEngine/FunctionsTests.cs
@@ -693,5 +693,17 @@ namespace ClosedXML_Tests.Excel.CalcEngine
 
             Assert.AreEqual(expectedResult, res, XLHelper.Epsilon);
         }
+
+        [TestCase("RIGHT(\"2020\", 2) + 1", 21)]
+        [TestCase("LEFT(\"20.2020\", 6) + 1", 21.202)]
+        [TestCase("2 + (\"3\" & \"4\")", 36)]
+        [TestCase("2 + \"3\" & \"4\"", "54")]
+        [TestCase("\"7\" & \"4\"", "74")]
+        public void TestStringSubExpression(string formula, object expectedResult)
+        {
+            var actual = XLWorkbook.EvaluateExpr(formula);
+
+            Assert.AreEqual(expectedResult, actual);
+        }
     }
 }


ClosedXML/Excel/CalcEngine/Expression.cs
```diff
@@ -125,6 +125,12 @@ public virtual Expression Optimize()
                 return ts.TotalDays;
             }
 
+            // handle string
+            if (v is string s && double.TryParse(s, out var doubleValue))
```
Pankraty: Should it be culture specific? Excel, I expect, uses current locale, do we have to do the same? @igitur, your opinion? 
Either way, better to specify locale explicitly and cover such cases with tests. For example, 1,000 is one thousand in en-US and one in ru-RU.

ClosedXML_Tests/Excel/CalcEngine/FunctionsTests.cs
```diff
@@ -693,5 +693,15 @@ public void MultipleUnaryOperators(string formula, double expectedResult)
 
             Assert.AreEqual(expectedResult, res, XLHelper.Epsilon);
         }
+
+        [TestCase("RIGHT(\"2020\", 2) + 1", 21)]
+        [TestCase("LEFT(\"20.2020\", 6) + 1", 21.202)]
+        [TestCase("MID(\"2020.2020\", 3, 6) + 1", 21.202)]
```
Pankraty: There is basically no difference between cases 2 and 3. I'd rather add something like '=2+3&4+5' to also test operators priority.

Pankraty: Or '=0.1+0.2="0.3"'

ClosedXML/Excel/CalcEngine/Expression.cs
mathlang: TryParse uses the current locale by default.

Pankraty: Yeah, I know, but to avoid confusion during possible future refactoring it is nice to have the information about the intention: if default locale was used accidentally or on purpose. In many other places we do not rely on current locale so rare exceptions must be emphasized. IMO

igitur: This relates to a comment I made somewhere in another PR. I think we should optionally pass through the `CultureInfo` instance in the `XLWorkbook` instance and use that everywhere where applicable. Default to `CurrentCulture`. We'll have to do a separate PR for that.

igitur:
Please check whether https://github.com/ClosedXML/ClosedXML/pull/1233 doesn't already address this issue.

mathlang:
It looks like #1233 has the same issue. It's the same code that was moved to a new ConvertibleObject class. I commented in #1233.

mathlang:
I added a few more test cases around the & operator. I had to change the direct cast to double because those formulas resolved to a string type.

ClosedXML_Tests/Excel/CalcEngine/FunctionsTests.cs
mathlang: I added a few more test cases around the & operator. I had to change the direct cast to double because those formulas resolved to a string type.

ClosedXML_Tests/Excel/CalcEngine/FunctionsTests.cs
```diff
@@ -693,5 +693,16 @@ public void MultipleUnaryOperators(string formula, double expectedResult)
 
             Assert.AreEqual(expectedResult, res, XLHelper.Epsilon);
         }
+
+        [TestCase("RIGHT(\"2020\", 2) + 1", 21)]
```
igitur: I would prefer not to explicitly coerce the evaluation results to `double`. I think rather leave the expected value as objects and let the assertions therefore also test the object type. i.e. something like this:

```c#
        [TestCase("RIGHT(\"2020\", 2) + 1", 21)]
        [TestCase("LEFT(\"20.2020\", 6) + 1", 21.202)]
        [TestCase("2 + (\"3\" & \"4\")", "36")]
        [TestCase("2 + \"3\" & \"4\"", "54")]
        [TestCase("\"3\" & \"4\"", 34)]
        public void TestStringSubExpression(string formula, object expectedResult)
        {
            var actual = XLWorkbook.EvaluateExpr(formula);

            Assert.AreEqual(expectedResult, actual);
        }
```
(one more test also included)

Pankraty: The only problem is that without casting to double one cannot specify a tolerance that may cause troubles due to inexact comparison.

igitur: But that doesn't seem to be a problem with these test cases.

Pankraty: OK then

mathlang: @igitur I have made the requested changes to the test.

igitur: @mathlang I don't see the changes.

mathlang: @igitur My bad, not sure what happened. They should be there now.

igitur:
Thanks for a good contribution. Keep them coming.

PR #1362, opened 2020-03-09
Set longText attribute for string items in pivot table

### LINKED ISSUE #1248: Cannot open file with pivot with long text in a source table column
## Cannot open file with pivot with long text in a source table column

- [x] Bug
- [ ] Feature

**Version of ClosedXML**
0.94.2

Cannot open the Excel file with a pivot when long text in a source table column.
Long text seems to be more than 300 characters.

I expect that the file can be opened also when long text are in the source of the pivot.

## Reproducibility

```c#
{
    bool useLongText = true;

    string longText = "1234567890";

    // generate 400 characters
    if (useLongText)
    {
        for (int i = 0; i < 40; i++)
        {
            longText += " 1234567890";
        }
    }

    DataTable dt = new DataTable("test");
    dt.Columns.Add("col1", typeof(string));
    dt.Columns.Add("col2", typeof(int));
    dt.Rows.Add(new object[] { longText, 2 });

    byte[] byteArray = null;

    {
        workbook.Worksheets.Add(dataTable: dt, sheetName: dt.TableName);

        // create the pivot
        IXLWorksheet worksheetPivot = workbook.Worksheets.Add("pivotTest");
        worksheetPivot.PivotTables.Add("Pivot1", worksheetPivot.Cell(1, 1), workbook.Worksheets.ToList()[0].Tables.ToList()[0]);

        using (var exportData = new MemoryStream())
        {
            workbook.SaveAs(exportData);
            exportData.Position = 0;
            byteArray = exportData.ToArray();
        }
    }

    File.WriteAllBytes("test.xlsx", byteArray);
}
```

igitur:
Fixes #1248


ClosedXML/Excel/XLWorkbook_Save.cs
```diff
@@ -2317,6 +2317,9 @@ private static void GenerateTableDefinitionPartContent(TableDefinitionPart table
                     }
                     else
                     {
+                        if (ptfi.DistinctValues.Any(v => ((string)v).Length > 255))
+                            sharedItems.LongText = true;
+
```
Pankraty: Maybe write a unit test?

igitur: Done.

PR #1365, opened 2020-03-10
Fix broken OleDB tests

igitur:
Commit 2beb1c6da2e5379d459b71e729e9acc0c92fd336 in PR #794 broke this.


igitur:
I think we missed this, because the tests are skipped for `netcoreapp2.0` and by default VS2017 ran only `netcoreapp2.0` tests. In VS2019 all frameworks' tests are run.

PR #1368, opened 2020-03-12
Clear inlined rich text data on saving (#1361)

### LINKED ISSUE #1361: Cannot clear a cell in certain XLSX files
- [x] Bug
- [ ] Feature

**Version of ClosedXML**
Tested both 0.94.2 and 0.95.0-beta2

This issue might be related to [this other issue I opened some time ago](https://github.com/ClosedXML/ClosedXML/issues/1330), since it happens with the same kind of files. As a reminder of our discussion in the other thread, these are the XLSX files generated by the Microsoft Reporting Services engine (so files generated from Sequel Service Reporting Services or the Report Viewer C# library, both official Microsoft products).

I have an XLSX file (attached as [test.xlsx](https://github.com/ClosedXML/ClosedXML/files/4019408/test.xlsx) , same file as in the other issue). 

In this file, I cannot clear (set an empty/null value) a cell. If I set an actual value to it it works, but setting an empty string doesn't, the cell remains with filled the old value.

Cell should be emptied when assigning null or empty string to its value

## Reproducibility
This code will reproduce the issue in the attached file. It should empty the first cell, but it doesn't.

```c#
{
    using (var ms = new MemoryStream(File.ReadAllBytes("test.xlsx")))
    {
        using (var workbook = new XLWorkbook(ms))
        {
            var ws = workbook.Worksheets.First();
            var cell = ws.Cell(1, 1);
            cell.Value = "";

            workbook.SaveAs("test_out.xlsx");
        }

    }
}
```

- [x] I attached a [sample spreadsheet](https://github.com/ClosedXML/ClosedXML/files/4019408/test.xlsx)

Pankraty:
Fixes #1361 
Depends on #1392


igitur:
I have to look at this in detail still, but I believe the actual solution was to add full support for inline rich text strings so that loading a file that contains it populates the `IXLCell.RichText` property. And then of course saving too. This PR might be an intermediary solution to reach that goal.

ClosedXML_Tests/Excel/RichText/XLRichStringTests.cs
```diff
@@ -668,5 +669,43 @@ public void ToStringTest()
...
+
+        [Test]
+        public void CanChangeInlinedRichText()
+        {
+            using (var outputStream = new MemoryStream())
+            {
+                using (var inputStream = TestHelper.GetStreamFromResource(TestHelper.GetResourcePath(@"Other\InlinedRichText\inputfile.xlsx")))
+                using (var workbook = new XLWorkbook(inputStream))
+                {
+                    workbook.Worksheets.First().Cell("A1").RichText.AddText("Changed");
```
Pankraty: > I have to look at this in detail still, but I believe the actual solution was to add full support for inline rich text strings so that loading a file that contains it populates the IXLCell.RichText property. 

It's been loaded fine:
image

And then saved too - but as a shared string.

igitur:
I expanded the test a bit to illustrate that additional problems.

igitur:
> I expanded the test a bit to illustrate that additional problems.

I fixed and pushed. Please let me know if you're OK with these changes.

Pankraty:
> I fixed and pushed. Please let me know if you're OK with these changes.

Great! I've started looking into it.

In general, I am fine with this, except now in resulted file we have `sharedStrings.xml` that is empty
image

 We have to handle this too.

Pankraty:
Ah. I did not realize we always had this, it's not something we brought with this PR. In that case, it's OK.

igitur:
I can't assign this to you, because this is 'your' PR ;-) But it's ready for review.

Pankraty:
I cannot approve it due to the same reason :) But it looks good to me.

PR #1369, opened 2020-03-12
Fix issues with loading comments (#1244)

### LINKED ISSUE #1244: input string is not in a correct format

- [ ] Bug
- [x] Feature

**Version of ClosedXML**

0.94.2

I can open this file with Excel with no problems. I can also open it with SpreadsheetDocument spreadsheetDocument = SpreadsheetDocument.Open("TaxPreparation2.xlsm", false); without any errors. However, when I try to open it with ClosedXML I get the exception:

System.FormatException: Input string was not in a correct format.
   at System.Number.ParseDouble(String value, NumberStyles options, NumberFormatInfo numfmt)
   at ClosedXML.Excel.XLWorkbook.GetPtValue(String value)
   at ClosedXML.Excel.XLWorkbook.LoadShapeProperties[T](IXLDrawing`1 xlDrawing, XElement shape)

open the file with no exceptions. i don't have any control over the file creation and I can't make manual edits with Excel before our app processes it so I'm hoping there is a code-based solution to this issue.

This is the only version of the tool I have used

## Reproducibility

```c#
public void Main()
{
   var excelWorkbook = new XLWorkbook("TaxPreparation2.xlsm");
}
```

Note: the file extension was originally XLSM but I had to rename it to ZIP for github to accept it
[TaxPreparation2.zip](https://github.com/ClosedXML/ClosedXML/files/3364138/TaxPreparation2.zip)

Pankraty:
Fixes #1244

In addition, I found that comments visibility was not loaded correctly - this is also fixed.


ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -1401,7 +1401,7 @@ private void LoadClientData<T>(IXLDrawing<T> drawing, XElement clientData)
             LoadDrawingProtection<T>(drawing, clientData);
 
             var visible = clientData.Elements().FirstOrDefault(e => e.Name.LocalName == "Visible");
-            drawing.Visible = visible != null && visible.Value.ToLower().StartsWith("t");
+            drawing.Visible = visible != null && !string.Equals("false", visible.Value, StringComparison.OrdinalIgnoreCase);
```
igitur: This is kind of undoing what was done in e7b0a78f465790aecf9ab0a9f395a3283490f208 . I'm not 100% certain, but it seems this was to support both `true`/`false` and `t`/`f` values.

igitur:
Please rebase.

ClosedXML/Excel/XLWorkbook_Load.cs
igitur: Or do you think we should just continue with your approach? All our tests are passing, so one day, when someone gives us a real test case that fails here, we can add support.

Pankraty: I think I can change this to accept both f and false, but no value has to be treated as true (unlike the previous version) otherwise visible comments disappear.

Pankraty:
Done

PR #1370, opened 2020-03-12
[WIP] Do not consider cell as having RichText unless it has been explicitly modified (#1271)

### LINKED ISSUE #1271: Cell with NumberFormat with thousand separator GetValue returns null

- [x] Bug

**Version of ClosedXML**

0.95.0.0

When reading from a cell with a number format with thousand separator the "GetValue<double?>()" returns null.

On debug I saw that if I inspect the cell with the debug and over RichText property is the cause of the issue. Seems like a locale issue with thousand separator.

Attached a code sample and Excel to reproduce.

Not sure, but at least with the latest stable version 0.94.2 the issue persists.

## Reproducibility

Attached a code full code sample to debug.

[Sample.zip](https://github.com/ClosedXML/ClosedXML/files/3494504/Sample.zip)

Pankraty:
Fixes #1271 

I think this can be tricky for the user to avoid unnecessary RichText initialization with a 100% guarantee and this may lead to errors. At the same time, we have to preserve existing behavior: `XLCell.RichText` should create a new instance of `XLRichText` when accessed if it did not exist.

So I tried to find a balance between two: `XLCell.RichText` does create a new instance but it is considered "default" and does not affect `Value` property until that rich text is modified in some way.

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -1457,7 +1457,11 @@ namespace ClosedXML.Excel
 
         public bool HasRichText
         {
-            get { return _richText != null; }
+            get
+            {
+                return _richText is XLRichText richText &&
+                       !richText.IsDefault;
+            }
         }
 
         IXLComment IXLCell.Comment
--- a/ClosedXML/Excel/RichText/XLFormattedText.cs
+++ b/ClosedXML/Excel/RichText/XLFormattedText.cs
@@ -11,10 +11,12 @@ namespace ClosedXML.Excel
 
         protected T Container;
         readonly IXLFontBase _defaultFont;
+
         public XLFormattedText(IXLFontBase defaultFont)
         {
             Length = 0;
             _defaultFont = defaultFont;
+            _isDefault = true;
         }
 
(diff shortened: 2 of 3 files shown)


igitur:
Thanks. There are other properties too that should be addressed. Look at #1184, please and see if we can do them in one go.

igitur:
https://github.com/ClosedXML/ClosedXML/pull/1185#issuecomment-485253376 : 

Properties that need private fields to indicate full initialisation:

- Hyperlink
- Comment
- RichText
- DataValidation

Pankraty:
Maybe it worths breaking an API a little?

E.g. `cell.Hyperlink` would return `null` instead of creating a new instance, while `cell.AddHyperlink()` would create one. Other options look like a pile of workarounds...

igitur:
I'm OK with that breaking change, but should we not then apply that approach to all these properties?

Pankraty:
Yes, this is what I meant. If we both agree then this PR is not needed.

igitur:
Let me first check. What would happen if one calls `AddHyperlink()` twice?

Pankraty:
Well, maybe AddHyperlink is not the best naming then... For data validations we have NewDataValidation property which is more explicit: it creates a new instance overwriting the existing one. So, for hyperlinks it would be NewHyperlink

Pankraty:
Will open a new PR

PR #1376, opened 2020-03-25
CEILING and FLOOR fixes

### LINKED ISSUE #1366: CEILING missing optional second parameter

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

e.g. 0.94.2

Ceiling function takes in two parameters in Excel 2013, where as the Ceiling function in ClosedXML has it hard coded to 1:

`ce.RegisterFunction("CEILING", 1, Ceiling);`

This causes an exception to be thrown when expressions such as this are encountered:

`=CEILING(2.845, 0.01)`

Exception should not be thrown. I noticed that the Floor function can accept up to 2 parameters, I am assuming this is for backward compatibility with older Excel that may have only supported one argument, where as Excel 2013+ needs two arguments:

`ce.RegisterFunction("FLOOR", 1, 2, Floor);`

No

## Reproducibility

Simply run code snippet

```c#
var wb = new XLWorkbook();
ws.Cell("A1").FormulaA1 = "=CEILING(0.5, 1)";
var answer = ws.Cell("A1").Value.ToString();
```

where as in Excel: 

image

mathlang:
- Change the signature of CEILING and FLOOR to accept only 2 arguments, matching Excel behavior
- Invalid arguments to the CEILING and FLOOR (positive number and negative significance) throw NumberException, matching Excel behavior
- Fixed the mode argument behavior in FLOOR.MATH to match Excel behavior
- Implemented CEILING.MATH

Fixes #1366

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Functions/MathTrig.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/MathTrig.cs
@@ -26,7 +26,8 @@ namespace ClosedXML.Excel.CalcEngine
             ce.RegisterFunction("ATAN2", 2, Atan2);
             ce.RegisterFunction("ATANH", 1, Atanh);
             ce.RegisterFunction("BASE", 2, 3, Base);
-            ce.RegisterFunction("CEILING", 1, Ceiling);
+            ce.RegisterFunction("CEILING", 2, Ceiling);
+            ce.RegisterFunction("CEILING.MATH", 1, 3, CeilingMath);
             ce.RegisterFunction("COMBIN", 2, Combin);
             ce.RegisterFunction("COMBINA", 2, CombinA);
             ce.RegisterFunction("COS", 1, Cos);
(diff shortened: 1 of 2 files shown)


ClosedXML/Excel/CalcEngine/Functions/MathTrig.cs
```diff
@@ -129,7 +130,34 @@ private static object Atan2(List<Expression> p)
 
         private static object Ceiling(List<Expression> p)
         {
-            return Math.Ceiling(p[0]);
+            double number = p[0];
+            double significance = 1;
+            if (p.Count > 1)
+                significance = p[1];
```
Pankraty: Since the number of arguments is exactly two this can be simplified to `double significance = p[1];`

mathlang: Good catch. I implemented your recommendation.

mathlang:
> What should happen when the significance = 0?

I added code and tests to explicitly replicate the Excel behavior.

PR #1379, opened 2020-03-28
Explicitly set document type on saving (#1375)

### LINKED ISSUE #1375: Changing SpreadsheetDocumentType
- [ ] Bug
- [x] Feature

**Version of ClosedXML**

develop branch

If you open an xltx file you can only save it as xltx.
The SpreadsheetDocumentType doesn't get changed and is a private field in the workbook class.

You should be able to open an xltx file and save it as xlsx.

No

Pankraty:
Fixes #1375

### DIFF
--- a/ClosedXML/Excel/XLWorkbook_Save.cs
+++ b/ClosedXML/Excel/XLWorkbook_Save.cs
@@ -122,6 +122,11 @@ namespace ClosedXML.Excel
 
             using (package)
             {
+                if (package.DocumentType != spreadsheetDocumentType)
+                {
+                    package.ChangeDocumentType(spreadsheetDocumentType);
+                }
+
                 CreateParts(package, options);
                 if (options.ValidatePackage) Validate(package);
             }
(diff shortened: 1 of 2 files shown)


igitur:
Should we support saving as other types?

Pankraty:
I don't see why not. Having xltx used as a template, it is natural to have an ability to save it as a workbook. And vice versa - to create a template from a given workbook.

igitur (approved):

thekokii:
are we missing something or this didn't make it to release?

igitur:
This is released. If you encounter a bug in this regards, please open a new issue.

thekokii:
Yes I can see it now, but isn't this achievable saving your excel as memorystream? We would like to open an .xlsm and later save it without the macro as .xlsx but returning it as a memorystream. Is this somehow achievable or a temporary file is needed?

chrisjshields:
Is it possible to add this feature when saving a file to a stream too?

I can potentially add it and create a pull request if need be.

chrisjshields:
Looking through the source, it still does not seem possible to actually change the SpreadsheetDocumentType, it is simply set when the spreadsheet is originally loaded. So I guess you would need to save the workbook as an xlsx file, reload it and save it again?

PR #1388, opened 2020-04-02
Remove the dependency on FastMember. Improve performance of InsertData (#797)

### LINKED ISSUE #797: Remove dependency on FastMember
I originally added this library, because I read that Reflection is slow. And we used a lot of reflection in the `.InsertData` and `.InsertTable` methods. I actually never tested the impact on performance for this change.

I think it's worth relooking at this issue. We could either remove its use totally or compile it inside ClosedXML. 

FastMember is licensed under Apache v2

Pankraty:
(_edit Francois_) Addresses #797 , but does not completely close it. We should still provide a stripped down version of dumping data with minimal type checking, probably in a new `InsertXXX` method.

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -1,5 +1,4 @@
-﻿using FastMember;
-using System;
+﻿using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Data;
@@ -867,10 +866,8 @@ namespace ClosedXML.Excel
             else
             {
                 const BindingFlags bindingFlags = BindingFlags.Public | BindingFlags.Instance | BindingFlags.Static;
-                var memberCache = new Dictionary<Type, IEnumerable<MemberInfo>>();
-                var accessorCache = new Dictionary<Type, TypeAccessor>();
-                IEnumerable<MemberInfo> members = null;
-                TypeAccessor accessor = null;
+                var memberCache = new Dictionary<Type, MemberInfo[]>();
+                MemberInfo[] members = null;
                 bool isPlainObject = itemType == typeof(object);
 
                 if (!isPlainObject)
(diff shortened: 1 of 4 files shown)


ClosedXML/Excel/XLTheme.cs
```diff
@@ -18,16 +18,49 @@ internal class XLTheme : IXLTheme
         public XLColor Hyperlink { get; set; }
         public XLColor FollowedHyperlink { get; set; }
 
-        private TypeAccessor accessor = TypeAccessor.Create(typeof(XLTheme));
-
         public XLColor ResolveThemeColor(XLThemeColor themeColor)
         {
-            var tc = themeColor.ToString();
```
igitur, line 25: I'd like to future-proof this method, in case new theme colours are added. I'd prefer to keep this approach, but just rely in reflection.

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -867,19 +866,17 @@ void resetRecordPosition()
...
 
                 if (!isPlainObject)
                 {
                     members = itemType.GetFields(bindingFlags).Cast<MemberInfo>()
                          .Concat(itemType.GetProperties(bindingFlags))
                          .Where(mi => !XLColumnAttribute.IgnoreMember(mi))
-                         .OrderBy(mi => XLColumnAttribute.GetOrder(mi));
-                    accessor = TypeAccessor.Create(itemType);
+                         .OrderBy(mi => XLColumnAttribute.GetOrder(mi))
+                         .ToArray();
```
igitur, line 879: Do you know of any Visual Studio plugins that can warn me about silly enumerator mistakes like these?

Pankraty: Resharper indicates possible multiple enumerations of `IEnumerable` (https://www.jetbrains.com/help/resharper/2020.1/PossibleMultipleEnumeration.html). This may help.

ClosedXML/Excel/XLTheme.cs
Pankraty: That's why I added a test `CanResolveAllThemeColors` - it will start failing if we forget to add a branch in the switch. The idea is, `ResolveThemeColor` is called way more often that new enum values are added, so it is beneficial to replace reflection with switch-based code.

igitur: But that will require a new build of ClosedXML. But OK, let's cross that bridge when we get there, if ever.

PR #1389, opened 2020-04-03
Refactoring of InsertData logic (#797)

### LINKED ISSUE #1479: InsertTable with an empty collection of objects ignores column names

- [x] Bug
- [ ] Feature
- [ ] Question

- [x] Yes
- [ ] No

**Version of ClosedXML**

e.g. 0.95.3

If I pass a collection of objects to `InsertTable` method the column names are generated automatically: either from the properties defined in `[XLColumn]` attribute or from the property names. But if I pass an empty collection there the resulted tables will have default column names: `Column1`, `Column2`, etc. Their number corresponds to the number of properties the class has, given that `XLColumn(Ignore=true)` does not have an effect too.

1. The column names must be the same as if the collection would not be empty: either from `XLColumn` attribute or from property names.
2. Properties with the attribute `XLColumn(Ignore=true)` must be ignored.

Same behavior in 0.94.2

## Reproducibility

```c#
    internal static class Program
    {
        private static void Main(string[] args)
        {
            var entities = new Test[0];

            var wb = new XLWorkbook();
            ws.FirstCell().InsertTable(entities, "Entities");

            wb.SaveAs("Table.xlsx");

            Console.ReadKey();
        }
    }

    public class Test
    {
        public string Name { get; set; }

        [XLColumn(Header = "This is a description")]
        public string Description { get; set; }

        [XLColumn(Ignore = true)]
        public string Hidden { get; set; }
    }
```

image

[Table.xlsx](https://github.com/ClosedXML/ClosedXML/files/4951828/Table.xlsx)

Cc @ehartwig

Pankraty:
It is a continuation of #1388. It does not improve performance very much but, in my opinion, makes `InsertDataInternal` method much easier to understand and to extend, if we need it in the future.

Also, it fixes #1479 and adds support for collections of nullable structs as a table source.

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -1,19 +1,17 @@
-﻿using System;
+﻿using ClosedXML.Excel.InsertData;
+using ClosedXML.Extensions;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Data;
 using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
-using System.Reflection;
 using System.Text;
 using System.Text.RegularExpressions;
 
 namespace ClosedXML.Excel
 {
-    using Attributes;
-    using ClosedXML.Extensions;
-
     [DebuggerDisplay("{Address}")]
     internal class XLCell : XLStylizedBase, IXLCell, IXLStylized
     {
@@ -729,10 +727,17 @@ namespace ClosedXML.Excel
 
         public IXLTable InsertTable<T>(IEnumerable<T> data, String tableName, Boolean createTable, Boolean addHeadings, Boolean transpose)
         {
+            var reader = InsertDataReaderFactory.Instance.CreateReader(data);
+            return InsertTableInternal(reader, tableName, createTable, addHeadings, transpose);
+        }
+
+        private IXLTable InsertTableInternal(IInsertDataReader reader, String tableName, Boolean createTable, Boolean addHeadings,
+            Boolean transpose)
+        {
             if (createTable && this.Worksheet.Tables.Any(t => t.Contains(this)))
                 throw new InvalidOperationException(String.Format("This cell '{0}' is already part of a table.", this.Address.ToString()));
 
-            var range = InsertDataInternal(data, addHeadings, transpose);
+            var range = InsertDataInternal(reader, addHeadings, transpose);
 
             if (createTable)
                 // Create a table and save it in the file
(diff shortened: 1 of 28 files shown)


ClosedXML/Excel/InsertData/InsertDataReaderFactory.cs
```diff
@@ -13,12 +13,15 @@ internal class InsertDataReaderFactory
 
         public static InsertDataReaderFactory Instance => _instance.Value;
 
-        public IInsertDataReader CreateReader<T>(IEnumerable<T> data)
+        public IInsertDataReader CreateReader(IEnumerable data)
```
igitur, line 16: Was it not possible to keep both the generic and non-generic as 2 overloads?

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -749,57 +757,35 @@ public IXLTable InsertTable(DataTable data, String tableName, Boolean createTabl
...
-
-            foreach (DataColumn col in data.Columns)
-            {
-                Worksheet.SetValue(col.ColumnName, _rowNumber, co);
-                co++;
+            var reader = InsertDataReaderFactory.Instance.CreateReader(data);
+            return InsertTableInternal(reader, tableName, createTable, addHeadings: true, transpose: false);
             }
-
-            ClearMerged();
```
igitur, line 768: I can't see whether this commit retains the `ClearMerged()` behaviour.

igitur: Oh... the `.Clear()` in https://github.com/ClosedXML/ClosedXML/pull/1389/commits/1f4f3e017bba03e4fe6f658fe28ee9fc863bdf22#diff-3b8a562c15b0aca901df66d96df87e10R839 takes care of that?

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -840,206 +826,43 @@ void resetRecordPosition()
...
-                                if (mi.MemberType == MemberTypes.Property && (mi as PropertyInfo).GetGetMethod().IsStatic)
-                                    Worksheet.SetValue((mi as PropertyInfo).GetValue(null, null), currentRowNumber, currentColumnNumber);
-                                else if (mi.MemberType == MemberTypes.Field && (mi as FieldInfo).IsStatic)
-                                    Worksheet.SetValue((mi as FieldInfo).GetValue(null), currentRowNumber, currentColumnNumber);
-                                else if (mi.MemberType == MemberTypes.Property)
-                                    Worksheet.SetValue((mi as PropertyInfo).GetValue(m, null), currentRowNumber, currentColumnNumber);
-                                else if (mi.MemberType == MemberTypes.Field)
-                                    Worksheet.SetValue((mi as FieldInfo).GetValue(m), currentRowNumber, currentColumnNumber);
-
+                    Worksheet.SetValue(value, currentRowNumber, currentColumnNumber);
```
igitur, line 859: Some whitespace issues here.

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -244,8 +244,6 @@ internal IXLCell SetValue<T>(T value, bool setTableHeader)
 
             SetInternalCellValueString(parsedValue, validate: true, parseToCachedValue: false);
 
-            CachedValue = null;
```
igitur, line 246: I'm skeptical about this line's removal.

igitur: https://github.com/ClosedXML/ClosedXML/pull/1389/commits/9a468dcb1068ba3f387da1b964ef84e668b0d6f1#diff-3b8a562c15b0aca901df66d96df87e10R223 maintains this.

ClosedXML/Excel/InsertData/InsertDataReaderFactory.cs
Pankraty: This would not add any value but would bring more confusion during type casting. Like, I've just found that `IEnumerable<string>` produces `ArrayReader` as the argument is treated as `IEnumerable<IEnumerable>`. Gonna fix this.

ClosedXML/Excel/Cells/XLCell.cs
Pankraty: Yes

igitur:
Would you be able to reduce the number of commits? I'm thinking something that reduces to e.g.
- A commit for enhancements to current extension methods (and the added tests)
- Single commit for the final data readers and their tests
- Implementation of the data readers in `XLCell.InsertData`.

I have some other comments too, but let's do that first.

igitur:
@Pankraty Or actually, I tried to squash and reorder the commits. Have a look at my https://github.com/igitur/ClosedXML/tree/pr1389-rework branch and whether you're OK with what I've done there.

Pankraty:
@igitur, thanks for the effort rebasing my branch, hopefully there were not too many merge conflicts. 
Your changes look good for me.

igitur:
No, only a few small merge conflicts.

igitur (changes requested):
More comments/questions

ClosedXML/Excel/InsertData/UntypedObjectReader.cs
```diff
@@ -0,0 +1,88 @@
...
+    {
+        private readonly IEnumerable<object> _data;
+        private readonly IEnumerable<IInsertDataReader> _readers;
+
+        public UntypedObjectReader(IEnumerable data)
+        {
+            _data = (data ?? new object[0]).Cast<object>();
+            _readers = CreateReaders().ToList();
+
+            IEnumerable<IInsertDataReader> CreateReaders()
```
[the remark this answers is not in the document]
Pankraty: It's like declaring "private-private" member that cannot be called from anywhere else. If it is supposed to be called from the constructor solely, for initializing a readonly field, there is no point to make it accessible from anywhere else.

ClosedXML/Extensions/EnumerableExtensions.cs
```diff
@@ -14,7 +15,19 @@ public static void ForEach<T>(this IEnumerable<T> source, Action<T> action)
 
         public static Type GetItemType<T>(this IEnumerable<T> source)
         {
-            return typeof(T);
+            return GetGenericArgument(source?.GetType()) ?? typeof(T);
+
+            Type GetGenericArgument(Type collectionType)
```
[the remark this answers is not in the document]
Pankraty: If we make this an independent method we'll have to add additional safety checks that make sure `collectionType` is a collection type for real. In a nested method we can rely on `source` being `IEnumerable` and can omit these checks.
Besides, the method implementation is focused on `IEnumerable<T>` only, so using it for other purposes will be wrong (or the method will be less efficient).

ClosedXML/Excel/InsertData/ObjectReader.cs
```diff
@@ -0,0 +1,103 @@
...
+            var itemType = data.GetItemType();
+            _members = itemType.GetFields(bindingFlags).Cast<MemberInfo>()
+                .Concat(itemType.GetProperties(bindingFlags))
+                .Where(mi => !XLColumnAttribute.IgnoreMember(mi))
+                .OrderBy(mi => XLColumnAttribute.GetOrder(mi))
+                .ToArray();
+
+            _staticMembers = _members.Select(IsStatic).ToArray();
+
+            bool IsStatic(MemberInfo memberInfo)
```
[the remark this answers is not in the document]
Pankraty: Done

igitur (approved):

PR #1392, opened 2020-04-12
Remove unnecessary delete

igitur:
Some refactoring and removal of code that had no effect.


ClosedXML/Excel/Rows/XLRowCollection.cs
```diff
@@ -72,11 +72,6 @@ public void Add(KeyValuePair<int, XLRow> item)
 
         public void Clear()
```
Pankraty, line 73: This method is also called from `XLWorksheet.Clear()` but it is never used. The only way to call it is to cast `IXLWorksheet` to `XLWorksheet` and call `Clear`; without the casting, the `XLRangeBase.Clear()` is called. So I suggest removing `XLWorksheet.Clear()` too to avoid confusion.

```
        [Test]
        public void WorksheetClear()
        {
            {
                {
                    var ws = wb.AddWorksheet();
                    ws.FirstCell().Value = "123";
                    wb.SaveAs(ms);
                }

                using (var wb = new XLWorkbook(ms))
                {
                    var ws = wb.Worksheets.First() as XLWorksheet;
                    ws.Clear();
                    wb.Save();
                }

                using (var wb = new XLWorkbook(ms))
                {
                    Assert.AreEqual("", ws.FirstCell().Value);
                }
            }
        }
```

igitur: Not sure if I understand you correctly, but `XLWorksheetInternals.Dispose()` calls `RowsCollection.Clear()`. This is how I discovered this unused code.

Pankraty: I mean, there is one more unused method `XLWorksheet.Clear()` that can be deleted too. Otherwise, it becomes broken that the test illustrates.

igitur: Oh, I see. Maybe `Clear` isn't the best name here anyway. All those methods should maybe be `Reset()`. Do you agree? Then, even if we don't use it, we can leave `XLWorksheet.Reset()` for possible later use.

Pankraty: If we keep the current `XLWorksheet.Clear()` (just rename it to `Reset()`) then we have to revert the changes you've done here in `XLRowCollection.Clear()`. In other words, to make a non-used method correct we'll have to make worksheet disposal slightly slower. I think this does not worth it, better get rid of unused method.

igitur: true

Issue #680, opened 2018-02-05
Inserting rows above/below rows with range formatting and data validation causes Closed XML to hang


jaylittle:
**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.91.0

**What is the current behavior?**

Execution does not complete.  It will hang on the InsertRowsAbove(1) call near the end.

**What is the expected behavior or new feature?**

Execution should complete.  Closed XML should not hang.  This code should produce a simple Excel file.

**Did this work in previous versions of our tool?  Which versions?**

No.  We have tried several versions and have been able to consistently reproduce this issue.  This can be replicated in 0.87.1 very easily - you just need to change XLDataType enum references to XLCellValues.

## Reproducibility

**Code to reproduce problem:**
```c#
private const string OUTPUT_FILE = "c:\\temp\\test_output.xlsx";

static void Main(string[] args)
{
  //If output file already exists, nuke it
  if (System.IO.File.Exists(OUTPUT_FILE))
  {
    System.IO.File.Delete(OUTPUT_FILE);
  }

  //Prep book and sheets
  var workbook = new ClosedXML.Excel.XLWorkbook();
  var dataWorksheet = workbook.AddWorksheet("Data");
  var listWorksheet = workbook.AddWorksheet("Lists");

  //Setup List Sheet
  listWorksheet.Cell(1, 1).Value = "List";
  for (var rowptr = 2; rowptr <= 101; rowptr++)
  {
    listWorksheet.Cell(rowptr, 1).Value = (rowptr - 1).ToString();
    listWorksheet.Cell(rowptr, 1).SetValue<string>((rowptr - 1).ToString());
    listWorksheet.Cell(rowptr, 1).DataType = ClosedXML.Excel.XLDataType.Text;
  }

  //Setup Data Validation (comment last two lines to prevent breakage)
  var listRange = listWorksheet.Range(2, 1, 101, 1);
  var ddValidation = dataWorksheet.Range(2, 1, 9999, 1).SetDataValidation();
  ddValidation.List(listRange, true);
  ddValidation.AllowedValues = ClosedXML.Excel.XLAllowedValues.List;

  //Setup Data Sheet (comment last line to prevent breakage)
  dataWorksheet.Cell(1, 1).Value = "Data";
  dataWorksheet.Range(2, 1, 9999, 1).DataType = ClosedXML.Excel.XLDataType.Text;

[...]

PR #1396, opened 2020-04-16
Avoid instantiation of cell properties on reading (#1184)

### LINKED ISSUE #1184: Better treatment of half-baked properties (eg RichText, Hyperlink etc)

Pankraty:
Fixes #1184

Historically, we had a few properties that were initialized on first access to them. It might be quite convenient to not have to initialize them manually but that also caused cells to change their state on simple inspection their properties via Watch window, or on outputting them in PowerShell.

There also were many nasty issues with data validation rules (#680), and with rich texts (#1271), so we [agreed](https://github.com/ClosedXML/ClosedXML/pull/1370#issuecomment-598219364) on introducing breaking changes that will let us avoid such issues all at once.

Initially, I was going to change the behavior of existing properties so that their getters would not change the object state, but then I realized that this will bring lots of `NullReferenceException`s in our users' code that can easily be missed before going to production. So I decided it will be safer to change the API in a way it won't even compile. Yes, it brings _some_ inconvenience in the upgrade process, but this is going to be a simple mechanical process - replace property accessors with corresponding methods - which is far better than getting unexpected exceptions in runtime.

**Breaking changes:**
* `IXLCell.Hyperlink` removed; use `GetHyperlink()` to get an existing hyperlink or to create a new one if it does not exist; instead of property setter, use `SetHyperlink()`; to replace the existing hyperlink with a fresh one, use `CreateHyperlink()`;
* `IXLCell.DataValidation` removed; use `GetDataValidation()`;
* `IXLCell.NewDataValidation` removed; use `CreateDataValidation()`;
* `IXLCell.Comment` removed; use `GetComment()` to access the existing comment or to create a new one if it does not exist; use `CreateComment()` to replace the existing comment with the fresh one;
* `IXLCell.RichText` removed; use `GetRichText()` to access the existing rich text or to create a new one if it does not exist; use `CreateRichText()` to replace the existing rich text with the fresh one.

**Obsolete members**
* `IXLCell.SetDataValidation()`
* `IXLRangeBase.SetDataValidation()`
* `IXLRanges.SetDataValidation()`

Their functionality is covered by `GetDataValidation()` and `CreateDataValidation()` so they will be removed in future.

### DIFF
--- a/ClosedXML/Excel/Cells/IXLCell.cs
+++ b/ClosedXML/Excel/Cells/IXLCell.cs
@@ -71,8 +71,6 @@ namespace ClosedXML.Excel
 
         Boolean HasSparkline { get; }
 
-        XLHyperlink Hyperlink { get; set; }
-
         /// <summary>
         /// Flag indicating that previously calculated cell value may be not valid anymore and has to be re-evaluated.
         /// </summary>
@@ -181,6 +179,11 @@ namespace ClosedXML.Excel
         IXLDataValidation CreateDataValidation();
 
         /// <summary>
+        /// Creates a new hyperlink replacing the existing one.
+        /// </summary>
+        XLHyperlink CreateHyperlink();
+
+        /// <summary>
         /// Replaces a value of the cell with a newly created rich text object.
         /// </summary>
         IXLRichText CreateRichText();
@@ -227,6 +230,9 @@ namespace ClosedXML.Excel
         /// </summary>
         String GetFormattedString();
 
+        /// <summary>
+        /// Returns a hyperlink for the cell, if any, or creates a new instance is there is no hyperlink.
+        /// </summary>
         XLHyperlink GetHyperlink();
 
         /// <summary>
@@ -392,6 +398,8 @@ namespace ClosedXML.Excel
         IXLCell SetFormulaA1(String formula);
 
         IXLCell SetFormulaR1C1(String formula);
+
+        void SetHyperlink(XLHyperlink hyperlink);
 
         /// <summary>
         /// Sets the cell's value.
--- a/ClosedXML/Excel/Cells/IXLCell.cs
+++ b/ClosedXML/Excel/Cells/IXLCell.cs
@@ -84,8 +84,6 @@ namespace ClosedXML.Excel
 
         IXLDataValidation NewDataValidation { get; }
 
-        IXLRichText RichText { get; }
-
         /// <summary>
         /// Gets or sets a value indicating whether this cell's text should be shared or not.
         /// </summary>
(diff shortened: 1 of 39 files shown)


igitur (approved):

cedvdb:
Rich text does not seem to display for me, maybe this is the root cause ?

```
                PrestationCell.CreateRichText();
                PrestationCell.GetRichText()
                        .AddText(label)
                        .SetFontColor(fontColor)
                        .SetStrikethrough(shouldStrikeThrough);
```

PR #1398, opened 2020-04-18
Update langversion

shaggygi:
Updated language version to allow for new syntax and improve code in future PRs.

### DIFF
--- a/ClosedXML/ClosedXML.csproj
+++ b/ClosedXML/ClosedXML.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net40;net46</TargetFrameworks>
-    <LangVersion>7.2</LangVersion>
+    <LangVersion>8.0</LangVersion>
     <AssemblyName>ClosedXML</AssemblyName>
     <PackageId>ClosedXML</PackageId>
     <Version>0.95.2</Version>
--- a/ClosedXML_Examples/ClosedXML_Examples.csproj
+++ b/ClosedXML_Examples/ClosedXML_Examples.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>netcoreapp2.0;net40;net46</TargetFrameworks>
-    <LangVersion>7.2</LangVersion>
+    <LangVersion>8.0</LangVersion>
     <OutputType>Exe</OutputType>
     <Version>0.95.2</Version>
     <NoWarn>$(NoWarn);NU1605</NoWarn>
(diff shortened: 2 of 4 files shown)


shaggygi:
> I think, it's better to specify the version explicitly (e.g. 8.0), this way all developers will be on the same page.

Good point.  Now updated to v8.0.

shaggygi:
Can this be merged now?

Pankraty:
Resolve merge conflicts please.

shaggygi:
Updated

igitur:
I'm hesitant about this. I just push an additional commit to illustrate that one of the C# 8 feature, default interface methods, are not supported in any of the target frameworks that we currently support (`net40`, `net46` and `netstandard2.0`).

`Excel\IXLWorksheet.cs(493,21): error CS8701: Target runtime doesn't support default interface implementation. [C:\projects\closedxml\ClosedXML\ClosedXML.csproj]`

Some of C#8's features are syntactical sugar, e.g. the inlined `using` declarations. But others genuinely require `netstandard2.1` or later. I'm not sure that we should add C#8 support if it means only partially supporting its features. Besides syntactical sugar, what are the benefits of supporting C#8 right now?

igitur:
> Resolve merge conflicts please.

I did this on the author's behalf.

shaggygi:
> Besides syntactical sugar, what are the benefits of supporting C#8 right now?

It was basically for the syntax sugar.  I'd hold off then if it means breaking things on other platforms.

Pankraty:
I, personally, like switch expressions and `using` declarations. So I guess, we could move to C# 8 and simply do not use the features that are not supported, like we do not use `ValueTuples` as it would require adding a new dependency in `net40`.

igitur:
On the other hand, I see in https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/struct#readonly-instance-members that marking readonly instance members can lead to performance improvements. 

The `using` declarations will be useful, yes.

I quickly played around with the new switch patterns and couldn't get the intellisense in my VS2019 working nicely. I'll investigate later.

I'm not saying yet that we should abandon this PR. Just that we should evaluate it carefully.

igitur:
See https://stackoverflow.com/questions/56651472/does-c-sharp-8-support-the-net-framework/57020770#57020770 for a list of C#8 features which work and don't work in .NET Framework.

Let's allow only these:
- Static local functions
- Using declarations
- Null-coalescing assignment
- Readonly members
- Disposable ref structs
- Positional patterns
- Tuple patterns
- Switch expressions
- Nullable reference types are also supported, but the new nullable attributes required to design the more complex nullable

PR #1399, opened 2020-04-18
.CellsUsed() should return only cells with content. 

### LINKED ISSUE #1367: Optionally pass through the `CultureInfo` instance in the `XLWorkbook` instance
Follow-up to #1359

This relates to a comment I made somewhere in another PR. I think we should optionally pass through the `CultureInfo` instance in the `XLWorkbook` instance and use that everywhere where applicable. Default to `CurrentCulture`. We'll have to do a separate PR for that.

_Originally posted by @igitur in https://github.com/ClosedXML/ClosedXML/pull/1359/files_

igitur:
Fixes #1367 . 
Regression was in 226a1c8689b0eed3fa5d7b5c5af9651fef8b5b29

`.CellsUsed()` should return only cells with content. Cells without content, but with formatting should not be returned here.

### DIFF
--- a/ClosedXML/Excel/Columns/XLColumn.cs
+++ b/ClosedXML/Excel/Columns/XLColumn.cs
@@ -97,7 +97,7 @@ namespace ClosedXML.Excel
         public override IXLCells Cells(Boolean usedCellsOnly)
         {
             if (usedCellsOnly)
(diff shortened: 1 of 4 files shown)


igitur:
v0.95.1 has been deployed to NuGet. Please test and let me know if this solves your problem.

PR #1400, opened 2020-04-21
Ensure auto-generated table name is unique across all worksheets

### LINKED ISSUE #1230: TotalsRowFunction issue
- [x] Bug

**Version of ClosedXML**

Latest from Nuget: 0.94.2

It seems like TotalsRowFunction fails to create a valid cell reference when these two conditions are met:
- The table column name has spaces (e.g., "AB CD")
- The name of the sheet where the table exists also has spaces (e.g., "Two words")

When you open the generated file in Excel you are prompted to accept the repair actions. The XML report includes a message like "Repaired Records: Table from /xl/tables/table2.xml part (Table)"

The resulting formula in the Totals cell is "=SUBTOTAL(109,Two #REF!)"

The formula in the Totals cell should be: "=SUBTOTAL(109,[AB CD])"

As far as I know it was working fine with version 0.92.

## Reproducibility

This issue seems to be related to case #773 (Bug with TotalsRowFunction). I adapted the code provided by healkar01 to reproduce the problem. Look at the "Two words2" sheet in the generated XLS file to see the issue in the Grand Total cell (B11)

```c#
void Main()
{

	InsertDataTable(wb, "OneWord", GenerateData(useMultiWordColumn: false));
	InsertDataTable(wb, "OneWord2", GenerateData(useMultiWordColumn: true));
	InsertDataTable(wb, "Two words", GenerateData(useMultiWordColumn: false));
        // the line below reproduces the issue
	InsertDataTable(wb, "Two words2", GenerateData(useMultiWordColumn: true)); 

	wb.SaveAs($@"C:\{DateTime.Now.ToString("yyyyMMdd-HHmmss")}-ClosedXml-Issue.xlsx");
}

static DataTable GenerateData(bool useMultiWordColumn)
{
	var columnName = useMultiWordColumn ? "AB CD" : "ABCD";
	DataTable dt = new DataTable();
	dt.Clear();
	dt.Columns.Add("Service", typeof(String));
	dt.Columns.Add(columnName, typeof(Int32));

	for (int i = 1; i < 10; i++)
	{
		DataRow dtRow = dt.NewRow();
		dtRow["Service"] = "Service " + i;
		dtRow[columnName] = 500;
		dt.Rows.Add(dtRow);
	}
	return dt;
}

static void InsertDataTable(XLWorkbook wb, string sheetName, DataTable dt)
{
	var ws = wb.Worksheets.Add(sheetName);

	// Insert our datatable into the data sheet at cell 1,1
	var table = ws.Cell(1, 1).InsertTable(dt, sheetName, true);

	ws.Columns().AdjustToContents();
	table.ShowTotalsRow = true;
	foreach (var field in table.Fields)
	{
		if (field.Index > 0 && field.Name != "")
		{
			field.TotalsRowFunction = XLTotalsRowFunction.Sum;
		}
	}

	table.Field(0).TotalsRowLabel = "Grand Total";
}
```

Thanks! This is a great tool, thank you for providing it. Let me know if you need more information.

igitur:
Ensure auto-generated table name is unique across all worksheets.

Fixes #1230


ClosedXML/Excel/XLWorksheet.cs
```diff
@@ -1599,13 +1599,15 @@ private void CheckRangeNotInTable(XLRange range)
 
         private string GetNewTableName(string baseName)
         {
+            var existingSheetNames = this.Workbook.Worksheets.SelectMany(ws => ws.Tables).Select(t => t.Name).ToList();
```
Pankraty, line 1602: It's `existingTableNames`, isn't it?

ClosedXML_Tests/Excel/Tables/TablesTests.cs
```diff
@@ -1114,6 +1114,21 @@ public void TableNotFound()
...
+                var ws1 = wb.AddWorksheet();
+                var t1 = ws1.FirstCell().InsertTable(Enumerable.Range(1, 10).Select(i => new { Number = i }));
+                Assert.AreEqual("Table1", t1.Name);
+
+                var ws2 = wb.AddWorksheet();
+                var t2 = ws2.FirstCell().InsertTable(Enumerable.Range(1, 10).Select(i => new { Number = i }));
+                Assert.AreEqual("Table2", t2.Name);
+            }
+        }
+
```
Pankraty, line 1131: It seems, the reference file `AddingDataSet.xlsx` has to be updated: it contained `Table1`, `Table11`, and `Table12` (not 1, 2, 3). I wonder how did we miss that...

ClosedXML/Excel/XLWorksheet.cs
Pankraty: Besides, I'd suggest to make it `new HashSet<string>(
                Workbook.Worksheets.SelectMany(ws => ws.Tables).Select(t => t.Name),
                StringComparer.OrdinalIgnoreCase);`

This will make searching slightly more efficient in cases when the user dumps out _many_ tables.

ClosedXML_Tests/Excel/Tables/TablesTests.cs
igitur: I've noticed that before. Just never had the motivation to chase the problem.

ClosedXML/Excel/XLWorksheet.cs
igitur: Pity `.ToHashSet()` appeared only in .NET 4.7.2.

igitur:
@Pankraty Stupid errors fixed.

ClosedXML/Excel/XLWorksheet.cs
```diff
@@ -1599,13 +1599,15 @@ private void CheckRangeNotInTable(XLRange range)
...
+            var existingTableNames = new HashSet<String>(this.Workbook.Worksheets.SelectMany(ws => ws.Tables).Select(t => t.Name));
+
             var i = 1;
             string tableName;
             do
             {
                 tableName = baseName + i.ToString();
                 i++;
-            } while (Tables.Any(t => t.Name == tableName));
+            } while (existingTableNames.Contains(tableName, StringComparer.OrdinalIgnoreCase));
```
Pankraty, line 1610: image

Wrong `Contains`, sorry. `HashSet` cannot optionate comparer in `Contains` by design as it has to compute hashcode on adding elements. That's why I proposed to pass a comparer in the constructor. In that case, we call `.Contains(tableName)` and the searching is case-insensitive.

igitur: Oooh, I see. Is it Resharper that gives you that warning?

Pankraty: In fact, that was just me suspicious about how is it possible to pass comparer in a hashset. But once you've said that I've noticed that there was a warning from Resharper indeed which I missed 😆 

image

PR #1402, opened 2020-04-22
IEnumerator improvements - avoid evaluation of empty sheets

### LINKED ISSUE #1390: NullReferenceException in CellRangeReference.GetEnumerator()

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.95

Calling Workbook.SaveTo(Stream) results in the exception mentioned above. The call to

Range.Worksheet.LastCellUsed()

in 

CellRangeReference.GetEnumerator() 

returns null.

Yes / v0.94

## Reproducibility
 
Unfortunately I cannot provide the workbook that causes this issue. If necessary I can try to reproduce the behaviour with a file I can share. This might be difficult as the workbook is quite complex, so I would need some directions as to what might be the cause of this. The full callstack is attached.

[callstack.txt](https://github.com/ClosedXML/ClosedXML/files/4437882/callstack.txt)

igitur:
Fixes #1390

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Expression.cs
+++ b/ClosedXML/Excel/CalcEngine/Expression.cs
@@ -509,10 +509,19 @@ namespace ClosedXML.Excel.CalcEngine
 
         public IEnumerator GetEnumerator()
         {
-            if (_value is string)
-                return new[] { (string)_value }.GetEnumerator();
-
-            return (_value as IEnumerable).GetEnumerator();
+            if (_value is string s)
+            {
+                yield return s;
+            }
+            else if (_value is IEnumerable ie)
+            {
+                foreach (var o in ie)
+                    yield return o;
+            }
+            else
+            {
+                yield return _value;
+            }
         }
 
         public override string LastParseItem
--- a/ClosedXML/Excel/CalcEngine/CellRangeReference.cs
+++ b/ClosedXML/Excel/CalcEngine/CellRangeReference.cs
@@ -23,12 +23,21 @@ namespace ClosedXML.Excel.CalcEngine
         // ** IEnumerable
         public IEnumerator GetEnumerator()
         {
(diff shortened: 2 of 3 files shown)


ClosedXML/Excel/CalcEngine/Expression.cs
```diff
@@ -509,10 +509,14 @@ public override object Evaluate()
...
-                return new[] { (string)_value }.GetEnumerator();
+            if (_value is string s)
+                yield return s;
 
-            return (_value as IEnumerable).GetEnumerator();
+            if (_value is IEnumerable ie)
+            {
+                foreach (var o in ie)
+                    yield return o;
+            }
```
igitur, line 519: Maybe if `_value` is not `string` or `IEnumerable`, then we should just `yield return _value`?

ClosedXML/Excel/CalcEngine/Expression.cs
```diff
@@ -509,10 +509,14 @@ public override object Evaluate()
 
         public IEnumerator GetEnumerator()
         {
-            if (_value is string)
-                return new[] { (string)_value }.GetEnumerator();
+            if (_value is string s)
+                yield return s;
 
-            return (_value as IEnumerable).GetEnumerator();
+            if (_value is IEnumerable ie)
```
Pankraty, line 515: Here we should use `else if`, otherwise we'll return a string and _then_ all chars it consists of. A known gotcha of `yield return` ;)

ClosedXML/Excel/CalcEngine/Expression.cs
Pankraty: Yes, it makes sense.

igitur:
@Pankraty I pushed new changed since you approved. Please just double-check that you still approve.

Pankraty:
Looks good

Issue #826, opened 2018-05-01
Implement a more consistent behavior for evaluation errors


Pankraty:
Inspired by #824, #585 

## Read and complete the full issue template

**Do you want to request a *feature* or report a *bug*?**
- [ ] Bug
- [X] Feature

**Version of ClosedXML**

0.93-beta2

**What is the current behavior?**

Dealing with errors occurring on formula evaluation is quite hard and counter-intuitive. Say, there are functions `HLOOKUP` and `VLOOKUP` that return `#N/A` when the value is not found. In ClosedXML they throw the exception making impossible to evaluate formulas that include such functions or refer to them (i.e. `=IF(ISNA(HLOOKUP(...), "Not found", ...)`).

**What is the expected behavior or new feature?**

1. Reading the value of the cell containing #N/A, #REF or other types of errors should not throw exceptions.
2. If the evaluation of the cell formula produces an error the `Value` of the cell should return the code of the error (enum containing codes for #NA, #REF, #DIV/0, etc.) and the `DataType` of the cell should be `Error` (a new value in the existing enum).
3. Putting a string representation of the error code (like '#N/A') should set the cell type to `Error` too (this is disputable, but that's how it works in Excel).
4. A text representation of the error code produced by formula (e.g. `="#N" & "/A"`) must not be treated as an error. 
5. Cells having errors should be possible to be saved to file and to be read from the file.

The same approach should be applied regarding all types of errors.

**Did this work in previous versions of our tool?  Which versions?**

Not that I knew

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

Without a code sample, it is unlikely that your issue will get attention. Don't be lazy. Do the effort and assist the developers to reproduce your problem. Code samples should be [minimal complete and verifiable](https://stackoverflow.com/help/mcve). Sample spreadsheets should be attached whenever applicable. Remove sensitive information.

**Code to reproduce problem:**
```c#
public void Main()
{
    //This code should not throw exceptions
    var ws = wb.Worksheets.Add("Sheet1");

    ws.Cell("A1").FormulaA1 = "=\"#N\" & \"/A\"";
    ws.Cell("A2").Value = "#N/A";
    ws.Cell("A3").FormulaA1 = "=HLOOKUP(0,C1:D3,1,FALSE)";

    ws.Range("B1:B3").FormulaR1C1 = "=ISNA(RC[-1])";

[...]

PR #1404, opened 2020-04-23
Replace calculation engine exceptions with XLError enum

igitur:


igitur:
@Pankraty Am I right in saying that this PR would address #826 ?

This PR currently is a WIP implementation of that. When a cell formula evaluates to an error, the appropriate enum value representing the error is returned. 

Internally I'm still using exceptions and catching them right at the end, but I think the exceptions should totally be removed and the error enum values should be tracked throughout, because exception catching is apparently slow.

I also had a thought to return a struct with a bit more info, basically the current enum representing the error type, but also a string with the error message to help the user a bit more with what went wrong. Like we're currently doing with sometimes passing a message into the exception.

Another idea... instead of creating a new struct, maybe we should just use the existing exception classes, but instead of throwing them, we pass back to the user as return values. The user could throw them if they want. Is that an abuse of exception classes?

Pankraty:
> instead of creating a new struct, maybe we should just use the existing exception classes, but instead of throwing them, we pass back to the user as return values. The user could throw them if they want. Is that an abuse of exception classes?

This would indeed be a very unexpected way of using exceptions so I would avoid it.

I don't fell strongly against throwing exceptions during evaluation. Yes, catching them is relatively slow, but the benefits are significant: in CalcEngine functions we don't have to bother about what result to return when one of the arguments is `#REF!`, or `#N/A`.

As one more option to consider, instead of structs for different types of errors we might use singlton class instances for each type of calculation error.

```
    public class XLReferenceError : XLCalculationError
    {
        private static readonly Lazy<XLReferenceError> _instance = new Lazy<XLReferenceError>(() => new XLReferenceError());
        public static XLReferenceError Instance = _instance.Value;
        private XLReferenceError()
        {
        }

        public override string ToString() => "#REF!";
    }
```

In `XLCell` there might be a property (e.g. `HasError`) checking that its value has any type derived from `XLCalculationError` and a method returning error details. This would be especially useful if we had a "quick" way to evaluate formula without collecting error details and a "slow" way with, but I am not sure this is a case for us.

PR #1414, opened 2020-04-29
Update data table name

### LINKED ISSUE #1411: DataTable Name not updating in Table Name field
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.95.2

The data table is attached to a worksheet, but still shows "Table1" in the Table Name field.

## Reproducibility
After spreadsheet is populated with data table:
- Click data table in worksheet
- Click Table Tools menu tab
- See Table Name field in Properties group.  Shows as Table1

```c#
public void Main()
{
    using var workbook = new XLWorkbook();
    DataTable table = new DataTable("RawDataTable");
    var worksheet = workbook.Worksheets.Add(table, "Raw Data");
}
```

shaggygi:
Fixes #1411

This updates the DataTable name when using similar code:

```C#
var xsTable = worksheet.Cell(1, 1).InsertTable(dataTable);
```
It also fixes the following since it calls into method above:

```C#
DataTable table = new DataTable("RawDataTable");
var worksheet = workbook.Worksheets.Add(table, "Raw Data");
```

I removed a couple of methods as they were confusing since the `tableName` is included with the `dataTable` argument.  This is a breaking change since they were part of the `IXLCell` interface, but being < v1... I'm assuming this is okay for now 😄 

Hope this helps.


ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -716,40 +716,30 @@ public IXLTable InsertTable<T>(IEnumerable<T> data, String tableName, Boolean cr
...
             else
                 // Create a table, but keep it in memory. Saved file will contain only "raw" data and column headers
-                return tableName == null ? range.AsTable() : range.AsTable(tableName);
+                return String.IsNullOrWhiteSpace(tableName) ? range.AsTable() : range.AsTable(tableName);
         }
 
         public IXLTable InsertTable(DataTable data)
         {
-            return InsertTable(data, null, true);
+            return InsertTable(data, true);
```
Pankraty, line 727: In my opinion, it is better to change this to 
```
return InsertTable(data, data.TableName, true);
```

and not remove methods from the interface. This way, we won't introduce breaking changes, and not limit the functionality. Some users may want to dump a table to Excel with an arbitrary name not having to modify the `DataTable` instance as it may be used elsewhere. E.g. `TableName` matches the name in the database, while the Excel tables are built for different dates, or different customers, or whatever, and their names should be specified manually.

shaggygi: @Pankraty good point.  I'll update your feedback.  Thx

shaggygi: Add InsertTable methods back.  I still think IXLCell/XLCell could be simplified with the following:

```C#
// Currently
public IXLTable InsertTable(DataTable data)
public IXLTable InsertTable(DataTable data, Boolean createTable)
public IXLTable InsertTable(DataTable data, String tableName)
public IXLTable InsertTable(DataTable data, String tableName, Boolean createTable)

// To...
public IXLTable InsertTable(DataTable data, Boolean createTable = true)
public IXLTable InsertTable(DataTable data, String tableName, Boolean createTable = true)
```

But at least the other changes fix my original issue.

shaggygi:
Can this be merged now?

PR #1417, opened 2020-04-30
Improve performance of operations involving intensive style edits (#1416)

### LINKED ISSUE #1416: Bad performance when setting borders
- [x] Bug/Performance
- [ ] Feature

**Version of ClosedXML**

0.95.2

Setting border styles is extremely slow when the amount is non-trivial.
In the attached example ClosedXml takes more than 2:30 min to set the borders

Setting the same borders in Excel takes only a fraction of a second. I would expect ClosedXml to offer similar performance or at least performance in the same ballpark.

## Reproducibility

```c#
using System;
using System.Diagnostics;
using ClosedXML.Excel;

namespace ExportExcelClosedXML
{
    class BorderPerf
    {
        static void Main()
        {
            Stopwatch watch = new Stopwatch();
            Stopwatch watch2 = new Stopwatch();
            watch.Start();

            var ws = workbook.Worksheets.Add("Table");

            watch2.Start();


            // The following block takes more than 2:30 min
            {
                var rngTable2 = ws.Range("B1:CU10000");

                var border = rngTable2.Style.Border;

                border.OutsideBorder = XLBorderStyleValues.Medium;
                border.OutsideBorderColor = XLColor.Black;

                border.InsideBorder = XLBorderStyleValues.Medium;
                border.InsideBorderColor = XLColor.Black;
            }

            watch2.Stop();

            workbook.SaveAs("TestClosedXML.xlsx");

            watch.Stop();
            Console.WriteLine("Total: " + watch.Elapsed);
            Console.WriteLine("Styles: " + watch2.Elapsed);
            Console.ReadLine();
        }
    }
}
```

Pankraty:
Fixes #1416

I suggest reviewing per commit.

1) Commit 1 - improvement from 2:33 to ~30
2) Commit 2 - from ~30 to ~12
3) Commit 3 - the effect is negligible
4) Commit 4 - from ~12 to ~9 (and reduced memory traffic)

Besides, I was experimenting with preserving the last obtained element in `XLRepositoryBase` - on intensive style edits high chances are that the next queried element is the same as was fetched before and we don't have to go to the dictionary. It saves another 2 seconds but I did not include this in PR as it will make our repositories non-thread safe. I know, we generally do not support multi-threading, but our repositories are shared globally, therefore they must be thread-safe; otherwise, concurrent requests in, say, ASP.Net application will skrew everything up.

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -1878,6 +1878,13 @@ namespace ClosedXML.Excel
             }
         }
 
+        void IXLStylized.ModifyStyle(Func<XLStyleKey, XLStyleKey> modification)
+        {
+            //XLCell cannot have children so the base method may be optimized
+            var styleKey = modification(StyleValue.Key);
+            StyleValue = XLStyleValue.FromKey(ref styleKey);
+        }
+
         protected override IEnumerable<XLStylizedBase> Children
         {
             get { yield break; }
--- a/ClosedXML/Excel/Style/XLStyleKey.cs
+++ b/ClosedXML/Excel/Style/XLStyleKey.cs
@@ -66,5 +66,23 @@ namespace ClosedXML.Excel
         public static bool operator ==(XLStyleKey left, XLStyleKey right) => left.Equals(right);
 
         public static bool operator !=(XLStyleKey left, XLStyleKey right) => !(left.Equals(right));
+
+        public void Deconstruct(
+            out XLAlignmentKey alignment,
+            out XLBorderKey border,
+            out XLFillKey fill,
+            out XLFontKey font,
+            out Boolean includeQuotePrefix,
+            out XLNumberFormatKey numberFormat,
+            out XLProtectionKey protection)
+        {
+            alignment = Alignment;
+            border = Border;
+            fill = Fill;
+            font = Font;
+            includeQuotePrefix = IncludeQuotePrefix;
+            numberFormat = NumberFormat;
+            protection = Protection;
+        }
     }
 }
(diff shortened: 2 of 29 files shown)


igitur:
I've been meaning to look at passing all structs by ref ever since it became a C# feature. IIRC, there are more places that we can do this?

ClosedXML/Excel/Style/XLStyleKey.cs
```diff
@@ -66,5 +66,23 @@ public override bool Equals(object obj)
         public static bool operator ==(XLStyleKey left, XLStyleKey right) => left.Equals(right);
 
         public static bool operator !=(XLStyleKey left, XLStyleKey right) => !(left.Equals(right));
+
+        public void Deconstruct(
```
igitur, line 70: And now I learn something new!

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -1877,6 +1877,12 @@ public override IEnumerable<IXLStyle> Styles
                 yield return Style;
             }
         }
+        void IXLStylized.ModifyStyle(Func<XLStyleKey, XLStyleKey> modification)
```
igitur, line 1880: Just so that I can give at least one suggestion for this PR, please add a blank line before this modification ;-)

Pankraty: I knew you will notice!! :)

federicoairoldials:
Hello,
coming late to the party... Has this correction been released in any of the subsequent version? I'm using the library to format large Excel files (2000 rows x 2000 columns), but I experience slowness also with latest stable version 0.104.

AntonSmolkov:
0.105.0 also still has the issue.

PR #1444, opened 2020-05-24
Avoid AutoFilter and Table overlap

### LINKED ISSUE #1213: Error with AutoFilter and Table

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.94.2

Error when opening file in Excel.  "We found a problem with some content in 'AddingDataTableAsWorksheet.xlsx'.  Repairs remove feature AutoFilter & Table from /xl/tables/table1.xml .  Repairing also remove the Table style.

Open the Excel file without error, and retains table style.

No

## Reproducibility

```c#
        private void Main()
        {
            var wb = new XLWorkbook();

            var dataTable = GetTable("Information");

            // Add a DataTable as a worksheet
            var ws = wb.Worksheets.Add(dataTable);

            // Add filters
            ws.RangeUsed().SetAutoFilter().Column(1).AddFilter(10)
              .AddFilter(100)
              .AddFilter(50);

            // Sort the filtered list
            ws.AutoFilter.Sort(1);

            wb.SaveAs("AddingDataTableAsWorksheet.xlsx");

        }

        private DataTable GetTable(String tableName)
        {
            DataTable table = new DataTable();
            table.TableName = tableName;
            table.Columns.Add("Dosage", typeof(int));
            table.Columns.Add("Drug", typeof(string));
            table.Columns.Add("Patient", typeof(string));
            table.Columns.Add("Date", typeof(DateTime));

            table.Rows.Add(25, "Indocin", "David", DateTime.Now);
            table.Rows.Add(50, "Enebrel", "Sam", DateTime.Now);
            table.Rows.Add(10, "Hydralazine", "Christoff", DateTime.Now);
            table.Rows.Add(21, "Combivent", "Janet", DateTime.Now);
            table.Rows.Add(100, "Dilantin", "Melanie", DateTime.Now);
            return table;
        }
```

igitur:
Fixes #1213

Some minor tweaks. 

Throw exceptions if autofilter and table might overlap.

Breaking changes: 
- Default table name is now equal to sheet name, to avoid Table1, Table2, etc.
- `ArgumentException` replaced with `InvalidOperationException` for overlapping autofilter and table ranges


ClosedXML_Tests/Excel/AutoFilters/AutoFilterTests.cs
```diff
@@ -122,11 +122,28 @@ public void CanCopyAutoFilterToNewSheetOnNewWorkbook()
...
+            var data = Enumerable.Range(1, 10).Select(i => new
+            {
+                Index = i,
+                String = $"String {i}"
+            });
+
+            var ws = wb.AddWorksheet();
+            ws.FirstCell().InsertTable(data);
+
+            Assert.Throws<ArgumentException>(() => ws.RangeUsed().SetAutoFilter());
```
Pankraty, line 144: Getting `ArgumentException` here is a bit confusing as we do not provide any argument. `InvalidOperationException` seems to fit better.

ClosedXML/Excel/XLWorksheets.cs
```diff
@@ -162,7 +162,7 @@ public IXLWorksheet Add(DataTable dataTable)
         public IXLWorksheet Add(DataTable dataTable, String sheetName)
         {
             var ws = Add(sheetName);
-            ws.Cell(1, 1).InsertTable(dataTable);
+            ws.Cell(1, 1).InsertTable(dataTable, sheetName);
```
Pankraty, line 165: Good idea

ClosedXML/Excel/AutoFilters/XLAutoFilter.cs
```diff
@@ -144,14 +146,18 @@ public XLAutoFilter Clear()
 
         public XLAutoFilter Set(IXLRangeBase range)
         {
+            var firstOverlappingTable = range.Worksheet.Tables.FirstOrDefault(t => t.RangeUsed().Intersects(range));
```
Pankraty, line 149: I guess we can safely assume there will not be many hundreds of table on a single worksheet, so this way of searching is efficient enough.

ClosedXML_Tests/Excel/AutoFilters/AutoFilterTests.cs
igitur: True. I'll change that. As I was coding this I thought about how we really need ClosedXML-specific exceptions.

ClosedXML/Excel/AutoFilters/XLAutoFilter.cs
igitur: On .NET Framework, the evaluation will not continue past the first match, so it's similar to the old approach. Apparently this changed on .NET Core.

Pankraty: I don't think this changed in .Net Core; either way, it won't stop until reaching the end if there are no intersections. And if we'd have thousands of tables on the same worksheet the search would be relatively slow.
But since there are usually just a few tables on a worksheet I think this is not a problem.

igitur: I thought of this discussion that I saw the other day. https://github.com/dotnet/runtime/issues/31554
But seems it's applicable only on ordered enumerables. Still something to take note of.

ClosedXML_Tests/Excel/AutoFilters/AutoFilterTests.cs
igitur: Then probably also `InvalidOperationException` in `XLWorksheet.CheckRangeNotOverlappingOtherEntities()`, right?

Pankraty: Sure

PR #1445, opened 2020-05-25
Fix the incorrect FirstRowUsed method (#1443)

### LINKED ISSUE #1443: FirstRowUsed is calling LastRowUsed
description:
in XLWorksheet I had the issue in release 0.95.2 that FirstRowUsed is calling LastRowUsed
this is very bad and results in a null response at the end.

release 0.95.2

refernce to code
https://github.com/ClosedXML/ClosedXML/blob/6bc5cb51d14810a7d599f3c2095bf82d17964cb0/ClosedXML/Excel/XLWorksheet.cs#L270

Pankraty:
Fixes #1443 

This is not a true regression as in 0.94.2 it worked the same way, and before that there was no such method. But it is a nasty bag nonetheless.

### DIFF
--- a/ClosedXML/Excel/XLWorksheet.cs
+++ b/ClosedXML/Excel/XLWorksheet.cs
@@ -268,7 +268,7 @@ namespace ClosedXML.Excel
 
         IXLRow IXLWorksheet.FirstRowUsed(XLCellsUsedOptions options)
         {
-            return LastRowUsed(options);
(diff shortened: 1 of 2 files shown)


igitur:
Thanks. Although not a regression, maybe we should release this as a hotfix for 0.95.x.

Pankraty:
OK, created a new PR #1446 for the RC branch.

Issue #1450, opened 2020-05-31
The copied Comment is not equal to the original by Equal method


b0bi79:
**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

**Version of ClosedXML**

e.g. 0.95.3

**What is the current behavior?**

Comment.Equals returns FALSE for equal comments. **The problem is that the XLFormattedText.Equals method compares objects using `!=` instead of Equals.**

**What is the expected behavior or new feature?**

Comment.Equals returns TRUE for equal comments.

**Is this a regression from the previous version?**

## Reproducibility
**Code to reproduce problem:**
```c#
sourceCell.Comment.AddText("test")
targetCell.CopyFrom(sourceCell);
Assert.IsTrue(targetCell.Comment.Equals(sourceCell.Comment));
```

igitur:
I guess it's debatable whether 2 comments with the same contents are equal or not in the true sense. Can you give me a more real-world use case of why this matters?

PR #1451, opened 2020-05-31
#1450 comment equals

b0bi79:


ClosedXML/Excel/RichText/XLFormattedText.cs
```diff
@@ -169,7 +169,7 @@ public bool Equals(IXLFormattedText<T> other)
 
             for (Int32 i = 0; i < count; i++)
             {
-                if (_richTexts.ElementAt(i) != other.ElementAt(i))
+                if (_richTexts.ElementAt(i).Equals(other.ElementAt(i)))
```
igitur, line 172: Should the `==` and `!=` operators not be overloaded to avoid this situation?

Pankraty: Agree, it's a good practice to have a consistent behavior between Equals and ==

NicholasNoise: Possible NRE.

PR #1462, opened 2020-06-10
Fix possible multiple enumeration issue.

NicholasNoise:
Some of fixed enumerations may lead to significant performance improvement, others may be just overkill with no affect at all.
Similar affect as #1388 or https://github.com/ClosedXML/ClosedXML.Report/pull/162

https://stackoverflow.com/questions/4621561/is-there-a-way-to-memorize-or-materialize-an-ienumerable
Materialization is not a simple question, but I point to the simple answer.

### DIFF
--- a/ClosedXML/Excel/AutoFilters/XLFilterColumn.cs
+++ b/ClosedXML/Excel/AutoFilters/XLFilterColumn.cs
@@ -233,7 +233,7 @@ namespace ClosedXML.Excel
                                        .SetTopBottomValue(value)
                                        .SetTopBottomType(type);
 
-            var values = GetValues(value, type, takeTop);
+            var values = GetValues(value, type, takeTop).ToArray();
 
             Clear();
             _autoFilter.Filters.Add(_column, new List<XLFilter>());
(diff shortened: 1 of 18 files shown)


ClosedXML/Excel/AutoFilters/XLFilterColumn.cs
```diff
@@ -233,7 +233,7 @@ private void SetTopBottom(Int32 value, XLTopBottomType type, Boolean takeTop = t
                                        .SetTopBottomValue(value)
                                        .SetTopBottomType(type);
 
-            var values = GetValues(value, type, takeTop);
+            var values = GetValues(value, type, takeTop).ToArray();
```
igitur, line 236: `.ToArray()` obviously has some overhead, both in terms of CPU cycles and memory. So unless it reduces enumerations, I would like to avoid it. From what I can see, this enumerable is evaluated only once in line 248's `foreach (double val in values)`. So how would this `.ToArray()` improve performance?

Pankraty: The loop you are referring to is inside the other loop so this change makes sense.

NicholasNoise: @Pankraty has explained the idea.
In additional I would say: IEnumerable is a great thing to implement lazy computation, but then and only then we properly materialize encapsulated value.

jahav:
Closing in favor of #2236.

PR #1463, opened 2020-06-13
fix an odd precise comparison

makeProjectGreatAgain:
Tried to solve problem with comparing doubles to doubles using  `== `
Changed:
`left.Width.Value == right.Width.Value))`
to
`(Math.Abs(left.Width.Value - right.Width.Value) < tolerance)))`
with tolerance = 1e-10;

### DIFF
--- a/ClosedXML/Excel/XLWorkbook_Save.cs
+++ b/ClosedXML/Excel/XLWorkbook_Save.cs
@@ -56,7 +56,7 @@ namespace ClosedXML.Excel
     public partial class XLWorkbook
     {
         private const Double ColumnWidthOffset = 0.710625;
-        private static readonly double tolerance = 1e-10;
+
         //private Dictionary<IXLStyle, StyleInfo> context.SharedStyles;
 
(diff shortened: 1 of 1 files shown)


ClosedXML/Excel/XLWorkbook_Save.cs
```diff
@@ -56,7 +56,7 @@ namespace ClosedXML.Excel
     public partial class XLWorkbook
     {
         private const Double ColumnWidthOffset = 0.710625;
-
+        private static readonly double tolerance = 1e-10;
```
Pankraty, line 59: Use XLHelper.Epsilon instead

makeProjectGreatAgain: >Have you experienced any issues caused by that comparison?

No, I haven't

>Use XLHelper.Epsilon instead

Done

igitur:
> Have you experienced any issues caused by that comparison?

Would also like to know.

makeProjectGreatAgain:
> > Have you experienced any issues caused by that comparison?
> 
> Would also like to know.

I already answered this above. No, I haven't. Perhaps this changes aren't necessary

PR #1467, opened 2020-06-19
#1456 Add CreateSafeSheetName method

### LINKED ISSUE #1456: createSafeSheetName method

- [ ] Bug
- [x] Feature
- [ ] Question

- [x] Yes
- [ ] No

"Worksheet names cannot contain any of the following characters: "

Please add `createSafeSheetName` like POI
https://github.com/apache/poi/blob/f509d1deae86866ed531f10f2eba7db17e098473/src/java/org/apache/poi/ss/util/WorkbookUtil.java#L67

makeProjectGreatAgain:
Added CreateSafeSheetName method to **XLHelper** class.

Creates a valid sheet name, which is conform to the rules:
- minimum length is 1
- maximum length is 31
- doesn't contain special chars: : 0x0000, 0x0003, / \ ? * ] [
- Sheet names must not begin or end with ' (apostrophe)
- never null

Fixes #1456

### DIFF
--- a/ClosedXML/XLHelper_ASF.cs
+++ b/ClosedXML/XLHelper_ASF.cs
@@ -1,0 +1,113 @@
+﻿// Keep this file CodeMaid organised and cleaned
+
+/* ====================================================================
+   Licensed to the Apache Software Foundation (ASF) under one or more
+   contributor license agreements.  See the NOTICE file distributed with
+   this work for additional information regarding copyright ownership.
+   The ASF licenses this file to You under the Apache License, Version 2.0
+   (the "License"); you may not use this file except in compliance with
+   the License.  You may obtain a copy of the License at
+       http://www.apache.org/licenses/LICENSE-2.0
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+==================================================================== */
+
+using System;
+using System.Linq;
+
+namespace ClosedXML.Excel
+{
+    public static partial class XLHelper
+    {
+        private const int MaxWorksheetNameCharsCount = 31;
+        private static readonly char[] illegalWorksheetCharacters = "\u0000\u0003:\\/?*[]".ToCharArray();
+
+        /// <summary>
+        /// Creates a valid sheet name, which conforms to the following rules.
+        /// A sheet name:
+        /// is never null,
+        /// has minimum length of 1 and
+        /// maximum length of 31,
+        /// doesn't contain special chars: (: 0x0000 0x0003 / \ ? * ] [).
+        /// Sheet names must not begin or end with ' (apostrophe)
+        /// </summary>
+        /// <param name="nameProposal">can be any string, will be truncated if necessary, allowed to be null</param>
+        /// <param name="replaceChar">the char to replace invalid characters.</param>
+        /// <returns>a valid string, "empty" if too short, "null" if null</returns>
+        // This method was ported and adapted from the POI project at https://github.com/apache/poi/blob/trunk/src/java/org/apache/poi/ss/util/WorkbookUtil.java
+        public static string CreateSafeSheetName(string nameProposal, char replaceChar = ' ')
+        {
+            if (illegalWorksheetCharacters.Contains(replaceChar) || replaceChar == '\'')
+                throw new ArgumentException("Invalid replacement character.", nameof(replaceChar));
+
+            if (nameProposal == null)
+            {
+                return "null";
+            }
+            if (nameProposal.Length < 1)
+            {
+                return "empty";
+            }
+            int length = Math.Min(MaxWorksheetNameCharsCount, nameProposal.Length);
+            var shortenedName = nameProposal.Substring(0, length);
+            var result = new System.Text.StringBuilder(shortenedName);
+            for (int i = 0; i < length; i++)
+            {
+                char ch = result[i];
+                if (illegalWorksheetCharacters.Contains(result[i]))
+                    result[i] = replaceChar;
+
+                if (ch == '\'' && (i == 0 || i == length - 1))
+                    result[i] = replaceChar;
+            }
+            return result.ToString();
+        }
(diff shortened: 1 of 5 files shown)


igitur:
Thanks. Did you port over the Java code from POI? I'm not a lawyer, but if you did, there might be a license conflict between POI (Apache) and ClosedXML (MIT). @pankraty, do you know?

makeProjectGreatAgain:
Thanks for your notice. My bad.
Yes, I did. I'm not a lawyer too and I don't know how to fulfill all requirements. All what I can say is that it was from Tony Qu's [port of POI Java project](https://github.com/tonyqus/npoi) with Apache-2.0 License

igitur:
Hi @YegorKozlov . It looks you were the only author of the original `createSafeSheetName` method in POI. If I understand the legal things correctly, even though there there is an incompatibility between the license of POI (Apache) and ClosedXML (MIT), we could still port this specific method if you grant explicit permission. The 2 projects have similar goals, so would you please consider this and let us know?

igitur:
Received per email

>Hello Francois,
> 
> I'm the author but not not the owner of this code. 
> The ASF licence allows you to port code any way you want providing you put a notice the your code is based on POI .  ClosedXML here falls into the Derivative Works category of the ASF licence.
> 
> Regards,
> Yegor

igitur:
I pushed some changed to your branch.
Moved method to `XLHelper_ASF.cs` file. 
Also ported `ValidateSheetName`

ClosedXML/XLHelper_ASF.cs
```diff
@@ -0,0 +1,142 @@
...
+                    case '\u0000':
+                    case '\u0003':
+                    case ':':
+                    case '/':
+                    case '\\':
+                    case '?':
+                    case '*':
+                    case ']':
+                    case '[':
+                        result[i] = replaceChar;
```
Pankraty, line 65: Since the method is public, I'd suggest checking that `replaceChar` is itself valid (i.e. is not either of :/*? and so on).

ClosedXML/XLHelper_ASF.cs
```diff
@@ -0,0 +1,142 @@
...
+            {
+                return "null";
+            }
+            if (nameProposal.Length < 1)
+            {
+                return "empty";
+            }
+            int MaxWorksheetNameCharsCount = 31;
+            int length = Math.Min(MaxWorksheetNameCharsCount, nameProposal.Length);
+            var shortenname = nameProposal.Substring(0, length);
```
Pankraty, line 49: Should it not be `shortenName`?

ClosedXML/XLHelper_ASF.cs
```diff
@@ -0,0 +1,142 @@
...
+        {
+            if (nameProposal == null)
+            {
+                return "null";
+            }
+            if (nameProposal.Length < 1)
+            {
+                return "empty";
+            }
+            int MaxWorksheetNameCharsCount = 31;
```
Pankraty, line 47: `MaxWorksheetNameCharsCount` should either to become a constant (and used in `ValidateSheetName` too) or to be renamed to `maxWorksheetNameCharsCount`.

ClosedXML/XLHelper_ASF.cs
```diff
@@ -0,0 +1,142 @@
...
+    public static partial class XLHelper
+    {
+        /// <summary>
+        /// Creates a valid sheet name, which is conform to the rules.
+        /// Rules:
+        /// never null
+        /// minimum length is 1
+        /// maximum length is 31
+        /// doesn't contain special chars: : 0x0000, 0x0003, / \ ? * ] [
+        /// Sheet names must not begin or end with ' (apostrophe)
```
Pankraty, line 31: In tooltips, line breaks are removed and the text without separators becomes hard to read. Please add commas or bullet points
image

ClosedXML_Tests/Excel/Worksheets/XLWorksheetTests.cs
```diff
@@ -329,6 +331,22 @@ public void WorksheetNameCannotEndWithApostrophe()
...
+
+        [Test]
+        public void WorksheetNameCannotContainIllegalCharacters()
+        {
+            foreach (var c in IllegalWorksheetCharacters)
+            {
+                var proposedName = $"Sheet{c}Name";
+                Assert.Throws<ArgumentException>(() => new XLWorkbook().AddWorksheet(proposedName));
+            }
+        }
```
Pankraty, line 348:
```suggestion
        private static char[] IllegalWorksheetCharacters = @":\/?*[]".ToCharArray();

        [TestCaseSource(nameof(IllegalWorksheetCharacters))]
        public void WorksheetNameCannotContainIllegalCharacters(char c)
        {
            var proposedName = $"Sheet{c}Name";
            Assert.Throws<ArgumentException>(() => new XLWorkbook().AddWorksheet(proposedName));
        }
```

ClosedXML_Tests/XLHelperTests.cs
```diff
@@ -148,5 +148,62 @@ public void ReplaceRelativeC()
             string result = XLHelper.ReplaceRelative("$A:$A", 2, "B");
             Assert.AreEqual("$A:$A", result);
         }
+
+        [Test]
+        public void TestCreateSafeSheetNames()
```
Pankraty, line 153: I suggest generalizing this test providing a set of `TestCase`s
```
        [TestCase("data?", "data ")]
        ...
        public void TestCreateSafeSheetNames(string input, string expected)
        {
            var actual = XLHelper.CreateSafeSheetName(input);
            Assert.AreEqual(expected, actual);
        }
```

igitur:
@Pankraty Your code comments are valid. I ported the method directly from the original Java class and made it compile in .NET with the minimal changes. Do you think we should address all your code comments and thereby deviate from the original code?

Pankraty:
Sorry for the late response. I doubt we are going to maintain the strict correspondence between our class and the original Java code, therefore I don't see a big issue if our version will differ from the original.

igitur:
> Sorry for the late response. I doubt we are going to maintain the strict correspondence between our class and the original Java code, therefore I don't see a big issue if our version will differ from the original.

I went ahead and made the changes already.

Pankraty (approved):

PR #1471, opened 2020-07-06
IXLCell.TryGetValue improvements

### LINKED ISSUE #1469: Problem with reading cell value which has formula

- [x] Bug
- [] Feature
- [x] Question

- [x] Yes
- [ ] No

**Version of ClosedXML**
0.95.3

e.g. 0.95.3

If cell contains formula e.g. =Today()+2 etc. then we get the cell value as integer not the expected value.  

If the cell contains the date formula e.g. =Today()+2 then we expect to get 07/04/2020 as cell value.
I have question here as how I get the value of the cell after calculating the formula set on it.   
Complete this.

No. Older version also not giving proper value. 

Regressions get higher priority. Test against the latest build of the previous minor version. For example, if you experience a problem on v0.95.3, check whether it the problem occurred in v0.94.2 too. 

## Reproducibility

```c#
public void Main()
{
    string fileName = @"<File Name>";
    var xlworkBook = new XLWorkbook(fileName);
    if (xlworkBook != null)
    {
               var sheet1 = xlworkBook.Worksheet("Sheet1");
               if (sheet1 != null)
                {
                    var rows = sheet1.RangeUsed().RowsUsed().Skip(1);
                    foreach (var row in rows)
                    {
                    var rowCell = row.Cell(1);
                    var rowString = rowCell.GetDateTime();// This will throw error if cell contains formula e.g. =Today()+2 etc. 
                    }
                }
            }
}
```

[DateFormats.xlsx](https://github.com/ClosedXML/ClosedXML/files/4863569/DateFormats.xlsx)

igitur:
Fixes #1469 

Make `TryGetValue` try harder to coerce between data types.

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -1755,14 +1755,35 @@ namespace ClosedXML.Excel
                 return false;
             }
 
-            if (!DateTime.TryParse(currentValue.ToString(), out DateTime ts))
+            if (currentValue is T v) { value = v; return true; }
+
+            if (currentValue.IsNumber())
             {
-                value = default;
-                return false;
+                var dbl1 = Convert.ToDouble(currentValue);
+                if (dbl1.IsValidOADateNumber())
+                {
+                    value = (T)Convert.ChangeType(DateTime.FromOADate(dbl1), typeof(T));
+                    return true;
+                }
             }
 
-            value = (T)Convert.ChangeType(ts, typeof(T));
-            return true;
+            if (DateTime.TryParse(currentValue.ToString(), out DateTime ts))
+            {
+                value = (T)Convert.ChangeType(ts, typeof(T));
+                return true;
+            }
+
+            // If the cell value is a string, e.g. "42020", we could theoretically coerce it to a DateTime, but this seems to go against what is expected.
+            // Leaving this code block here, though. Maybe we revert our decision later.
+
+            //if (Double.TryParse(currentValue.ObjectToInvariantString(), out Double dbl2) && dbl2.IsValidOADateNumber())
+            //{
+            //    value = (T)Convert.ChangeType(DateTime.FromOADate(dbl2), typeof(T));
+            //    return true;
+            //}
+
+            value = default;
+            return false;
         }
 
         private static bool TryGetTimeSpanValue<T>(out T value, object currentValue)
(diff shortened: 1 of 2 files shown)


ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -1755,16 +1755,34 @@ private static bool TryGetDateTimeValue<T>(out T value, object currentValue)
...
+            if (currentValue.IsNumber())
             {
-                value = default;
-                return false;
+                var dbl1 = Convert.ToDouble(currentValue);
+                if (dbl1.IsValidOADateNumber())
+                {
+                    value = (T)Convert.ChangeType(DateTime.FromOADate(dbl1), typeof(T));
+                    return true;
+            }
```
Pankraty, line 1767: The indentation is screwed up here, probably during the rebase.

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -1755,16 +1755,34 @@ private static bool TryGetDateTimeValue<T>(out T value, object currentValue)
...
+            {
             value = (T)Convert.ChangeType(ts, typeof(T));
             return true;
         }
 
+            if (Double.TryParse(currentValue.ObjectToInvariantString(), out Double dbl2) && dbl2.IsValidOADateNumber())
+            {
+                value = (T)Convert.ChangeType(DateTime.FromOADate(dbl2), typeof(T));
+                return true;
+            }
```
Pankraty, line 1780: This block is not covered by tests; there might be something like this:
```
[Test]
public void TryGetValue_DateTime_Good3()
{
    IXLWorksheet ws = new XLWorkbook().Worksheets.Add("Sheet1");
    bool success = ws.Cell("A1").SetFormulaA1("=\"44\"&\"020\"").TryGetValue(out DateTime outValue);
    Assert.IsTrue(success);
    Assert.AreEqual(new DateTime(2020, 07, 08), outValue);
}
```
But I am not 100% sure this should work this way. Excel, for example, leaves text to be text, even if a cell type is Date

image

igitur: But this does work in Excel:
image

We could possibly move the string to Date conversion into the calculation engine, but I don't really see a down-side to leaving it here. If the user calls `cell.GetValue<DateTime>()` then I assume he really wants to get out a DateTime as much as possible.

igitur: After thinking about it over the weekend, I now agree that we shouldn't do a "double-hop" to coerce values from one type to another. Converting a string to a number and then a number to a date, would be a "double-hop". This is only for the `TryGetValue<>()` cases. In the calc engine, coercing can be a bit more lenient, as is already the case. See updated unit test: `TryGetValue_DateTime_BadButFormulaGood`

PR #1472, opened 2020-07-06
Support IEnumerable for IXLCell.SetValue()

### LINKED ISSUE #1470: XLCell.SetValue does not handle IEnumerable
- [x] Bug
- [ ] Feature
- [ ] Question

- [ ] Yes
- [x] No

**Version of ClosedXML**

0.95.3

The documentation for `IXLCell.SetValue` says it handles an `IEnumerable` similar to using `IXLCell.Value` to set the value, but it throws an Exception instead.

Either the documentation of `SetValue` should not say it accepts `IEnumerable` or it should have the same behaviour as assigning to `Value` for an `IEnumerable`.

No, same behaviour in 0.94.2

## Reproducibility

```c#
public void Main()
{
    var wb = new XLWorkbook();
    var ws = wb.Worksheets.Add("Collections");

    var listOfStrings = new List<String>();
    listOfStrings.Add("House");
    listOfStrings.Add("Car");
    ws.Cell(1, 1).SetValue(listOfStrings); // throws Exception with message "Unable to set cell value to System.Collections.Generic.List`1[System.String]"
}
```

igitur:
Fixes #1470

### DIFF
--- a/ClosedXML_Sandbox/PerformanceRunner.cs
+++ b/ClosedXML_Sandbox/PerformanceRunner.cs
@@ -42,13 +42,13 @@ namespace ClosedXML_Sandbox
 
         public static void OpenTestFile()
         {
-            using (var wb = new XLWorkbook("test.xlsx"))
-            {
-                wb.RecalculateAllFormulas();
-                var ws = wb.Worksheets.First();
-                var cell = ws.FirstCellUsed();
-                Console.WriteLine(cell.Value);
-            }
+            var wb = new XLWorkbook();
+            var ws = wb.Worksheets.Add("Collections");
+
+            var listOfStrings = new List<String>();
+            listOfStrings.Add("House");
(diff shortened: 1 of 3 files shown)


ClosedXML_Sandbox/PerformanceRunner.cs
```diff
@@ -42,13 +42,13 @@ public static void RunInsertTable()
...
         public static void OpenTestFile()
         {
-            using (var wb = new XLWorkbook("test.xlsx"))
-            {
-                wb.RecalculateAllFormulas();
-                var ws = wb.Worksheets.First();
-                var cell = ws.FirstCellUsed();
-                Console.WriteLine(cell.Value);
-            }
+            var wb = new XLWorkbook();
```
igitur, line 45: @Pankraty You missed this accidental change! :-P

PR #1485, opened 2020-08-03
Improve Rows() method performance (#1482)

### LINKED ISSUE #1482: accessing rows using elementAt is extremely slow on large spreadsheets

- [x] Bug
- [ ] Feature
- [ ] Question

- [X] Yes
- [ ] No

**Version of ClosedXML**
0.95.3

On a spreadsheet with 80,000 + rows, I called this:
var cell = worksheet.Rows().ElementAt(i).Cell(1).Value?.ToString().Trim();
the access time is about 3 seconds.

For it to take much less time. Instead, I had to use a foreach loop.

Yes, I saw it happening in 0.94.2 as well.

## Reproducibility

        public void TestSpreadsheet()
        {
            using (var workBook = new XLWorkbook("C:/test spreadsheet.xlsx"))
            {
                var worksheet = workBook.Worksheets.ElementAt(0);
                var row = worksheet.Rows().ElementAt(1);
                var cell = row.Cell(0).Value?.ToString().Trim();
            }
        }

Pankraty:
Fixes #1482 

Benchmark:
```
var number = 1000;
for (int i = 0; i < 10; i++)
{
    using var ms = new MemoryStream();
    var data = Enumerable.Range(0, number).Select(i => Guid.NewGuid());
    var wb = new XLWorkbook();
    wb.AddWorksheet().FirstCell().InsertData(data);
    wb.SaveAs(ms);

    using var workBook = new XLWorkbook(ms);
    var sw = new Stopwatch();
    sw.Start();
    var worksheet = workBook.Worksheets.ElementAt(0);
    var row = worksheet.Rows().ElementAt(1);
    var cell = row.Cell(1).Value?.ToString().Trim();
    Console.WriteLine($"{number}\t{sw.ElapsedMilliseconds}");
    number *= 2;
}
```
Output:
```
1000    3
2000    3
4000    10
8000    42
16000   174
32000   757
64000   2759
128000  10737
256000  45718
512000  188466
```

### DIFF
--- a/ClosedXML/Excel/Columns/XLColumns.cs
+++ b/ClosedXML/Excel/Columns/XLColumns.cs
@@ -8,8 +8,12 @@ namespace ClosedXML.Excel
 
     internal class XLColumns : XLStylizedBase, IXLColumns, IXLStylized
     {
-        private readonly List<XLColumn> _columns = new List<XLColumn>();
+        private readonly List<XLColumn> _columnsCollection = new List<XLColumn>();
         private readonly XLWorksheet _worksheet;
+        private bool IsMaterialized => _lazyEnumerable == null;
+
+        private IEnumerable<XLColumn> _lazyEnumerable;
+        private IEnumerable<XLColumn> Columns => _lazyEnumerable ?? _columnsCollection.AsEnumerable();
 
         /// <summary>
         /// Create a new instance of <see cref="XLColumns"/>.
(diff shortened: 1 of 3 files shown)


Pankraty:
Step 1 - replace a list with a hashset:
```
1000    2
2000    1
4000    1
8000    2
16000   12
32000   13
64000   29
128000  63
256000  142
512000  274
```

igitur:
Great, thanks. Is it necessary to base this on the `InsertData` refactoring?

Pankraty:
Not at all. It is just a draft, I've created it to not lose the intermediate results. Will do a rebase when I'm done.

Pankraty:
Step 2 - support lazily enumerated collections in `XLRows` (-30% to execution time)
```
512000  193
```

Pankraty:
I am not 100% sure we have to apply both improvements; maybe the first commit is good enough as it keeps the implementation simpler.

SimantoR:
Is this ready for merge or does it need more work? I would love to help.

SimantoR:
Amazing boost! Opened, parsed and close ~1 mil rows in 1 column in 10s on my system. Iterating over two rows as two separate tasks came to ~18s. I wonder if this is multi-thread ready.

Pankraty:
ClosedXML is not thread-safe, that's for sure.

SimantoR:
> ClosedXML is not thread-safe, that's for sure.

I wonder what it might take. I'll try reading through the repository but what do you think needs to change for that? If the whole pdf should be in memory, we could try to slice it without allocating more memory if we can slice it properly. Or did I get it absurdly wrong?

Pankraty:
If we speak about reading only, it'll take changing internal collections (such as Dictionary) to thread-safe analogues. And I think our CalcEngine is statefull which also have to be changed.
As for writing, it would be much harder, especially when inserting/deleting ranges, rows, columns involved.
And you be honest, I doubt that with multi-threading we'd benefit much, because most of the cases performance is low are related to large number of entities processed in memory and/or inefficient algorithms (which we try to improve when notice their inefficiency). By introducing concurrency we'd only slow down that processing. 
More beneficial, in my opinion, would be to introduce some sort of delayed reading - say, if the user only reads the data from a single worksheet, there is no need to read other worksheets, styles, conditional formats, etc. But this requires a complete refactoring of the most complicated part - loading and saving - and I am not sure is possible without breaking the existing API.

SimantoR:
That sounds like a great plan. It would help in specific cases where the time to load each worksheet is spread out. Maybe a flag to specify lazy loading?

igitur (approved):

PR #1488, opened 2020-08-04
UnaryExpression will now handle string values properly

### LINKED ISSUE #1487: UnaryExpression does not handle strings

- [x] Bug
- [ ] Feature
- [ ] Question

- [x] Yes
- [ ] No

**Version of ClosedXML**

0.95.3

When placed inside an expression, +A1 or -A1 will return 0 if A1 is a string

The Excel behavior is as follows:

Leading + should be ignored. CalcEngine.Parse already does that:
            ```
// skip leading +'s
while (_len > _ptr && _expr[_ptr] == '+')
    _ptr++;
```

- String = #VALUE in Excel

No

## Reproducibility

```c#
public void Main()
{
    using (var wb = new ClosedXML.Excel.XLWorkbook())
    {
        ws.Cell("A1").Value = "StringValue";

        // Must be within another expression (IF) because parser will strip + if by itself (eg: =+A1)
        var plusString = ws.Evaluate("=IF(1 = 1, +A1, +A1)");
        // plusString = 0                

        var minusString = ws.Evaluate("IF(1 = 1, -A1, -A1)");
        // minusTring = 0
    }
}
```

mathlang:
Plus will not be ignored from UnaryExpression
Minus string that can't be parsed to double will not properly throw CellValueException

Fixes #1487

### DIFF
--- a/ClosedXML_Tests/Excel/CalcEngine/ExpressionTests.cs
+++ b/ClosedXML_Tests/Excel/CalcEngine/ExpressionTests.cs
@@ -1,0 +1,44 @@
+﻿using ClosedXML.Excel;
+using ClosedXML.Excel.CalcEngine;
+using ClosedXML.Excel.CalcEngine.Exceptions;
+using NUnit.Framework;
+using System;
+using System.Globalization;
+using System.Threading;
+
+namespace ClosedXML_Tests.Excel.CalcEngine
+{
+    [TestFixture]
+    public class ExpressionTests
+    {
+        [OneTimeSetUp]
+        public void SetCultureInfo()
+        {
+            Thread.CurrentThread.CurrentCulture = CultureInfo.CreateSpecificCulture("en-US");
+        }
+
+        [Test]
+        public void UnaryExpression()
+        {
+            // String
+            Assert.AreEqual(XLWorkbook.EvaluateExpr("=+\"TestString\""), "TestString"); // Plus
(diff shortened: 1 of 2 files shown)


igitur (commented):
Thanks. Shouldn't there be a test for the TKID.ADD case too?

mathlang:
> Thanks. Shouldn't there be a test for the TKID.ADD case too?

Done, I added tests for other data types as well

ClosedXML_Tests/Excel/CalcEngine/ExpressionTests.cs
```diff
@@ -0,0 +1,44 @@
...
+    public class ExpressionTests
+    {
+        [OneTimeSetUp]
+        public void SetCultureInfo()
+        {
+            Thread.CurrentThread.CurrentCulture = CultureInfo.CreateSpecificCulture("en-US");
+        }
+
+        [Test]
+        public void UnaryExpression()
```
igitur, line 21: In this test the expected and actual parameters are the wrong way around. First parameter should be expected value.

igitur:
Thanks. I'm not yet convinced by this PR. In dd04e51adb83c2f6d30023044eb52794d0332ceb there was a deliberate change to treat strings as `0`. I understand the problem, but I think the solution should be in another part of the code. Your current solution results in parsing the string twice: in `double.TryParse` and then again in the implicit operator in `(double)Expression`. Let me think a bit more about this.

jahav:
Closing because done in #1812.

PR #1510, opened 2020-09-03
KnownTypeMapping

cwuethrich:
This change allows us to choose the known type for types which are `IConvertible`.

We have a type called `Date`. A type mapping allows us to convert properties of this type to the system type `DateTime`.

It is also an extension to issue #953. At the moment an `Enum` is handled as a `String`. With this short extension we can choose between `String`, `Int32` or whatever we cant to.

### DIFF
--- a/ClosedXML/Excel/XLWorkbook.cs
+++ b/ClosedXML/Excel/XLWorkbook.cs
@@ -1,4 +1,4 @@
-using ClosedXML.Excel.CalcEngine;
+﻿using ClosedXML.Excel.CalcEngine;
 using DocumentFormat.OpenXml;
 using System;
 using System.Collections.Generic;
@@ -195,6 +195,11 @@ namespace ClosedXML.Excel
         ///   <para>All new worksheets will use these outline options.</para>
         /// </summary>
         public IXLOutline Outline { get; set; }
+
+        /// <summary>
+        ///   Gets the set of known type mappings.
+        /// </summary>
+        public IDictionary<Type, Type> KnownTypeMapping { get; } = new Dictionary<Type, Type>();
 
         /// <summary>
         ///   Gets or sets the workbook's properties.
--- a/ClosedXML/Excel/XLWorkbook.cs
+++ b/ClosedXML/Excel/XLWorkbook.cs
@@ -1,4 +1,4 @@
-﻿using ClosedXML.Excel.CalcEngine;
+using ClosedXML.Excel.CalcEngine;
 using DocumentFormat.OpenXml;
 using System;
 using System.Collections.Generic;
--- a/ClosedXML_Tests/Excel/Cells/XLCellTests.cs
+++ b/ClosedXML_Tests/Excel/Cells/XLCellTests.cs
@@ -1,4 +1,4 @@
-﻿using ClosedXML.Excel;
+using ClosedXML.Excel;
 using NUnit.Framework;
 using System;
 using System.Collections.Generic;
--- a/ClosedXML_Tests/Excel/Cells/XLCellTests.cs
+++ b/ClosedXML_Tests/Excel/Cells/XLCellTests.cs
@@ -1,4 +1,4 @@
-using ClosedXML.Excel;
+﻿using ClosedXML.Excel;
 using NUnit.Framework;
 using System;
 using System.Collections.Generic;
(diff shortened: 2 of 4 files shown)


igitur:
Please rebase this on the latest `develop` branch.

igitur (changes requested):

ClosedXML/Excel/XLWorkbook.cs
```diff
@@ -1,4 +1,4 @@
-using ClosedXML.Excel.CalcEngine;
+﻿using ClosedXML.Excel.CalcEngine;
```
igitur, line 1: Please remove the `BOM` mark from the file.

ClosedXML_Tests/Excel/Cells/XLCellTests.cs
```diff
@@ -1,4 +1,4 @@
-using ClosedXML.Excel;
+﻿using ClosedXML.Excel;
```
igitur, line 1: Please remove the `BOM` mark from the file.

ClosedXML_Tests/Excel/Cells/XLCellTests.cs
```diff
@@ -192,6 +192,23 @@ public void InsertData_with_Nulls_IEnumerable()
             Assert.AreEqual(String.Empty, ws.Cell("A5").Value);
         }
 
+        [Test]
+        public void InsertData_Objects_with_KnownTypeMapping()
```
igitur, line 196: Besides `enum`s that can be either `int` or `string`, are there other use cases of this enhancement?

igitur (changes requested):

ClosedXML_Tests/Excel/Cells/XLCellTests.cs
cwuethrich: As mentioned we do have a type called `Date`. But this is a specific struct.

With version 0.95.3 this type got exported as a string. We can't export the data with the newest version. There is always the exception "Unable to set cell value to...". With this small extension we could export it as `DateTime`.

ClosedXML/Excel/XLWorkbook.cs
```diff
@@ -196,6 +196,11 @@ public IXLWorksheets Worksheets
         /// </summary>
         public IXLOutline Outline { get; set; }
 
+        /// <summary>
+        ///   Gets the set of known type mappings.
+        /// </summary>
+        public IDictionary<Type, Type> KnownTypeMapping { get; } = new Dictionary<Type, Type>();
```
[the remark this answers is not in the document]
cwuethrich: Sorry for that, it's done now.

ClosedXML_Tests/Excel/Cells/XLCellTests.cs
cwuethrich: I can make a test with `Date` as well. I'm just not sure if it is worth it to include this type.

igitur: Yes, please do the `Date` struct as well so that I can understand. @Pankraty has some alternative ideas on how to handle this.

igitur:
Are you experiencing this hurdle only with `InsertData` or `InsertTable`? Because I don't see why we need special handling when a user does `cell.Value = enumValue;`. In those cases the user can do type conversion on his own, don't you agree?

igitur:
@cwuethrich Please also allow edits by maintainers on this PR.

cwuethrich:
> Are you experiencing this hurdle only with `InsertData` or `InsertTable`? Because I don't see why we need special handling when a user does `cell.Value = enumValue;`. In those cases the user can do type conversion on his own, don't you agree?

Yes I agree. And for `DataTable` it should be enough to implement the types of ADO.Net. But we have this problem with `List<T>`.

igitur:
> Yes I agree. And for `DataTable` it should be enough to implement the types of ADO.Net. But we have this problem with `List<T>`.

Can I ask that you open an issue first that highlights all the problems that you're encountering. Once we have all the examples, we can think of the solution.

cwuethrich:
I gonna open an issue a few days. And i will close this pull request after that.

ClosedXML_Tests/Excel/Cells/XLCellTests.cs
cwuethrich: Sorry, with Version 0.92.1 the unknown types got exported as string. 0.95.3 is the newest version.

igitur:
By the way, an easy work-around for this is just to use LINQ and project your datatable onto a list of anonymous objects that hold the correct values.

Something like this:

```c#
cell.InsertData(dataTable.Rows.Cast<DataRow>().Select(r => new { Id = (int)row["Id"], EnumInt = (int)row["SomeEnum"], EnumString = (string)row["SomeEnum"] });
```

vbjay:
https://docs.microsoft.com/en-us/dotnet/api/system.data.datatableextensions.asenumerable?view=netcore-3.1

vbjay:
https://www.dotnetperls.com/datarow-field

igitur:
Thanks @vbjay I wasn't aware of those.  So even more options for workarounds.

vbjay:
Sure. You can combine both to get an Ienumerable<DataRow> and from that get
x fields from each row.

cwuethrich:
I don't use unknown types in `DataRow`. Also because the serialization of them is really bad. There is the version of the assembly included.

I created an issue and I created a new branch on a personal repository.

PR #1519, opened 2020-09-14
Add FISHER formula

iamartyom:
Implementation FISHER formula with test cases

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Functions/Statistical.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/Statistical.cs
@@ -210,7 +210,7 @@ namespace ClosedXML.Excel.CalcEngine
         private static object Fisher(List<Expression> p)
         {
             if (!double.TryParse(p?[0].Evaluate().ToString(), out double value)) throw new CellValueException("Parameter non numeric.");
-            if (value <= -1 || value >= 1) throw new NumberException("Incorrect value. Should be: -1 >= x <= 1.");
+            if (value <= -1 || value >= 1) throw new NumberException("Incorrect value. Should be: -1 > x < 1.");
 
             return 0.5 * Math.Log((1 + value) / (1 - value));
         }
--- a/ClosedXML/Excel/CalcEngine/Functions/Statistical.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/Statistical.cs
@@ -31,7 +31,7 @@ namespace ClosedXML.Excel.CalcEngine
             //EXPONDIST	Returns the exponential distribution
             //FDIST	Returns the F probability distribution
             //FINV	Returns the inverse of the F probability distribution
-            //FISHER	Returns the Fisher transformation
+            ce.RegisterFunction("FISHER", 1, Fisher); // Returns the Fisher transformation
             //FISHERINV	Returns the inverse of the Fisher transformation
             //FORECAST	Returns a value along a linear trend
             //FREQUENCY	Returns a frequency distribution as a vertical array
@@ -205,6 +205,11 @@ namespace ClosedXML.Excel.CalcEngine
 
             // done
             return count;
+        }
+
+        private static object Fisher(List<Expression> p)
+        {
+            return GetTally(p, true).Fisher();
         }
 
         private static object Max(List<Expression> p)
--- a/ClosedXML/Excel/CalcEngine/Functions/Statistical.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/Statistical.cs
@@ -209,7 +209,10 @@ namespace ClosedXML.Excel.CalcEngine
 
         private static object Fisher(List<Expression> p)
         {
-            return GetTally(p, true).Fisher();
+            if (!double.TryParse(p?[0].Evaluate().ToString(), out double value)) throw new CellValueException("Parameter non numeric.");
+            if (value <= -1 || value >= 1) throw new NumberException("Incorrect value. Should be: -1 >= x <= 1.");
+
+            return 0.5 * Math.Log((1 + value) / (1 - value));
         }
 
         private static object Max(List<Expression> p)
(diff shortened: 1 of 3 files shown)


ClosedXML/Excel/CalcEngine/Functions/Statistical.cs
```diff
@@ -207,6 +207,11 @@ private static object CountIfs(List<Expression> p)
             return count;
         }
 
+        private static object Fisher(List<Expression> p)
+        {
+            return GetTally(p, true).Fisher();
```
igitur, line 212: Since this function takes only 1 parameter, I don't think it's necessary to use the `Tally` class, right?

ClosedXML/Excel/CalcEngine/Functions/Tally.cs
```diff
@@ -165,6 +165,15 @@ public double Min()
             return nums.Length == 0 ? 0 : nums.Min();
         }
 
+        public double Fisher()
+        {
+            var nums = NumericValuesInternal();
+            if (nums.Length == 0) throw new ArgumentException("Parameter non numeric.");
```
igitur, line 171: Check the online Excel documentation for the function and throw one of the appropriate ClosedXML exceptions in the `ClosedXML.Excel.CalcEngine.Exceptions` namespace.

ClosedXML_Tests/Excel/CalcEngine/StatisticalTests.cs
```diff
@@ -242,6 +242,46 @@ public void Init()
             workbook = SetupWorkbook();
         }
 
+        [Test]
+        public void Fisher()
+        {
+            var ws = workbook.Worksheets.First();
+            double value;
+
+            Assert.That(() => ws.Evaluate(@"=FISHER(D3:D45)"),
```
igitur, line 251: Split into a test for the exceptions and a test for the calculated values, using a parametric test, i.e. use the `TestCase` attribute. See other tests for examples.

igitur (changes requested):
Thanks for this. Please review my comments and let me know if you agree.

ClosedXML/Excel/CalcEngine/Functions/Tally.cs
iamartyom: I fixed this. I don't see this specific exceptions initially.

ClosedXML_Tests/Excel/CalcEngine/StatisticalTests.cs
iamartyom: I changed tests for correct cases use for this [TestCase] attribute. Should we also separate tests for exceptions?

ClosedXML/Excel/CalcEngine/Functions/Statistical.cs
iamartyom: In this case, we are using the Tally class, which has a some useful logic, so I used that. In this case, it filters out non-numeric parameters. I think we should use it.

Another thing is that this class may have a somewhat inappropriate name and violates the principle of single responsibility. I think it make sense in future move logic to statistical function in Statistical class and logic for check parameters leave on this class and after this rename it. It make sense?

iamartyom: Initially I don't correct understand you and I removed my old message.

ClosedXML/Excel/CalcEngine/Functions/Tally.cs
iamartyom: Now I updated this code and I don't use now Tally class.
This is working fine but I don't sure about this:
`double.TryParse(p?[0].Evaluate().ToString(), out double value)`
Is this a good solution for checking if a value is a double? I don't sure. If not please tell me about good way and I will update this check. In this moment I don't good know this codebase. Thanks.

ClosedXML/Excel/CalcEngine/Functions/Statistical.cs
iamartyom: Now I updated this code and I don't use now Tally class.

This is working fine but I don't sure about this:
`double.TryParse(p?[0].Evaluate().ToString(), out double value)`
Is this a good solution for checking if a value is a double? I don't sure. If not please tell me about good way and I will update this check. In this moment I don't good know this codebase.

Thanks.

ClosedXML_Tests/Excel/CalcEngine/StatisticalTests.cs
```diff
@@ -242,6 +242,40 @@ public void Init()
...
+                Throws.TypeOf<CellValueException>().With.Message.EqualTo("Parameter non numeric."));
+
+            Assert.That(() => ws.Evaluate(@"=FISHER(H3)"),
+                Throws.TypeOf<NumberException>().With.Message.EqualTo("Incorrect value. Should be: -1 >= x <= 1."));
+
+            Assert.That(() => ws.Evaluate(@"=FISHER(-1)"),
+                Throws.TypeOf<NumberException>().With.Message.EqualTo("Incorrect value. Should be: -1 >= x <= 1."));
+
+            Assert.That(() => ws.Evaluate(@"=FISHER(1)"),
+                Throws.TypeOf<NumberException>().With.Message.EqualTo("Incorrect value. Should be: -1 >= x <= 1."));
```
Pankraty, line 263: If `-1` and `1` are invalid arguments why there are `>=` and `<=` in the error message? I'd expect `>` and `<` there.

And yes, better make it a single test with multiple `TestCase`s.

iamartyom: @Pankraty Yep. You are right about message. I fixed it and improve test for incorrect cases. Please check.

igitur:
I'd like to have #1523 merged first so that we can treat non-numerical parameters better here.

igitur:
Thanks for the contribution. I'll complete the rest when #1523 is merged.

PR #1523, opened 2020-09-17
Coercion conventions

igitur:
Something that comes up time and again is that somethings strings are compared with case INsensitivity. Sometimes not. Sometimes strings should be treated as 0. Sometimes only empty strings should be treated as 0.

This PR adds `CoercionConvention` that is an optional parameter to all `Expression` (sub)classes. Default to `CoercionConvention.NullAsZero`, which is the current behaviour. 

Now, when evaluation expressions, the convention can be passed through or overridden with `Expression.WithCoercionConvention()` for one-time alternative behaviour.

Also adds a global handler for type conversion exceptions and throws `CellValueException` it its place.

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Expression.cs
+++ b/ClosedXML/Excel/CalcEngine/Expression.cs
@@ -8,9 +8,25 @@ using System.Threading;
 
 namespace ClosedXML.Excel.CalcEngine
 {
+    [Flags]
+    internal enum CoercionConvention
+    {
+        None = 0,
+        EmptyStringAsZero = 1 << 0,
+        NonEmptyStringAsZero = 1 << 1,
+        NullAsZero = 1 << 2,
+        CaseInsensitive = 1 << 3,
+
+        Default = NullAsZero,
+        AnyStringAsZero = EmptyStringAsZero | NonEmptyStringAsZero,
+        StringOrNullAsZero = AnyStringAsZero | NullAsZero
+    }
+
     internal abstract class ExpressionBase
     {
         public abstract string LastParseItem { get; }
+
+        public abstract Expression WithCoercionConvention(CoercionConvention convention);
     }
 
     /// <summary>
@@ -29,6 +45,7 @@ namespace ClosedXML.Excel.CalcEngine
 
         #region ** fields
 
+        protected readonly CoercionConvention _convention;
         internal readonly Token _token;
 
         #endregion ** fields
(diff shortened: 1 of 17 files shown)


ClosedXML/Excel/CalcEngine/Expression.cs
```diff
@@ -2,15 +2,29 @@
...
+    {
+        None = 0,
+        EmptyStringAsZero = 1 << 0,
+        NonEmptyStringAsZero = 1 << 1,
+        NullAsZero = 1 << 2,
+        CaseInsensitive = 1 << 3,
+
+        AnyStringAsZero = EmptyStringAsZero | NonEmptyStringAsZero,
+        StringOrNullAsZero = AnyStringAsZero | NullAsZero
+    }
```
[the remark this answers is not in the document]
igitur: Good idea.

igitur:
Need to address #1530 in this.

igitur:
It's not entirely as elegant as I hoped, but at least this gives us the flexibility that we needed. So now, we can treat empty strings as 0 when we need to and not when we don't.

One can now use the old way for expression coersion, e.g. `double value = (double)p[0];` or coerce more explicitly with `double value = p[0].Coerce<double>(coercionConvention);`

ClosedXML/Excel/CalcEngine/Expression.cs
```diff
@@ -2,15 +2,29 @@
...
+
     internal abstract class ExpressionBase
     {
         public abstract string LastParseItem { get; }
+
+        public abstract Expression WithCoercionConvention(CoercionConvention convention);
     }
```
Pankraty, line 28: I wonder, what is the purpose of this class. It is not used anywhere except from the `Expression` class, and `WithCoercionConvention` returns an instance of `Expression` and not `ExpressionBase` as one would expect. Can't we just drop it and make corresponding members of `Expression` virtual?

ClosedXML/Excel/CalcEngine/Functions/DateAndTime.cs
```diff
@@ -142,9 +142,12 @@ private static object Datedif(List<Expression> p)
 
         private static object Datevalue(List<Expression> p)
```
Pankraty, line 143: Rename to `DateValue`

Pankraty: Also `DateDif` a few lines above

Pankraty (approved):
Approved with a couple of notes for your consideration.

jahav:
I am closing because type coercion is already mostly implemented #1812 . The code in PR is no longer mergable and coercion is solved a different way.

Issue #1511, opened 2020-09-03
Can't load pivot table with borders


cwuethrich:
## Read and complete the full issue template

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

**Version of ClosedXML**

0.95.3

**What is the current behavior?**

Can't load a pivot table with borders. Already `new XLWorkbook("PivotWithBorder.xlsx")` fails.

**What is the expected behavior or new feature?**

Also a pivot table with borders can be loaded with a `Workbook`.

**Is this a regression from the previous version?**

It worked with 0.94.2, Doesn't work since 0.95.0. 

## Reproducibility

**Code to reproduce problem:**
```c#
public void Main()
{
    var wb = new XLWorkbook(@"C:\temp\PivotWithBorder.xlsx");
}
```

- [X] I attached a sample spreadsheet.  (You can drag files on to this issue)
[PivotWithBorder.xlsx](https://github.com/ClosedXML/ClosedXML/files/5171168/PivotWithBorder.xlsx)

SfWt:
I can reproduce the failure with the provided file in my app using ClosedXml 0.95.3 also.

From my investigation the value of the "_container" element of the _ClosedXML.Excel.XLBorder_ class is null and therefore calls to the _Set..._ methods are failing with a NullReferenceException. I did the following change to the XLBorder constructor and the file was loaded successfully. But because I'm not experienced enough with the CloseXML code I'm not sure if this is just a workaround.

``` C#
public XLBorder(IXLStylized container, XLStyle style, XLBorderValue value)
{
    //_container = container;
    _container = container ?? new XLStylizedEmpty(style ?? XLStyle.Default);
    _style = style ?? _container.Style as XLStyle ?? XLStyle.CreateEmptyStyle();
    _value = value;
}
```

Maybe this can be of any help in fixing this issue.

igitur:
Pivot tables are still work in progress and pivot table formatting is even more experimental.

cwuethrich:
> Pivot tables are still work in progress and pivot table formatting is even more experimental.

I don't really use pivot tables. But o have to load a workbook with a pivot table, because I have to load data from another worksheet. So with this exception it isn't possible to load the workbook.

jahav:
Resolved in #1928

PR #1529, opened 2020-09-21
use XLStylizedEmpty in pivot table

cwuethrich:
At least don't throw an exception when loading a pivot table with borders. #1511


cwuethrich:
Is it possible to include this change? in our case, we don't use the styling of the table. But if there is a style, we can't load the whole file.

binn:
Any update on this merge?

jahav (commented):
The core bug is that fluent API for border always expects container. 

Not all types that have a style use IStylized semantic (that is the type of the `_container`). IStylized expects a hierarchical structure with ranges, but pivot table styles are more like a set of conditions that apply differential styles to all cells in the pivot table. When user changes pivot table report, conditions are reevaluated and style the user sees changes.

So basically replace  `return _container.Style;` with `return _style;` in the XLBorder.

jahav:
I can't add commits to a PR from a fork, I added them in #1928 with commit from this PR. File doesn't crash on load and I get some border properties.

PR #1539, opened 2020-10-06
Fix call calculated value invalidation if named ranges are changed (#1536)

### LINKED ISSUE #1536: IXLCell.Value returns incorrect value when using formula that references variable that was not yet added.
**NOTE**: This same code throws `ClosedXML.Excel.CalcEngine.Exceptions.NameNotRecognizedException: 'The identifier `variableA` was not recognised.'` on build from 'develop' branch. I think this was changed (is that now as designed?) in this commit:
https://github.com/ClosedXML/ClosedXML/commit/f2063a73cdd7c4d95f920de4241e95858e4b3b77

- [x] Bug
- [ ] Feature
- [ ] Question

- [ ] Yes
- [x] No

I used the latest officially released Nuget package v = `0.95.3`

**Version of ClosedXML**

Nuget package v = `0.95.3`

In the current released `0.95.3` version we have an issue that if formula has usage of named variable and we access `.Value` before this named variable is added to the worksheet and we try to access `formulaCell.Value` after we added named variable to `worksheet` we get `.Value` incorrectly as old value. Example

```c#
var wb = new XLWorkbook();
var ws = wb.Worksheets.Add("Formulas");

IXLCell formulaCell = ws.Cell(1, 1);
formulaCell.FormulaA1 = $"=variableA*5";

Console.WriteLine($"Expected = 0, Actual = {formulaCell.Value}"); //gets 0

ws.Cell(1, 2).Value = "5";
ws.Cell(1, 2).AddToNamed("variableA");

formulaCell.InvalidateFormula();
ws.RecalculateAllFormulas();

Console.WriteLine($"Expected = 25, Actual = {formulaCell.Value}"); //gets 0
```

Note that this works fine if I comment out first `formulaCell.Value` access:
```c#
var wb = new XLWorkbook();
var ws = wb.Worksheets.Add("Formulas");

IXLCell formulaCell = ws.Cell(1, 1);
formulaCell.FormulaA1 = $"=variableA*5";

//Console.WriteLine($"Expected = 0, Actual = {formulaCell.Value}"); // don't make access here

ws.Cell(1, 2).Value = "5";
ws.Cell(1, 2).AddToNamed("variableA");

formulaCell.InvalidateFormula();
ws.RecalculateAllFormulas();

Console.WriteLine($"Expected = 25, Actual = {formulaCell.Value}"); //gets 25
```

`formulaCell.Value` is correctly invalidated and results to `25` at the last line.

No

## Reproducibility
```c#
using ClosedXML.Excel;
using System;

class Program
{
    {
        var wb = new XLWorkbook();
        var ws = wb.Worksheets.Add("Formulas");

        IXLCell formulaCell = ws.Cell(1, 1);
        formulaCell.FormulaA1 = $"=variableA*5";

        Console.WriteLine($"Expected = 0, Actual = {formulaCell.Value}");

        ws.Cell(1, 2).Value = $"5";
        ws.Cell(1, 2).AddToNamed("variableA");

        formulaCell.InvalidateFormula();
        ws.RecalculateAllFormulas();

        Console.WriteLine($"Expected = 25, Actual = {formulaCell.Value}");
    }
}
```

cc: @igitur

Pankraty:
Fixes #1536 

Previously, when checking if cell cached value has to be recalculated, we only analyzed precedent cells. But if a formula uses named ranges and they start to point to another location, the cached value was not properly invalidated. Now, instead of `PrecedentCells`, we obtain a collection consisting of named ranges and cells (under `IXLVersionable` interface) that affect the current cell. 

I also had to remove `ExpressionCache` from `CalcEngine` because it might store a reference to the old range after a named range was modified.

### DIFF
--- a/ClosedXML.Tests/Excel/CalcEngine/FormulaCachingTests.cs
+++ b/ClosedXML.Tests/Excel/CalcEngine/FormulaCachingTests.cs
@@ -11,192 +11,172 @@ namespace ClosedXML.Tests.Excel.CalcEngine
         [Test]
         public void NewWorkbookDoesNotNeedRecalculation()
         {
-            using (var wb = new XLWorkbook())
-            {
-                var sheet = wb.Worksheets.Add("TestSheet");
-                var cell = sheet.Cell(1, 1);
+            using var wb = new XLWorkbook();
+            var sheet = wb.Worksheets.Add("TestSheet");
+            var cell = sheet.Cell(1, 1);
 
-                Assert.AreEqual(0, wb.RecalculationCounter);
-                Assert.IsFalse(cell.NeedsRecalculation);
-            }
+            Assert.AreEqual(0, wb.RecalculationCounter);
+            Assert.IsFalse(cell.NeedsRecalculation);
         }
 
         [Test]
         public void EditCellCausesCounterIncreasing()
         {
-            using (var wb = new XLWorkbook())
-            {
-                var sheet = wb.Worksheets.Add("TestSheet");
-                var cell = sheet.Cell(1, 1);
-                cell.Value = "1234567";
+            using var wb = new XLWorkbook();
+            var sheet = wb.Worksheets.Add("TestSheet");
+            var cell = sheet.Cell(1, 1);
+            cell.Value = "1234567";
 
-                Assert.Greater(wb.RecalculationCounter, 0);
-            }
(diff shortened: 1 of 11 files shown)


igitur:
Rebased.

nghiaphamhn:
I got this error.. when will the bug be resolved.

PR #1542, opened 2020-10-10
Add support for FilterPrivacy when saving document (#1502)

### LINKED ISSUE #1502: Is absolute path stored in the document?

- [ ] Bug
- [ ] Feature
- [x] Question

- [x] Yes
- [ ] No

**Version of ClosedXML**

Tried both 0.95.999.2099 and 0.95.3

If you open a created file in Excel, select File > Info it will warn you that the absolute path is stored in the document.

image

I would prefer that the path is not stored in the document. In Excel it is possible to remove this value so it is probably not necessary.

I tried some previous versions, version 0.94.0 for instance, with the same result.

## Reproducibility

```c#
        {
            {
                var worksheet = workbook.Worksheets.Add("Sheet1");
             
                workbook.SaveAs("HelloWorld.xlsx");
            }
        }

```

brogowski:
Resolves #1502


Pankraty (commented):
I suggest adding one more test where open a reference file with `FilterPrivacy = true` and check the value is read correctly. This way we'll make sure we read and write the property from/to the right place, not just in a random section of a file.

brogowski:
> I suggest adding one more test where open a reference file with `FilterPrivacy = true` and check the value is read correctly. This way we'll make sure we read and write the property from/to the right place, not just in a random section of a file.

I've added the test case - please take a look at the sample xlsx if it's OK as a reference file.

ClosedXML_Tests/ClosedXML_Tests.csproj
```diff
@@ -45,6 +45,10 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.7.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Remove="Resource\TryToLoad\LO\FilterPrivacyEnabledWorkbook.xlsx" />
+  </ItemGroup>
+
```
Pankraty, line 51: This is not needed, `xlsx`s are included as resources by file mask.

brogowski:
Hi @Pankraty, any updates on this one? Should I change anything?

igitur:
I wondered whether this should rather be a new property on `XLWorkbook` instead of a `SaveOptions` parameter. But let's leave it like it is for now. It makes it a very explicit action to enable this.

igitur (approved):

igitur:
@brogowski Thanks for the contribution. Post a picture of yourself in the T-shirt, ok?

PR #1544, opened 2020-10-17
add null check to ProcessRangeRemoved

mts-13:
XLDataValidations.ProcessRangeRemoved () throws a NullReferenceException when adding a row above a cell that has data validation set.
As a solution, I added a null check.

### DIFF
--- a/ClosedXML/Excel/DataValidation/XLDataValidations.cs
+++ b/ClosedXML/Excel/DataValidation/XLDataValidations.cs
@@ -254,7 +254,10 @@ namespace ClosedXML.Excel
             var entry = _dataValidationIndex.GetIntersectedRanges((XLRangeAddress) range.RangeAddress)
                 .SingleOrDefault(e => Equals(e.RangeAddress, range.RangeAddress));
 
-            _dataValidationIndex.Remove(entry.RangeAddress);
+            if (entry != null)
+            {
+                _dataValidationIndex.Remove(entry.RangeAddress);
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


Pankraty:
Even though the fix is clearly correct, the unit test reproducing the issue is needed - to avoid regression in future.

mts-13:
I'm sorry.
Added a test case.

This issue occurs with the following operations on files that use data validation:
Save the file after adding a row above the first row or after adding a column to the left of the first column.

Is this fix correct?

ClosedXML/Excel/DataValidation/XLDataValidations.cs
```diff
@@ -254,7 +254,10 @@ private void ProcessRangeRemoved(IXLRange range)
             var entry = _dataValidationIndex.GetIntersectedRanges((XLRangeAddress) range.RangeAddress)
                 .SingleOrDefault(e => Equals(e.RangeAddress, range.RangeAddress));
 
-            _dataValidationIndex.Remove(entry.RangeAddress);
```
Pankraty, line 257: This `entry.RangeAddress` right after `.SingleOrDefault` looks like a time bomb waiting for NRE.
 I suspected there must be some explanation and sure enough - before 5fe157f18b40a90a5524812fa1f358a2d5ca0b2c the entire `entry` was passed into `Remove` which handled null by itself. Then, during refactoring, the `.RangeAddress` was mechanically added without too much attention and 💥 .

Pankraty (approved):

Pankraty:
@mts-13 thank you for your contribution!

PR #1546, opened 2020-10-19
Add chart export function

steffi3011:
Since the chart export was not possible until now, it has now been implemented and integrated into the existing system. A loading function for charts does not yet exists, but it can of course be added in the future. But in my opinion, it still is a major progress to make chart exports possible at all.

The GenerateWorksheetPartContent(..) function in the XLWorkBook.Save.cs File was extended by the region charts. In it, a chart anchor is set, through which the chart part can be positioned. A plot area is then created in the chart part, in which the respective chart type is then inserted. You can also set many different settings that influence the layout of the charts.

To add a chart to the worksheet, a new object XLChart must be created and filled. This is where the axes are defined, the data they contain, and the formatting that determines how the chart is displayed in the Excel sheet.

The data is either taken from a referenced table or can be inserted manually. For this purpose, the data is presented in individual series, which must be added to the chart. The object ChartSeriesData created for this purpose contains mainly the values of the categories, the categories themselves and the series names. If the data is taken from a referenced table, the direct reference must also be specified.

Charts with any content can be created and added to the Excel worksheet.

### DIFF
--- a/ClosedXML/Excel/Charts/ChartAxis.cs
+++ b/ClosedXML/Excel/Charts/ChartAxis.cs
@@ -1,0 +1,45 @@
+using System;
+using DocumentFormat.OpenXml.Drawing.Charts;
+
+namespace ClosedXML.Excel.Charts
+{
+    public class ChartAxis
+    {
+        private static UInt32 s_uniqueId = 1;
+
+        public String Name { get; set; }
+        public ChartAxisType Type { get; set; }
+        public ChartAxis RelatedAxis { get; set; }
+        public bool Invisible { get; set; }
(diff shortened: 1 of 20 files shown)


igitur:
Thanks, this is a big change, well done. But we're going to have to discuss the design a lot. It would help if you could add unit tests.

jahav:
I have decided to close it without merging. 

ClosedXML is full of half-finished features that kind of work (=they don't) and I don't want another. The process should start with making a correct structures to so chart can be passed through and since chart definition is rather complex (at least 237 elements per chapter 21.2), I am not taking this.

Also, there are no tests, documentation or comments.

PR #1551, opened 2020-11-02
Implement DATEDIF function

MiNGRotten:
Implement DATEDIF function for calculate difference between two dates.
MS specs: https://support.microsoft.com/en-us/office/datedif-function-25dba1a4-2812-480b-84dd-8b32a451b35c

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Functions/DateAndTime.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/DateAndTime.cs
@@ -117,40 +117,27 @@ namespace ClosedXML.Excel.CalcEngine.Functions
 
         private static object Datedif(List<Expression> p)
         {
-            DateTime startDate;
-            DateTime endDate;
-            var unit = (string)p[2];
-
-            if (p[0]._token.Value.GetType() == typeof(string))
-                startDate = DateTime.FromOADate((int)Datevalue(new List<Expression> { p[0] }));
-            else
-                startDate = DateTime.FromOADate(p[0]);
-
-            if (p[1]._token.Value.GetType() == typeof(string))
-                endDate = DateTime.FromOADate((int)Datevalue(new List<Expression> { p[1] }));
-            else
-                endDate = DateTime.FromOADate(p[1]);
+            DateTime startDate = p[0];
+            DateTime endDate = p[1];
+            string unit = p[2];
 
             if (startDate > endDate)
-                throw new NumberException();
+                throw new NumberException("The start date is greater than the end date");
 
-            switch (unit.ToUpper())
+            return (unit.ToUpper()) switch
             {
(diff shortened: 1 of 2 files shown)


igitur:
I refactored the method. There were some edge cases that weren't calculated correctly.

ClosedXML/Excel/CalcEngine/Functions/DateAndTime.cs
```diff
@@ -113,6 +115,31 @@ private static object Date(List<Expression> p)
...
+            DateTime endDate = p[1];
+            string unit = p[2];
+
+            if (startDate > endDate)
+                throw new NumberException("The start date is greater than the end date");
+
+            return (unit.ToUpper()) switch
+            {
+                "Y" => endDate.Year - startDate.Year - (new DateTime(startDate.Year, endDate.Month, endDate.Day) < startDate ? 1 : 0),
+                "M" => Math.Truncate((endDate.Year - startDate.Year) * 12d + endDate.Month - startDate.Month),
```
Pankraty, line 130: Nope.
image

Pankraty (changes requested):
Not all edge cases have been covered, apparently

ClosedXML/Excel/CalcEngine/Functions/DateAndTime.cs
igitur: Good catch. Fixed.

igitur:
I'll squash and rebase after final approval.

ClosedXML/Excel/CalcEngine/Functions/DateAndTime.cs
```diff
@@ -113,6 +115,31 @@ private static object Date(List<Expression> p)
...
+            string unit = p[2];
+
+            if (startDate > endDate)
+                throw new NumberException("The start date is greater than the end date");
+
+            return (unit.ToUpper()) switch
+            {
+                "Y" => endDate.Year - startDate.Year - (new DateTime(startDate.Year, endDate.Month, endDate.Day) < startDate ? 1 : 0),
+                "M" => Math.Truncate((endDate.Year - startDate.Year) * 12d + endDate.Month - startDate.Month - (endDate.Day < startDate.Day ? 1 : 0)),
+                "D" => Math.Truncate(endDate.Subtract(startDate).TotalDays),
```
Pankraty, line 131: Nope. 
image

This should do the trick:
```suggestion
                "D" => Math.Truncate(endDate.Date.Subtract(startDate.Date).TotalDays),
```

Or even without `Math.Truncate` - CLR does not seem to care about daylight saving time and returns whole numbers anyway.

igitur:
Now I understand why Microsoft discourages use of this function ;-)

PR #1552, opened 2020-11-03
Strong-name sign ClosedXML and remove ClosedXML.Signed package

### LINKED ISSUE #1547: Strong-name sign ClosedXML and deprecate ClosedXML.Signed package
- [x] Feature

- [x] Yes

**Version of ClosedXML**

e.g. 0.95.3

At the moment there are two distinct nuget packages for ClosedXML:
- [ClosedXML](https://www.nuget.org/packages/ClosedXML/) (**not** signed with strong-name)
- [ClosedXML.Signed](https://www.nuget.org/packages/ClosedXML.Signed/) (**signed** with strong-name)

Given improvements in .NET Core (re: binding redirects) and recent updates to Microsoft's guidance on strong-naming, I'd suggest we strong-name the `ClosedXML` assembly, and retire the `ClosedXML.Signed` package.

Any objections before I put together a PR?

---

Related links: 

- [Strong Name Signing, dotnet/corefx](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/strong-name-signing.md)
- [Strong naming guidance, docs.microsoft.com](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/strong-naming)
- [Build great libraries using .NET Standard](https://www.youtube.com/watch?v=hLFyycJVo0I&feature=youtu.be&t=1470)

augustoproiete:
... Resolve #1547


igitur:
So after releasing this, do we make the ClosedXML.Signed package reference the standard ClosedXML package? And for how long? I don't want to have to maintain 2 packages indefinitely if there is not reason.

augustoproiete:
@igitur That's also an option.

My initial proposal was to retire the `ClosedXML.Signed` package in the next release, given that it is no longer needed - if anyone wants upgrade to the newest release, it will be strong-name signed already. I still believe this is best path forward.

@Pankraty pointed out that keeping the two packages is low-cost, so continuing to release the two packages together as it is today wouldn't be a big deal in terms of maintenance. It wouldn't resolve [the fork problem](https://github.com/ClosedXML/ClosedXML/issues/1547#issuecomment-733491150) at the package resolution level, as they would continue to be two different packages that need to be resolved by NuGet, but it would resolve the problem at the assembly resolution level as they would have the same version & strong-name.

This PR as of this writing assumes that `ClosedXML.Signed` will no longer exist moving forward, but I'm happy to update the PR to either bring `ClosedXML.Signed` back (and just add the strong-name to `ClosedXML`) or to turn `ClosedXML.Signed` into a meta-package that references `ClosedXML` as you suggested above.

igitur (approved):

PR #1558, opened 2020-11-07
Fixed file is used in another process error

### LINKED ISSUE #1557: Using ClosedXML.Excel.Drawings.XLPictures.Add(string imageFile) sometimes throwing file lock error

- [x] Bug
- [ ] Feature
- [ ] Question

- [x] Yes
- [ ] No

**Version of ClosedXML**

e.g. 0.95.3

With many concurrent requests, we are getting following error in adding image in excel - "The process cannot access the file <FileName> because it is being used by another process"

There should not be error since file is accessed for only reading. This can be easily achieved by opening image file in mode read and share read done by File.OpenRead method.

No

## Reproducibility
NA - Error may come up with large number of requests.

ChiragRupani:
Resolves #1557

### DIFF
--- a/ClosedXML/Excel/Drawings/XLPictures.cs
+++ b/ClosedXML/Excel/Drawings/XLPictures.cs
@@ -74,7 +74,7 @@ namespace ClosedXML.Excel.Drawings
 
         public IXLPicture Add(string imageFile)
         {
-            using (var fs = File.Open(imageFile, FileMode.Open))
+            using (var fs = File.OpenRead(imageFile))
             {
                 var picture = new XLPicture(_worksheet, fs);
                 _pictures.Add(picture);
--- a/ClosedXML_Tests/Excel/ImageHandling/PictureTests.cs
+++ b/ClosedXML_Tests/Excel/ImageHandling/PictureTests.cs
@@ -6,6 +6,7 @@ using System.Drawing;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Threading.Tasks;
 
 namespace ClosedXML_Tests
 {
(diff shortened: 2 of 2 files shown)


Pankraty (commented):
The fix looks good but could you also add a test for this? It does not even need to involve concurrency: just open a file for reading and, while it's open, try to insert the same file as a picture to xlsx.

ClosedXML/Excel/Drawings/XLPictures.cs
```diff
@@ -74,7 +74,7 @@ public IXLPicture Add(Bitmap bitmap, string name)
 
         public IXLPicture Add(string imageFile)
         {
-            using (var fs = File.Open(imageFile, FileMode.Open))
+            using (var fs = File.OpenRead(imageFile))
```
igitur, line 77: I thought `File.Open(imageFile, FileMode.Open)` and `File.OpenRead(imageFile)` were equivalent. What's the diffierence?

Pankraty: image

ChiragRupani: Adding to it, Default [open ](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.open?view=net-5.0)- opens `FileStream` on specified path with read/write access with no sharing. I am seeing failures in our Asp.Net Core app because of that.

ChiragRupani:
> 
> 
> The fix looks good but could you also add a test for this? It does not even need to involve concurrency: just open a file for reading and, while it's open, try to insert the same file as a picture to xlsx.

Opening file may not test correctly since we need to open with Read Mode and Read Sharing. Instead I added parallel invoke for adding same twice in excel, here I am able to reproduce the issue with existing code and passing with the fix.

igitur (approved):

igitur:
Thanks for the fix.

igitur:
After merging this, tests are failing on AppVeyor. Please check it out and try to fix the error. https://ci.appveyor.com/project/ClosedXML/closedxml/builds/36271811/job/lfrcgiapraffisll

ChiragRupani:
I checked it - it happening due to adding new cell in excel workbook is not [thread safe](https://github.com/ClosedXML/ClosedXML/blob/7f077373f6e1aada0a8728126505ed07b7fd6766/ClosedXML/Excel/Cells/XLCellsCollection.cs#L54). The fix is to create and use workbook separately for each thread.

PR: #1562

PR #1562, opened 2020-11-12
Added concurrent check in thread safe way for excel workbook

### LINKED ISSUE #1557: Using ClosedXML.Excel.Drawings.XLPictures.Add(string imageFile) sometimes throwing file lock error

ChiragRupani:
Resolves #1557

### DIFF
--- a/ClosedXML_Tests/Excel/ImageHandling/PictureTests.cs
+++ b/ClosedXML_Tests/Excel/ImageHandling/PictureTests.cs
@@ -94,50 +94,45 @@ namespace ClosedXML_Tests
         [Test]
         public void CanAddPictureConcurrentlyFromFile()
         {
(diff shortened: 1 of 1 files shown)


igitur:
Great thanks. Keep hanging out around here. We can use your talents. ;-)

PR #1565, opened 2020-11-17
WIP: Sorting cells having formulas (#105)

### LINKED ISSUE #105: Sort does not work properly when using Formula's
An excel doc is generated with columns A thru O.  The headers are there and sized right.  
In column A there is just an item number, in column K there is a formula to compare against 2 other cells.

After the iterations of filling in the values from every record in the table, the next part of the code builds the headers, then the sorting done by:  Sort("A"); 

The result is that in my first data row, which is row 4  (the first 3 rows are header or other info) A4 is right, it has the first item, it sorted right.   but K4 ?  totally wrong value.  Click and it says: =(C$2-E33/(F33-E33) and then K5?  is for row 65, not 5.

Is it not possible to do a Sort if the Excel spreadsheet uses Formula's for values?  I thought that the reference C$2 means C2 absolute;  so when the others are in there without '$' then they would be relative indexers and would update?

reference info:
Problem:
Microsoft Excel: What are Absolute and Relative Cell References?

Solution:
Absolute means that the reference in a formula will always look for the value in a specific cell, even if you copy the cell to a different location. The format for an absolute reference is $A$1 (referring to cell A1).

Relative references are in the format A1 and if copied to a different cell the cell reference changes relative to the new location. For example, if the formula specifies A1 and is copied to a new location two rows down the new formula would specify A3.

http://help.isu.edu/index.php?action=faq&catid=38&docid=410

Pankraty:
This is a work in progress

✔️ The issue nr. 1. Sorting algorithm swaps cells leaving their A1-style formulas unchanged.  
✔️ The issue nr. 2. Cells with formulas are compared by `InnerText` which is nonsense. As a consequence, formula `A10+B10` goes before `A2+B2` no matter what are the results of those formulas.
🟥  The issue nr. 3. When two cells are compared, ClosedXML analyzes their `DataType` and if they both have `Text` they are compared as text, and 2 considered larger than 10. 

I think we should implement a comparer that will handle different combinations of data types, blanks, formulas, etc. But right now I don't have enough time. Will try to get back to this sometime later.

Fixes #105

### DIFF
--- a/ClosedXML/Excel/Cells/XLCellsCollection.cs
+++ b/ClosedXML/Excel/Cells/XLCellsCollection.cs
@@ -317,6 +317,9 @@ namespace ClosedXML.Excel
                     if (cell1 == null) cell1 = worksheet.Cell(sp1.Row, sp1.Column);
                     if (cell2 == null) cell2 = worksheet.Cell(sp2.Row, sp2.Column);
 
+                    var formula1 = cell1.HasFormula ? cell1.FormulaR1C1 : (string)null;
+                    var formula2 = cell2.HasFormula ? cell2.FormulaR1C1 : (string)null;
+
                     //if (cell1 != null)
                     //{
                     cell1.Address = new XLAddress(cell1.Worksheet, sp2.Row, sp2.Column, false, false);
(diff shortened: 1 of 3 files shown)


jahav:
Closing in favor of #2413.

This is a draft and the other PR also unifies formula representation to A1 and uses ClosedParser for A1 <-> R1C1 conversion. Later two issues have been sorted out in other PRs to 0.104.0 (mostly AutoFilter) and the first one is solved by the PR.

PR #1569, opened 2020-11-19
Load property "Modified" when loading workbook

### LINKED ISSUE #1568: The Property Properties.Modified is not read when loading a Workbook
- [x] Bug
- [ ] Feature
- [ ] Question

- [ ] Yes
- [X] No=> No need. I checked the code

=> No need. I checked the code

**Version of ClosedXML**

e.g. 0.95.3

When I load a XLWorkbook the property "Modified" is DateTime.MinValue instead of the documents value.

The property "Modified" should have the value of the document.
This can be solved in XLWorkbook_Load in SetProperties by adding the following code:
```
            if (p.Modified != null)
                Properties.Modified = p.Modified.Value;
```

I don't think so.

```c#
[Test]
public void ModifiedRepro ()
{
  // NOTE: arrange
  var now = DateTime.Now;
  using (var workBookForSave = new XLWorkbook ())
  {
    workBookForSave.AddWorksheet ();
    workBookForSave.Properties.Modified = now;
    workBookForSave.SaveAs ("C:\\temp\\bla.xlsx");
  }

  // NOTE: act
  using (var workBookForLoad = new XLWorkbook ("C:\\temp\\bla.xlsx"))
  {
    var modified = workBookForLoad.Properties.Modified;

    // NOTE: assert
    /* this fails with:
      Message: 
        Expected: 2020-11-19 15:35:22.2226398
        But was:  0001-01-01 00:00:00
    */
    Assert.That (modified, Is.EqualTo (now));
  }
}
```

rominator1983:
Closes #1568

### DIFF
--- a/ClosedXML/Excel/XLWorkbook_Load.cs
+++ b/ClosedXML/Excel/XLWorkbook_Load.cs
@@ -3028,6 +3028,8 @@ namespace ClosedXML.Excel
             Properties.Comments = p.Description;
             if (p.Created != null)
                 Properties.Created = p.Created.Value;
+            if (p.Modified != null)
+                Properties.Modified = p.Modified.Value;
             Properties.Keywords = p.Keywords;
             Properties.LastModifiedBy = p.LastModifiedBy;
             Properties.Status = p.ContentStatus;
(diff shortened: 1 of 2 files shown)


Pankraty (commented):
Add a test please. For example, using MemoryStream save a workbook with this property set then load it from that stream and check the property value - this will cover both saving and loading.

rominator1983:
there you go

rominator1983:
I also tested the load/save-behavior of all other properties in the class

Pankraty (approved):

Pankraty:
Thanks for your contribution! Keep going! =)

PR #1576, opened 2020-12-07
Considering options when checking for empty cell

rominator1983:
Change Behavior of IsEmpty(XLCellsUsedOptions) to consider XLCellsUsedOptions.Contents instead of always returning false if a content is set.

### DIFF
--- a/ClosedXML_Tests/Excel/Cells/XLCellTests.cs
+++ b/ClosedXML_Tests/Excel/Cells/XLCellTests.cs
@@ -261,7 +261,7 @@ namespace ClosedXML_Tests
         {
             IXLWorksheet ws = new XLWorkbook().Worksheets.Add("Sheet1");
             IXLCell cell = ws.Cell(1, 1);
-            cell.Comment.AddText ("comment");
+            cell.Comment.AddText("comment");
             bool actual = cell.IsEmpty();
             bool expected = false;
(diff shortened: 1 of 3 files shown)


ClosedXML_Tests/Excel/Cells/XLCellTests.cs
```diff
@@ -256,6 +256,172 @@ public void IsEmpty6()
...
 
+        [Test]
+        public void IsEmpty_Comment()
+        {
+            IXLWorksheet ws = new XLWorkbook().Worksheets.Add("Sheet1");
+            IXLCell cell = ws.Cell(1, 1);
+            cell.Comment.AddText("comment");
+            bool actual = cell.IsEmpty();
+            bool expected = false;
+            Assert.AreEqual(expected, actual);
```
Pankraty, line 267: `Assert.False(cell.IsEmpty());` would be shorter, though it's just a matter of preference.

Pankraty (commented):
I am impressed with the number of tests - good job! But may I ask you to also include the tests making sure `XLCellsUsedOptions.Contents` works as expected when a cell has a) formula b) rich text?

rominator1983:
nevermind. If anyone interested there is a better supprted version of ClosedXml at https://github.com/stesee/ClosedXML

PR #1577, opened 2020-12-07
make test agnostic of culture settings of developer machine

rominator1983:
this test was always red on my machine with Austrian sesttings


ClosedXML_Tests/Excel/Cells/XLCellTests.cs
```diff
@@ -104,13 +104,29 @@ public void Double_NaN_is_a_string()
...
+                Assert.AreEqual (1.5, cell.SetValue (1.5).GetValue<double?> ());
+                Assert.AreEqual (2, cell.SetValue (1.5).GetValue<int?> ());
+                Assert.AreEqual (2.5, cell.SetValue ("2.5").GetValue<double?> ());
+                Assert.Throws<FormatException> (() => cell.SetValue ("text").GetValue<double?> ());
+
+            }
+            finally
+            {
+                Thread.CurrentThread.CurrentCulture = backupCulture;
+            }
```
Pankraty, line 129: Why not doing it like this?
```suggestion
            var cell = new XLWorkbook().AddWorksheet().FirstCell();

            Assert.IsNull(cell.Clear().GetValue<double?>());
            Assert.AreEqual(1.5, cell.SetValue(1.5).GetValue<double?>());
            Assert.AreEqual(2, cell.SetValue(1.5).GetValue<int?>());
            Assert.AreEqual(2.5, cell.SetValue(2.5.ToString()).GetValue<double?>()); // <-- See here
            Assert.Throws<FormatException>(() => cell.SetValue("text").GetValue<double?>());
```

PR #1619, opened 2021-02-21
Implement RelationshipErrorHandler

### LINKED ISSUE #249: Invalid Hyperlink: Malformed URI is embedded as a hyperlink in the document.
I get the Exception Invalid Hyperlink: Malformed URI is embedded as a hyperlink in the document.
at
```
   bei DocumentFormat.OpenXml.Packaging.OpenXmlPackage.Load()
   bei DocumentFormat.OpenXml.Packaging.OpenXmlPackage.OpenCore(String path, Boolean readWriteMode)
   bei DocumentFormat.OpenXml.Packaging.SpreadsheetDocument.Open(String path, Boolean isEditable, OpenSettings openSettings)
   bei ClosedXML.Excel.XLWorkbook.LoadSheets(String fileName) in C:\Git\ClosedXML\ClosedXML\Excel\XLWorkbook_Load.cs:Zeile 44.
   bei ClosedXML.Excel.XLWorkbook..ctor(String file) in C:\Git\ClosedXML\ClosedXML\Excel\XLWorkbook.cs:Zeile 682.
```
I googled a little bit and apparently it is a common problem. You can't open a xlsx with a malformed Uri :(. Here is a Solution to the Problem http://ericwhite.com/blog/handling-invalid-hyperlinks-openxmlpackageexception-in-the-open-xml-sdk/ .
I looked in the sheet1.xml.rels and found the malformed Hyperlink, it has as Target:
" mailto:test.test@test-test.com;%2009.06.2015%20gibt%20ww%20Stahl ".

Maybe you can Implement the Solution from the Link or they already implemented a fitting overload as announced in the blog-post.

igitur:
Fixes #249
Depends on https://github.com/OfficeDev/Open-XML-SDK/issues/887


twsouthwick:
v2.12.3 should be available soon with the fix.

igitur:
Thanks for the speedy fix, as usual, @twsouthwick

igitur:
@Pankraty Work in progress, but if you can start looking at this, it would be appreciated.

Pankraty:
Sorry, this got lost among incoming messages. 
I guess, first 4 commits can be safely merged as a separate PR.

ClosedXML/Excel/XLWorkbook.cs
```diff
@@ -664,6 +664,8 @@ public XLWorkbook()
         internal XLWorkbook(String file, Boolean asTemplate)
             : this(new LoadOptions())
         {
+            if (!asTemplate) throw new NotSupportedException();
```
Pankraty, line 667: Why don't we fallback to ordinary loading? As a user, I would expect these two calls producing identical result:
```
var wb = new XLWorkbook(fileName);
var wb = new XLWorkbook(fileName, asTemplate: false);
```

ClosedXML.Tests/Excel/Loading/LoadingTests.cs
```diff
@@ -32,6 +32,19 @@ public void CanSuccessfullyLoadLOFiles(string file)
             TestHelper.LoadFile(file);
         }
 
+#if !_NET40_
+        // .NET 4.0 does not recognise the malformed mailto URI in the file
+        [TestCaseSource(nameof(MalformedFiles))]
+        public void CanSuccessfullyLoadMalformedFilesWithRemoveMalformedHyperlinksRelationshipErrorHandlerFactory(string file)
```
Pankraty, line 38: Perhaps, we should also have a test checking that opening these files _without_ `RemoveMalformedHyperlinksRelationshipErrorHandlerFactory` fails. If that test starts to fail one day, that will be an indicator for us that the factory is not needed anymore and we can consider removing (deprecating) it.

PR #1620, opened 2021-02-21
Add test to copy .bmp picture to different workbook. 

igitur:
Attempts to address problem in https://dev.to/masanori_msl/net-core-closedxml-copy-excel-sheets-2a5h but can't reproduce yet.


igitur:
Fixed in #1623

PR #1623, opened 2021-02-23
Support .emf and .wmf pictures

### LINKED ISSUE #1621: Some pictures caused System.ArgumentException

- [x] Bug
- [ ] Feature
- [ ] Question

- [ x] Yes
- [ ] No

**Version of ClosedXML**

NuGet 0.95.4

1. I created a file by using template of Microsoft Excel.
2. When I tried to copy pictures by IXLPicture.CopyTo, I got an exception like below.

```
Unhandled exception. System.ArgumentException: The picture format in the stream and the parameter don't match
   at ClosedXML.Excel.Drawings.XLPicture..ctor(IXLWorksheet worksheet, Stream stream, XLPictureFormat format)
   at ClosedXML.Excel.Drawings.XLPictures.Add(Stream stream, XLPictureFormat format)
   at ClosedXML.Excel.Drawings.XLPictures.Add(Stream stream, XLPictureFormat format, String name)
   at ClosedXML.Excel.XLWorksheet.AddPicture(Stream stream, XLPictureFormat format, String name)
   at ClosedXML.Excel.Drawings.XLPicture.CopyTo(XLWorksheet targetSheet)
   at ClosedXML.Excel.Drawings.XLPicture.CopyTo(IXLWorksheet targetSheet)
   at CopySheetSample.Program.Main(String[] args) in C:\Users\example\OneDrive\Documents\workspace\5_0_101\CopySheetSample\Program.cs:line 25
```

* When I saved the problematic picture and added it manually, no exceptions were occurred.
* The problematic picture's format(IXLPicture.Format) was "Bmp"

No exceptions will be occurred or skip copying the problematic pictures

No. I got the same exception when I use v0.95.3.

## Reproducibility
```c#
using System;
using System.IO;
using ClosedXML.Excel;

namespace CopySheetSample
{
    class Program
    {
        {
            byte[] fileData;
            using(XLWorkbook bookFrom = new XLWorkbook(@"C:\Users\example\Desktop\sample.xlsx"))
            {
                using(MemoryStream streamTo = new MemoryStream())
                {
                    using(XLWorkbook bookTo = new XLWorkbook())
                    {
                        foreach(IXLWorksheet sheetFrom in bookFrom.Worksheets)
                        {
                            IXLWorksheet sheetTo = bookTo.Worksheets.Add(sheetFrom.Name);
                            foreach(IXLPicture picture in sheetFrom.Pictures)
                            {
                                Console.WriteLine(picture.Format);
                                picture.CopyTo(sheetTo);
                            }
                        }
                        bookTo.SaveAs(streamTo);
                    }
                    fileData = streamTo.ToArray();
                }
            }
            using(var stream = new FileStream(@"C:\Users\example\Desktop\Sample_" + DateTime.Now.ToString("yyyyMMddHHmmssfff") + ".xlsx", FileMode.Create))
            {
                stream.Write(fileData, 0, fileData.Length);
            }
        }
    }
}
```

- [ x] I attached a sample spreadsheet.  (You can drag files on to this issue)
[sample.xlsx](https://github.com/ClosedXML/ClosedXML/files/6026790/sample.xlsx)

igitur:
This PR adds support for .emf and .wmf pictures. 

It also changes the image loading from `new Bitmap(stream)` to `Image.FromStream(stream)`. The latter detects the .emf raw format correctly.

Fixes #1621


ClosedXML/Excel/Drawings/XLPicture.cs
```diff
@@ -246,6 +244,8 @@ public void Delete()
             Worksheet.Pictures.Delete(this.Name);
         }
 
+        #region IDisposeable
```
Pankraty, line 247: Minor: `IDisposable` (in endregion too)

ClosedXML/Excel/Drawings/PictureEnums.cs
```diff
@@ -8,15 +8,16 @@ public enum XLMarkerPosition
...
+        Unknown,
+        Bmp,
+        Gif,
+        Png,
+        Tiff,
+        Icon,
+        Pcx,
+        Jpeg,
+        Emf,
+        Wmf
```
Pankraty, line 20: Technically, this is a breaking change, though the obstacles in which this may have effect are quite exotic.

ClosedXML/Excel/Drawings/XLPicture.cs
```diff
@@ -436,13 +438,13 @@ private static string GetMimeType(Image i)
...
         {
-            this.OriginalWidth = bitmap.Width;
-            this.OriginalHeight = bitmap.Height;
+            this.OriginalWidth = image.Width;
+            this.OriginalHeight = image.Height;
 
-            this.width = bitmap.Width;
-            this.height = bitmap.Height;
+            this.width = image.Width;
+            this.height = image.Height;
```
Pankraty, line 447: Can we also rename private fields to match the convention? Like, `_width`, `_height`, `_id`, etc.?

Pankraty (approved):
Generally, approved, with one spelling issue and two points for consideration.

ClosedXML/Excel/Drawings/PictureEnums.cs
igitur: I'm happy to accept that risk. I highly doubt it will affect anyone.

ClosedXML/Excel/Drawings/XLPicture.cs
igitur: If I had my choice from the start, I would choose `width_`, etc. It helps with the autocompletion in Visual Studio. I hate having to type the `_`. But yes, I will do this one that you suggest.

igitur: Should
`private const String InvalidNameChars = @":\/?*[]";`
become
`private const String _invalidNameChars = @":\/?*[]";` ?

igitur: Microsoft's convention is `InvalidNameChars`

https://stackoverflow.com/questions/242534/c-sharp-naming-convention-for-constants

Pankraty: I was toooo slooowww :)

ClosedXML/Excel/Drawings/PictureEnums.cs
Pankraty: I do not object either, just pointed that maybe we should put a label on this. May not worth it though.

PR #1624, opened 2021-02-26
Upgrade netcoreapp2.0 to netcoreapp3.1 in Sandbox and Examples

igitur:
@Pankraty Any objections? Seems like a default VS2019 installation doesn't include `netcoreapp2.x` anymore.


Pankraty (approved):
No objections

PR #1626, opened 2021-02-26
Write boolean formula value as lowercase

igitur:
As mentioned in https://github.com/OfficeDev/Open-XML-SDK/pull/890#issuecomment-783743768 boolean values have to be `0`/`1` or `true`/`false` (lower case only).

In .NET `Boolean.ToString()` returns a propercase string, i.e. `True`/`False`.


ClosedXML/Extensions/ObjectExtensions.cs
```diff
@@ -83,52 +55,24 @@ public static string ObjectToInvariantString(this object value)
             if (value == null)
```
Pankraty, line 55: Can't this also be put inside switch?

igitur: Do you mean like this?

```c#
public static string ObjectToInvariantString(this object value)
{
    return value switch
    {
        sbyte v => v.ToString(CultureInfo.InvariantCulture),
        byte v => v.ToString(CultureInfo.InvariantCulture),
        short v => v.ToString(CultureInfo.InvariantCulture),
        ushort v => v.ToString(CultureInfo.InvariantCulture),
        int v => v.ToString(CultureInfo.InvariantCulture),
        uint v => v.ToString(CultureInfo.InvariantCulture),
        long v => v.ToString(CultureInfo.InvariantCulture),
        ulong v => v.ToString(CultureInfo.InvariantCulture),
        float v => v.ToString("G7", CultureInfo.InvariantCulture),// Specify precision explicitly for backward compatibility
        double v => v.ToString("G15", CultureInfo.InvariantCulture),// Specify precision explicitly for backward compatibility
        decimal v => v.ToString(CultureInfo.InvariantCulture),
        TimeSpan ts => ts.ToString("c", CultureInfo.InvariantCulture),
        DateTime d => d.ToString(CultureInfo.InvariantCulture),
        bool b => b.ToString().ToLowerInvariant(),
        _ => value?.ToString() ?? string.Empty,
    };
}
```

igitur: Ohhh. like this...

Didn't know one could mix type and value switches.

```c#
public static string ObjectToInvariantString(this object value)
{
    return value switch
    {
        null => string.Empty,
        sbyte v => v.ToString(CultureInfo.InvariantCulture),
        byte v => v.ToString(CultureInfo.InvariantCulture),
        short v => v.ToString(CultureInfo.InvariantCulture),
        ushort v => v.ToString(CultureInfo.InvariantCulture),
        int v => v.ToString(CultureInfo.InvariantCulture),
        uint v => v.ToString(CultureInfo.InvariantCulture),
        long v => v.ToString(CultureInfo.InvariantCulture),
        ulong v => v.ToString(CultureInfo.InvariantCulture),
        float v => v.ToString("G7", CultureInfo.InvariantCulture),// Specify precision explicitly for backward compatibility
        double v => v.ToString("G15", CultureInfo.InvariantCulture),// Specify precision explicitly for backward compatibility
        decimal v => v.ToString(CultureInfo.InvariantCulture),
        TimeSpan ts => ts.ToString("c", CultureInfo.InvariantCulture),
        DateTime d => d.ToString(CultureInfo.InvariantCulture),
        bool b => b.ToString().ToLowerInvariant(),
        _ => value.ToString(),
    };
}
```

Pankraty: Exactly

Pankraty (approved):

Issue #1545, opened 2020-10-18
Thousands of System.MissingMethodException: Method not found


yarecky1:
**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

**Version of ClosedXML**

0.95.3

**What is the current behavior?**

I have thousands of: 
```
System.MissingMethodException: Method not found: 'System.String ExcelNumberFormat.NumberFormat.Format(System.Object, System.Globalization.CultureInfo)'.
   at ClosedXML.Extensions.FormatExtensions.ToExcelFormat(Object o, String format)
   at ClosedXML.Excel.XLCell.GetFormattedString()
```
...so many that they don't fit in entire ImmediateWindow in Visual Studio.

These thousands of exceptions slow down dramatically entire process of building excel file.
I see also you swallow 2 exceptions here.
```
public string GetFormattedString()
{
    var format = GetFormat();
    try
    {
        return Value.ToExcelFormat(format);
    }
    catch { }

    try
    {
        return CachedValue.ToExcelFormat(format);
    }
    catch { }

    return _cellValue;
}
``` 

**What is the expected behavior or new feature?**

To fix the bug?

**Is this a regression from the previous version?**

Yes.

## Reproducibility

No code. Please just fix it.

Pankraty:
Let's start by providing a reproducible example, shall we?

yarecky1:
In worksheet I have several cells with template strings like `{id1}` or just `id2`.
```c#
internal void Test()
{
    var path = Path.GetFullPath(AppDomain.CurrentDomain.BaseDirectory + @"Resources\Test.xltx");

    using (var source = new XLWorkbook(path, XLEventTracking.Disabled))
    {
        source.Worksheet("TEST").Search("{id1}").Value = "TEST1";
        source.Worksheet("TEST").Search("id2").Value = "TEST2";
    }
}
```
One code line `source...` generates 6 `MissingMethodException`.

[Test.xlsx](https://github.com/ClosedXML/ClosedXML/files/5397285/Test.xlsx)
I use xltx instead of xlsx - to be able to attach it here, I changed the extension.

igitur:
> No code. Please just fix it.

Wow.

yarecky1:
>> No code. Please just fix it.

> Wow.

Swallowing exception...

Netclick19:
Hi,
[...]

PR #1632, opened 2021-03-03
Fixed the error that caused the massive amounts of System.MissingMet…

charlie-work:
Fixed the error that caused the massive amounts of System.MissingMethodException: Method not found messages.  This also speeds up the process considerably.

This is in reference to issue #1545.  The problem mentioned occurs because ExcelNumberFormat does not support an empty or null string for formatting.  By not passing the empty string as a format, this is avoided.


igitur:
Thanks for the contribution. A few things: please add a unit test to ensure that we don't ever undo this fix. Also ensure that all current unit tests pass. I see a few are now failing.

igitur:
Just curious... are you not running the unit tests on your own machine before pushing the changes?

charlie-work:
it doesn't run with what i pulled from the repo. also, no idea how to do these so...  i've never written unit tests or anything. no idea how to do that. :/

charlie-work:
i can't get this to pass with your other tests. it works fine on my machine. so i'll just abandon your nuget package and keep my changes local.

igitur:
I can try to help, but so far nobody has been able to provide a simple reproducible test case to show me exactly what the issue is. If you open a new issue and fill in the issue template exactly as requested, then I can inspect.

charlie-work:
ok. i'll open a bug report

Murphybro2:
Was this fix ever committed? I can't seem to find it in use anywhere..

charlie-work:
not my fix.  i don't know how to write unit tests to satisfy their needs.  i just make the changes to my local package and uploaded to my private nuget server.

PR #1639, opened 2021-03-08
Do not attempt to create directory with empty string

igitur:


Pankraty (commented):
Can we add a test for this?

igitur:
> Can we add a test for this?

I thought about that, but would the test run reliable on the CI server? I didn't want to assume write access to the default dir. 

But ok, let's see.

Pankraty (approved):

PR #1641, opened 2021-03-15
Avoid confusion between named ranges and functions with same name

igitur:
This solves 2 issues:
- Boolean parameters were parsed as identifiers and only parsed to Boolean values during evaluation. This now changes it to parse Boolean literal value tokens when the expression is parsed.
- Peeks ahead for `(` token to distinguish between named ranges and functions, e.g. `=RAND * 10` vs `=RAND() * 10`.


Pankraty (approved):
Thanks for the clear commit history - reviewing it all at once would be much harder.

igitur:
> Thanks for the clear commit history - reviewing it all at once would be much harder.

Yes, especially with the field renaming!

I keep thinking we've ironed out all the bugs in the calculation engine, and then more just keeps popping up.

I found this one while reviewing your #1539 PR.

PR #1647, opened 2021-04-05
Improve RowsUsed and ColumnsUsed performance

igitur:
Checks `Worksheet.Internals.CellsCollection` for rows/columns used before  iterating.

### DIFF
--- a/ClosedXML/Excel/Ranges/XLRange.cs
+++ b/ClosedXML/Excel/Ranges/XLRange.cs
@@ -1,4 +1,3 @@
-using ClosedXML.Extensions;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -650,7 +649,8 @@ namespace ClosedXML.Excel
         internal XLRangeRows RowsUsed(XLCellsUsedOptions options, Func<IXLRangeRow, Boolean> predicate = null)
         {
             XLRangeRows rows = new XLRangeRows();
-            Int32 rowCount = RowCount();
+            Int32 rowCount = RowCount(options);
+
             for (Int32 ro = 1; ro <= rowCount; ro++)
             {
                 var row = Row(ro);
@@ -686,7 +686,8 @@ namespace ClosedXML.Excel
         internal virtual XLRangeColumns ColumnsUsed(XLCellsUsedOptions options, Func<IXLRangeColumn, Boolean> predicate = null)
         {
             XLRangeColumns columns = new XLRangeColumns();
-            Int32 columnCount = ColumnCount();
+            Int32 columnCount = ColumnCount(options);
+
             for (Int32 co = 1; co <= columnCount; co++)
             {
                 var column = Column(co);
--- a/ClosedXML.Tests/Excel/Columns/ColumnTests.cs
+++ b/ClosedXML.Tests/Excel/Columns/ColumnTests.cs
@@ -24,6 +24,16 @@ namespace ClosedXML.Tests.Excel
         }
 
         [Test]
+        public void ColumnsUsedIsFast()
+        {
+            using var wb = new XLWorkbook();
+            var ws = wb.AddWorksheet();
+            ws.FirstCell().SetValue("Hello world!");
+            var columnsUsed = ws.Row(1).AsRange().ColumnsUsed();
+            Assert.AreEqual(1, columnsUsed.Count());
+        }
+
+        [Test]
         public void CopyColumn()
         {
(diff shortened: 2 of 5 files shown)


ClosedXML/Excel/Ranges/XLRange.cs
```diff
@@ -651,6 +650,12 @@ internal XLRangeRows RowsUsed(XLCellsUsedOptions options, Func<IXLRangeRow, Bool
         {
             XLRangeRows rows = new XLRangeRows();
             Int32 rowCount = RowCount();
+
+            if (Worksheet.Internals.RowsCollection.Any())
+                rowCount = Math.Min(rowCount, Math.Max(Worksheet.Internals.RowsCollection.MaxRowUsed, Worksheet.Internals.CellsCollection.MaxRowUsed));
+            else
+                rowCount = Math.Min(rowCount, Worksheet.Internals.CellsCollection.MaxRowUsed);
```
Pankraty, line 657: The effect is drastic, and this works correctly for like 99.99% of real-life cases, but there may be some edge cases. E.g.
```
        [Test]
        public void RowsUsedWithDataValidation()
        {
            var ws = wb.AddWorksheet();
            ws.FirstCell().SetValue("Hello world!");
            ws.Range("A1:A100").CreateDataValidation().WholeNumber.EqualTo(1);
            var rowsUsed = ws.Column(1).AsRange().RowsUsed(XLCellsUsedOptions.DataValidation); // Or XLCellsUsedOptions.All
            Assert.AreEqual(100, rowsUsed.Count());
        }
```
Same for conditional formats, merged ranges, and sparklines. We should probably use a "fast track" when none of these options is specified, and process them separately when they are present.

As for the `predicate`, it complicates the things even more, because it is always possible to provide such that will produce the wrong result. For example
```
        [Test]
        public void RowsUsedWithPredicate()
        {
            var ws = wb.AddWorksheet();
            ws.FirstCell().SetValue("Hello world!");
            ws.Range("B1:B100").Value = "Not empty";

            var rowsUsed = ws.Column(1).AsRange().RowsUsed(r => r.Cell(1).CellRight().Value == "Not empty");
            Assert.AreEqual(100, rowsUsed.Count());
        }
```
Maybe we should consider deprecation of versions with predicates at all - I don't think they are widely used, else we'd have lots of bug reports regarding them.

igitur:
```c#
[Test]
public void RowsUsedWithPredicate()
{
    var ws = wb.AddWorksheet();
    ws.FirstCell().SetValue("Hello world!");
    ws.Range("B1:B100").Value = "Not empty";

    var rowsUsed = ws.Column(1).AsRange().RowsUsed(r => r.Cell(1).CellRight().Value == "Not empty");
    Assert.AreEqual(100, rowsUsed.Count());
}
```

This is an very odd use-case, but by design, `rowsUsed.Count() == 1`. A single column's rows don't span the entire row, but only the extent of columns covered.  I wouldn't want to remove the generic `IXLRange.RowsUsed(predicate)`. I actually use that sometimes. But maybe we can have `Column(1).AsRange()` return a new class `IXLColumnRange` where `.RowsUsed(predidate)` is overloaded with a `NotImplementedException` or something like that inside? But the purpose would then just be to remove an accidental misunderstanding and misuse of `IXLColumn.AsRange().RowsUsed()`. This can maybe just be fixed with better doctext.

igitur:
```c#
[Test]
public void RowsUsedWithPredicate()
{
    var ws = wb.AddWorksheet();
    ws.FirstCell().SetValue("Hello world!");
    ws.Range("B1:B100").Value = "Not empty";

    var rowsUsed = ws.Column(1).AsRange().RowsUsed(r => r.Cell(1).CellRight().Value == "Not empty");
    Assert.AreEqual(100, rowsUsed.Count());
}
```

Added this test and reworked code to ensure it works.

igitur:
```c#
[Test]
public void RowsUsedWithDataValidation()
{
    var ws = wb.AddWorksheet();
    ws.FirstCell().SetValue("Hello world!");
    ws.Range("A1:A100").CreateDataValidation().WholeNumber.EqualTo(1);
    var rowsUsed = ws.Column(1).AsRange().RowsUsed(XLCellsUsedOptions.DataValidation); // Or XLCellsUsedOptions.All
    Assert.AreEqual(100, rowsUsed.Count());
}
```

Added this test and reworked code to ensure it works.

Pankraty (approved):

PR #1658, opened 2021-04-20
Fixes related to shifting ranges and sorting

### LINKED ISSUE #880: Addresses in named ranges do not become invalid when the range itself is deleted

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.93

Say, we have a named range referring to the range `A1:B2`. If rows from 1 to 5 are deleted the named range starts to refer to `A1:B1`. If, instead, first two columns are deleted the named range becomes `A1:A2`.

The named range should become invalid: e.g. `Sheet1!#REF!`.

No

**Notes**

This issue should be addressed together with #686.

Also, I noticed that when multiple rows are deleted (e.g. `ws.Rows(1, 100).Delete();`) the formula shifting is performed for every single row. This may be improved by doing a single shift for the specified number of rows. Same is true for columns.

## Reproducibility
```c#
[Test]
public void NamedRangeBecomesInvalidOnRangeDeleting()
{
    {
        var ws = wb.Worksheets.Add("Sheet 1");
        ws.Range("A1:B2").AddToNamed("Simple", XLScope.Workbook);
        wb.NamedRanges.Add("Compound", new XLRanges
        {
            ws.Range("C1:D2"),
            ws.Range("A10:D15")
        });

        ws.Rows(1,5).Delete();

        Assert.AreEqual(2, wb.NamedRanges.Count());
        Assert.AreEqual(0, wb.NamedRanges.ValidNamedRanges().Count());
        Assert.AreEqual("'Sheet 1'!#REF!", wb.NamedRanges.ElementAt(0).RefersTo);
        Assert.AreEqual("'Sheet 1'!#REF!,'Sheet 1'!A5:D10", wb.NamedRanges.ElementAt(0).RefersTo);
    }
}
```

clenzmeier:
Fix shifting formulas (during insert/delete/sort) [this fixes #880 and is at least a partial fix for other related issues; however, data validation shifting is still broken]
Fix shifting during sort (so that cells have their formulas shifted)
Fix sorting of differently typed values
Some performance improvements (structs can be slow!)


Pankraty:
@clenzmeier thanks for the contribution and sorry it took us so long to start looking at this!

This is a pretty large PR aimed to solve several independent issues. May I ask you to split this in at least two parts: one for performance optimizations, the other fixing the shifting/sorting issues? Note that @igitur is currently re-working sort and your changes may be conflicting.

jahav:
I took a look at this and it's a kitchen sink plus either the original code has been deleted in the meantime or it doesn't match how I want it done. There are ~9 changes:

* `XLCell.ShiftFormula` - the core of the PR. This method is slated for deletion. Any fomula shift should be done through converion to R1C1 at original location and turn the R1C1 at the target location back to A1. It should also be done in `FormulaSlice` (e.g. #2413 for sorting).
* It adds a comparison between double and empty text in `Expression.cs` (e.g. 1 > ""). That is from time when formulas didn't use types. The scalar values in formulas should now be compared (through `CalculationVisitor` for `BinaryNode`) by `AnyValue`/`ScalarValue`. They should be correct for all cases and deal with automatic coertion.
* `XLCell.FormulaA1` - something with array formulas, they are done through `IXLRangeBase.FormaulArrayA1` and formulas surrounded by `{}` are invalid.
* `XLCell.SetHyperlink` - Adds public API, which is not something I want to do.
* `XLRangeRow.CompareTo` - some changes for comparisons, but that is a dead function.
* Changes to caching of style keys hash key. Style keys are now immutable, so no need for cache hashing or deal with changes.
* A lot of replacements of RangeAddress access with stored intermediate props in various places and few other minor things.

Therefore I am closing this.

PR #1672, opened 2021-05-27
Add new methods for adding pictures that don't require System.Drawing.Bitmap to work

SuNNjek:
Currently adding a picture to a worksheet requires System.Drawing.Bitmap to work. However, the only use for it is reading out format and size of the image.
This PR adds methods that take the format and size as parameters to be passed in. This way, pictures can be added on platforms that don't support System.Drawing.Bitmap, such as UWP, Android and iOS.

### DIFF
--- a/ClosedXML.Tests/Excel/ImageHandling/PictureTests.cs
+++ b/ClosedXML.Tests/Excel/ImageHandling/PictureTests.cs
@@ -31,8 +31,6 @@ namespace ClosedXML.Tests
                     Assert.AreEqual(XLPictureFormat.Jpeg, picture.Format);
                     Assert.AreEqual(200, picture.Width);
                     Assert.AreEqual(200, picture.Height);
-                    Assert.AreEqual(72, picture.DpiX);
-                    Assert.AreEqual(72, picture.DpiY);
                 }
             }
         }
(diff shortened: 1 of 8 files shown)


Pankraty:
I've made some changes to reduce code duplication across constructors, but unable to push to your branch. Please, either cherry-pick changes from https://github.com/Pankraty/ClosedXML/commit/35e05e256d25cccf5b2577a7d203f1147778ec1d or set a checkbox "Allow maintainers to change PR"

SuNNjek:
The checkbox should be checked already, at least for me

Pankraty:
My bad, used the wrong SSH key.

jahav:
System.Drawing.Common dependency has been removed in #1805 and if there is a need to be able to load additional formats, it should be implemented through a different https://github.com/ClosedXML/ClosedXML/wiki/Graphic-Engine instead of additional public API. 

Note that the use of image formats other than PNG/JPG is discrouraged:
> A producer that wants interoperability should use one of the following standard formats:
> * image/png ISO/IEC 15948:2003, http://www.libpng.org/pub/png/spec/
> * image/jpeg, http://www.w3.org/Graphics/JPEG
>                                      -- ECMA-376, Part1, §15.2.14

PR #1695, opened 2021-07-16
Fix warnings

leotsarev:
I'm trying to make ClosedXML multitarget net5, and there is some new warnings (as errors).

I think that fixes are useful (at least, not harmful), on their own


leotsarev:
@igitur could you please push this? it's baseline for some other work

leotsarev:
Still useful

jahav (approved):

jahav:
One eternity later... Merged.

Issue #1694, opened 2021-07-16
System.Drawing.Common is deprecated on non-Windows platforms


leotsarev:
## Read and complete the full issue template

Do not randomly delete sections. They are here for a reason.

**Do you want to request a *feature* or report a *bug*?**
- [ ] Bug
- [ ] Feature
- [x] Question

**Did you test against the latest CI build?**
- [ ] Yes
- [ ] No
- [x] n/a
If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**
0.95.1

**What is the current behavior?**
https://github.com/dotnet/designs/blob/a15a8c7a324c482c5e38b73e46c2afe56f6bd504/accepted/2021/system-drawing-win-only/system-drawing-win-only.md
System.Drawing is officially deprecated everywhere but Windows.

**What is the expected behavior or new feature?**
Either:
Port to another crossplatform drawing library
Or:
Mark corresponding methods (like AdjustContents) with SupportedOSPlatformAttribute(Windows)  which requires adding net5+ TFM

**Is this a regression from the previous version?**
no

## Reproducibility
n/a

igitur:
Thanks. Good suggestions. Would you mind contributing to the project by logging a PR for this?

igitur:
Oh, sorry, does #1696 address this?

leotsarev:
@igitur  — not yet. If it's ok, I'll contribute PR on top of #1696 
Meanwhile, could you please review and push #1695  & #1696 ?

steffi3011:
Hello, has anything been done here yet? And if not, can you tell when adjustments are planned?

leotsarev:
@igitur what's the plan going forward for #1695 & #1696 ?

fjmorel:
In the meantime, at least users of this library would get warnings when using methods which depend on it.

lazaromenezes:
FYI: .Net 6 removed the System.Drawing.Common support for non-Windows systems: https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/system-drawing-common-windows-only

Trying the runtime config flag hasn't been working for us so far. 

The suggested solution of installing **libgdlplus** works fine for our 3.1 environment tho.

dgmoex:
[...]

PR #1696, opened 2021-07-16
Net5 multitarget

leotsarev:
Based on #1695.
Idea is to multi-target to net5 to use `SupportedOSPlatformAttribute` which is useful for #1694


leotsarev:
superseded by https://github.com/ClosedXML/ClosedXML/pull/1866

Issue #954, opened 2018-07-18
Ensure all ADO.NET data types are supported


igitur:
As discussed in https://github.com/ClosedXML/ClosedXML/issues/953#issuecomment-405855395

Full list documented in 
- MS SQL: https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql-server-data-type-mappings
- OLEDB: https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/ole-db-data-type-mappings
- ODBC: https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/odbc-data-type-mappings

igitur:
The list of distinct .NET framework types according to https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql-server-data-type-mappings below. I also add my suggested types to convert to in ClosedXML
- Boolean → Boolean
- Byte → Double
- ~Byte[] → String (?)~
- ~Char[] → String~
- DateTime → DateTime
- DateTimeOffset → DateTime
- Decimal → Double
- Double → Double
- Guid → String
- Int16 → Double
- Int32 → Double
- Int64 → Double
- ~Object → string~
- Single → Double
- String → String
- TimeSpan → TimeSpan
- ~Xml → String~

Struck out types that I think we should ignore. Arrays are ambiguous because the user might want the values to go into separate cells.

igitur:
Fixed in #1705

PR #1705, opened 2021-08-15
Support inserting DateTimeOffset into table

CoenraadS:
Continuation of https://github.com/ClosedXML/ClosedXML/pull/1305

>Ah ok, that makes more sense. For InsertTable we want to support all data types that are natively supported by SQL. Is DateTimeOffset supported by SQL? And if so, please add unit tests for InsertTable, which is the root issue here.

`DateTimeOffset` is supported by SQL: https://www.w3schools.com/sql/sql_datatypes.asp
Relates to: https://github.com/ClosedXML/ClosedXML/issues/954

### DIFF
--- a/ClosedXML.Tests/Excel/Cells/XLCellTests.cs
+++ b/ClosedXML.Tests/Excel/Cells/XLCellTests.cs
@@ -1176,5 +1176,13 @@ namespace ClosedXML.Tests
 
             Assert.Throws<FormatException>(() => c.ToString("dummy"));
         }
+
+        [Test]
+        public void ConvertOtherSupportedTypes()
+        {
+            Assert.AreEqual("", XLCell.ConvertOtherSupportedTypes(DBNull.Value));
+            Assert.AreEqual("748bdf0c-3e7d-415e-967d-a875a27634ed", XLCell.ConvertOtherSupportedTypes(new Guid("748BDF0C-3E7D-415E-967D-A875A27634ED")));
+            Assert.AreEqual(DateTime.Today, XLCell.ConvertOtherSupportedTypes(DateTime.Today));
+        }
     }
 }
--- a/ClosedXML.Tests/Excel/Cells/XLCellTests.cs
+++ b/ClosedXML.Tests/Excel/Cells/XLCellTests.cs
@@ -569,6 +569,21 @@ namespace ClosedXML.Tests
             Assert.IsFalse(ws.Cell("A4").TryGetValue(out double? _));
         }
 
+        [TestCase(2019, 11, 5, 11, 30, 5, 0, ExpectedResult = "2019-11-05 11:30:05.000")]
+        [TestCase(2019, 11, 5, 11, 30, 5, 2, ExpectedResult = "2019-11-05 11:30:05.000")]
+        [TestCase(2019, 11, 5, 11, 30, 5, -10, ExpectedResult = "2019-11-05 11:30:05.000")]
+        public string ValueSetDateTimeOffset(int year, int month, int days, int hours, int minutes, int seconds, int offsetInHours)
+        {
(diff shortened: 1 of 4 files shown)


gleneccles:
Hi guys. Just wondering when this is going to be released. I am very close to being able to replace EPPlus and this is one of the blockers.

lmazaprisma:
Hola chicos, nosotros tambien estamos a la espera de este fix.
When do you think it will be ready?
Thanks.
Best regards.

igitur:
@CoenraadS If the `DateTimeOffset` value is in a different timezone to the local timezone, should the resulting `Datetime` value in the cell be the local time or the time in the other timezone?

igitur:
If I have this an MS SQL Server table:
image
and I use PowerQuery to dump it in Excel, then I see
image
which means the timezone offset is just discarded.

So there is a motivation to do it like that.

However, if I were to choose, I would convert all dates to either UTC or local timezone.

igitur:
@Pankraty @jahav Your opinions?

leonardojaviermaza:
igitur, Is correct, the Excel doesn't work with time zone. Although, the problem is ClosedXML doesn't work with data type DateTimeZone, and it throw an exception. It 's desirable that not throw an exception and map this type.

igitur:
> igitur, Is correct, the Excel doesn't work with time zone. Although, the problem is ClosedXML doesn't work with data type DateTimeZone, and it throw an exception. It 's desirable that not throw an exception and map this type.

Yes, so what I'm asking is HOW do you map it to `DateTime`?

gleneccles:
For my use case it doesn't matter. Someone is going to like it one way and someone else will like it another way. A setting to configure this would solve the problem, but I wouldn't hold up the release for it

leonardojaviermaza:
> Yes, so what I'm asking is HOW do you map it to `DateTime`?

Yes, when you use DateTimeOffset in any TimeZone, if you map to DateTime, the last one show de time in correct form.
For example
1) var dateTimeOffset = new DateTimeOffset(new DateTime(2022, 06, 29, 0, 0, 0, DateTimeKind.Utc));
2) dateTimeOffset.ToOffset(new TimeSpan(-3, 0, 0));

Map dateTimeOffset.DateTime
If you map on line 1 show 2022/06/29
If you map on line 2 show 2021/06/28 21:00

I hope be clear

Pankraty:
I would go with a simple `dateTimeOffest.DateTime` which effectively discards the timezone - I think, this would be the least surprising behavior. And when a user needs any more comprehensive conversion, they can (should) perform upfront conversion whichever way they like.

ClosedXML.Tests/Excel/Cells/XLCellTests.cs
```diff
@@ -569,6 +569,20 @@ public void TryGetValue_Nullable()
             Assert.IsFalse(ws.Cell("A4").TryGetValue(out double? _));
         }
 
+        [Test]
+        public void ValueSetDateTimeOffset()
+        {
+            IXLWorksheet ws = new XLWorkbook().Worksheets.Add("Sheet1");
+            IXLCell cell = ws.Cell(1, 1);
+            var expected = new DateTimeOffset(2019, 11, 5, 11, 30, 5, new TimeSpan(0));
```
Pankraty: Better rename the variable to `input` and make `expected` a string variable with plain string, without using the `format`. Also, we should probably use non-zero offset here.

igitur: Refactored this.

igitur:
I pushed some commits to include a larger refactoring of setting values internally. Clearer separation between values that we support 'natively' and broader support to accommodate types from a dataset.

igitur:
> I would go with a simple `dateTimeOffest.DateTime` which effectively discards the timezone 

Ok, I'll go with this.

Pankraty (approved):

PR #1706, opened 2021-08-15
Make tests culture invarient

### LINKED ISSUE #1709: Unit Tests with Decimal Numbers fail with different CultureInfo

- [x] Bug
- [ ] Feature
- [ ] Question

- [x] Yes
- [ ] No

**Version of ClosedXML**

0.95.4

Some Unit Tests, which supply and parse doubles, fail with CultureInfo like german, because a value of 123.4 is parsed as 2 parameters (german exchanges behaviour of . and , but C# double.ToString() ignores CultureInfo and always prints decimal separators as .)

Have Unit tests insert numbers with the correct decimal separator for the current CultureInfo, or even introduce UnitTests for different CultureInfos in general.

I doubt it.

## Reproducibility

No custom code needed. Just run the Unit Tests on a system set to german language, of force the CultureInfo to german in some other manner.

- [ ] I attached a sample spreadsheet.  (You can drag files on to this issue) <- The issue is in code only.

CoenraadS:
I had same problem as https://github.com/ClosedXML/ClosedXML/pull/1577 for my locale. This is a simpler fix.

Also closes #1709

### DIFF
--- a/ClosedXML.Tests/Excel/CalcEngine/MathTrigTests.cs
+++ b/ClosedXML.Tests/Excel/CalcEngine/MathTrigTests.cs
@@ -9,6 +9,7 @@ using System.Linq;
 namespace ClosedXML.Tests.Excel.CalcEngine
 {
     [TestFixture]
+    [SetCulture("")]
     public class MathTrigTests
     {
         private readonly double tolerance = 1e-10;
--- a/ClosedXML.Tests/Excel/Cells/XLCellTests.cs
+++ b/ClosedXML.Tests/Excel/Cells/XLCellTests.cs
@@ -11,6 +11,7 @@ using System.Threading;
 namespace ClosedXML.Tests
 {
     [TestFixture]
+    [SetCulture("")]
     public class XLCellTests
     {
         [Test]


CoenraadS:
However this disguises the problem that consumers of this ClosedXml must set thread culture to invariant before using it.

Shad0wlife:
Actually found that fix in my testing as well, but couldn't open a PR. Thanks for that!

ClosedXML.Tests/Excel/CalcEngine/MathTrigTests.cs
```diff
@@ -9,6 +9,7 @@
 namespace ClosedXML.Tests.Excel.CalcEngine
 {
     [TestFixture]
+    [SetCulture("")]
```
jahav, line 12: This is problematic, because what it essentially does is mask potential problems when culture of user is different than invariant culture. It's desirable that tests are run in many cultures and still work correctly, after all ClosedXml is also run with different current cultures.

The problem with failing test is that interpolated string formats MROUND formula with a current culture and passes it as an input to CalcEngine. By formatting it in current culture, it sends invalid data (valid formula has to satisfy the grammar).

Correct fix would be to create a correct input for CalcEngine, something like

```
string.Format(CultureInfo.InvariantCulture, "MROUND({0}, {1})", number, multiple)
```

ClosedXML.Tests/Excel/Cells/XLCellTests.cs
```diff
@@ -11,6 +11,7 @@
 namespace ClosedXML.Tests
 {
     [TestFixture]
+    [SetCulture("")]
```
jahav, line 14: Dtto, by setting a culture we are potentially masking an error in implementation when current culture differs from invariant one. Although in this case it is a problem that the test takes a string and tries to convert it to number and that conversion should be done through local culture (not a great idea, but part of the spec).

> ECMA-36718.17.2.6 
> An implementation is permitted to provide an implicit conversion from string-constant to number. However, the
rules by which such conversions take place are implementation-defined. [Example: An implementation might
choose to accept "123"+10 by converting the string "123" to the number 123. Such conversions might be locale-
specific in that a string-constant such as "10,56" might be converted to 10.56 in some locales, but not in others,
depending on the radix point character. end example]

Most of tests use CultureInfo.CurrentCulture, so I guess its better to update TryGetValue_Nullable
```
            ws.Cell("A3").SetValue(2.5.ToString(CultureInfo.CurrentCulture));
```

jahav (changes requested):
Sorry about long delay of PR.

PR #1725, opened 2021-10-06
Fix shape disappear if there was no image

### LINKED ISSUE #1252: Does save or save as not support saving shape?
Version of ClosedXML

0.94.2

Does save or save as not support saving shape?
There is a shape in my aaa. xlsx file and a text in the shape.
Open the file and save it (save or save as). There are no shapes in the saved file.
Open it with 7zip and find that drawings are not available under xl. Does the current version not support drawings yet?

Code to reproduce problem:

        {
            //Console.WriteLine("Hello World!");

            try
            {
                //using (var workbook = XLWorkbook.OpenFromTemplate(@"D:\aaa.xlsx"))
                using (var workbook = new XLWorkbook(@"D:\aaa.xlsx"))
                {
                    //var worksheet = workbook.Worksheets.Add("DA01");

                    var worksheet = workbook.Worksheet("aaa");
                    worksheet.Cell("A6").Value = "Test";
                    workbook.Save();
                    //workbook.SaveAs(@"D:\aaa_saveas.xlsx");

                }

                Console.WriteLine("finished");
            }
            catch(Exception ex) {
                Console.WriteLine(ex.Message);
            }

            Console.ReadKey();
        }

adm90:
Thank you for always great tools.
We have tried to fix the bug, so please check it.

Fixes #1252 

image

Fix shape disappear if there was no image.
It also fixes an issue where image reference links would appear broken when some images were deleted.

### This file is No-Image, shape only.
- Expect: Keep shapes.
- Actual: All shapes disappear.
image

### Delete only one image when there are multiple images.
(Added Test case CanDeletePictureOnlyOne)
- Expect: One image is delete.
- Actual: Image link is not deleted.
image

- After correction:
image

### DIFF
--- a/ClosedXML.Tests/Excel/ImageHandling/PictureTests.cs
+++ b/ClosedXML.Tests/Excel/ImageHandling/PictureTests.cs
@@ -289,6 +289,31 @@ namespace ClosedXML.Tests
         }
 
         [Test]
+        public void CanDeletePictureOnlyOne()
+        {
+            using (var ms = new MemoryStream())
+            {
+                int originalCount;
+
+                using (var stream = TestHelper.GetStreamFromResource(TestHelper.GetResourcePath(@"Examples\ImageHandling\ImageAnchors.xlsx")))
+                using (var wb = new XLWorkbook(stream))
+                {
+                    var ws = wb.Worksheets.First();
+                    originalCount = ws.Pictures.Count;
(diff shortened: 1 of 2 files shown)


rominator1983:
I pulled your commit to here https://github.com/stesee/ClosedXML/pull/20

Issue #1715, opened 2021-08-27
Add support for Median() statistics function.


silvertop:
## Read and complete the full issue template

Do not randomly delete sections. They are here for a reason.

**Do you want to request a *feature* or report a *bug*?**
- [ ] Bug
- [x ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.95.4

**What is the current behavior?**

Exception is generated stating that "Median() is not supported".

**What is the expected behavior or new feature?**

Median() works.

**Is this a regression from the previous version?**

No

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

ws.Cell(row, col).FormulaA1 = $"=MEDIAN(C2:C30)";

**Code to reproduce problem:**
```c#
public void Main()
{
    // Code standards:
    // - Fully runnable. I should be able to copy and paste this code into a 
    //   console application and run it without having to edit it much.
    // - Declare all your variables (this follows from the previous point)
    // - The code should be a minimal code sample to illustrate issue. The code 
    //   samples on the wiki are good examples of the terseness that I want. Don't
    //   post your full application.

     XLWorkbook wb = new XLWorkbook("c:\temp\test.xlsx"))
     IXLWorksheet ws = wb.Worksheets.Add(LimitWorksheetNameLength("Summary"));
     int row = 32;
     int col = 3;
     ws.Cell(row, col).FormulaA1 = "=AVERAGE(C2:C30)";               // works
     ws.Cell(row + 1, col).FormulaA1 = "=MEDIAN(C2:C30)";          // Exceptions
}


![MEDIAN Exception](https://user-images.githubusercontent.com/6248847/131186342-0240dd7a-5048-4664-a1d9-020a491546e4.jpg)



- [ ] I attached a sample spreadsheet.  (You can drag files on to this issue)

CaptBli:
We need MEDIAN.  This would help us.  Thank you for posting this.

StevenJonSmith:
Can I throw ROW() into this request as well please.

PR #1726, opened 2021-10-07
Implements MEDIAN function

jorgegiro:
I've implemented the MEDIAN function. This is my first contribution. I'll be happy to receive feedback and suggestions on changes that I should make.

Resolves ClosedXML/ClosedXML#1715


rominator1983:
I pulled your commits to here https://github.com/stesee/ClosedXML/pull/19

PR #1730, opened 2021-10-14
Warn ongdiplus

leotsarev:
Based on #1696 and #1695

Idea is to warn on Linux on all APIs that require GDI+


rominator1983:
in the meantime you can check out https://github.com/stesee/ClosedXML where there was a switch to SkiaSharp and support for linux

leotsarev:
Superseded by https://github.com/ClosedXML/ClosedXML/pull/1866

PR #1742, opened 2021-11-29
Ensure DateTimes with a locale prefix in their format code are still correctly interpreted as DateTimes

Jonathan--V:
### Motivation

XLWorkbook_Load.GetDataTypeFromCell attempts to deduce the data type using the format code. Normally, this works well. However, in the case of a DateTime format string with a Locale prefix containing a 0, the guess is a number. This then causes anything using the value to receive the number of days since 1/1/1900 (such as 43847) rather than an actual DateTime. This, in turn, leads to a (caught) exception when ExcelNumberFormat tries to format the number as a DateTime. So, XLCell.Value and XLCell.GetFormattedString() will both return the number of days since 1/1/1900 instead of a DateTime.

### Solution

Ignore any characters ocurring within square brackets when deducing the data type.

The locale codes containing 0s are then skipped over and subsequent characters can be used for deducing the format code. This solution is consistent with the current solution for skipping values in double quotes.

All tests pass with the proposed change. It is possible that there is some situation where the new version could let a number be incorrectly interpreted as a DateTime, but I have not found it.

### What this doesn't address

- Dates specified as being in a locale that uses a language other than English will continue to be incorrect if they use any words in that language.
- Dates that start with "[$-x-sysdate]" such as "[$-x-sysdate]dddd, mmmm dd, yyyy" will continue to use the format code ignoring the [$-x-sysdate] prefix, which is different from Excel's behaviour.


jahav:
This has been merged as part of #2184.

PR #1761, opened 2022-02-04
Remove the custom preprocessor symbols

0xced:
The .NET SDK already provides target framework [preprocessor symbols](https://docs.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-a-target-framework), there is no need to define additional ones.

Also, it's 2022 so the commented out code that says "keeping it around for in case we bring back .NET3.5 support" has been removed.

### DIFF
--- a/ClosedXML.Examples/ClosedXML.Examples.csproj
+++ b/ClosedXML.Examples/ClosedXML.Examples.csproj
@@ -13,18 +13,6 @@
     <RootNamespace>ClosedXML.Examples</RootNamespace>
   </PropertyGroup>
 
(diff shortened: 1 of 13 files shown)


rominator1983:
already did that here https://github.com/ClosedXML/ClosedXML you can get it from https://www.nuget.org/packages/DocumentPartner.ClosedXML

0xced:
Thanks for merging! Does that mean that you are taking over the project maintenance, Amir?

AmirBosch:
@0xced , for now yes , until @igitur  comes back :)

PR #1767, opened 2022-02-20
Pivot Table custom captions GrandTotalCaption and DataCaption save+load

eagleoriginal:
Posiibility setup and load Custom Captions in previously created excel templates for example.


rominator1983:
I dragged your changes over here https://github.com/stesee/ClosedXML/pull/26 and added a missing test for the null values.

Once it is merged it will become available here https://www.nuget.org/packages/DocumentPartner.ClosedXML

PR #1776, opened 2022-03-22
Add default format code for numbering formats to avoid exception in xl\styles.xml file

dontbyte:
Add default format code for numbering formats, otherwise Excel will throw an Error in xl\styles.xml file (and all other sheel{...}.xml files as well.


ggrote:
Thanks for the fix. Will it be integrated in the develop branch?

PR #1789, opened 2022-05-03
[WIP] Add method to get a worksheet by it's ID

### LINKED ISSUE #1786: Get a worksheet by ID

- [ ] Bug
- [X] Feature
- [ ] Question

- [ ] Yes
- [X] No ► I had a look in the latest source and saw that the issue is still there

**Version of ClosedXML**

0.95.4

It's only possible to get a sheet by position or by name, but not by `SheetId`. this leads to porblems if a user renames a sheet or reorders the sheets. The `SheetId` on the other hand will not change. 

It's possible to get a sheet by its ID or at least its possible to read the ID. The property is already there: 
https://github.com/ClosedXML/ClosedXML/blob/ce7fb695a70d2e870078db4efe143a9e90ac7ad2/ClosedXML/Excel/XLWorksheet.cs#L150

No

## Reproducibility

Consider this template file with two sheets: 

```
Template file: 
---------------
1: "Sheet A" (Position 1)
2: "Sheet B" (Position 2) 

User 1 renamed the sheets to German: 
------------------------------------------
1: "Blatt A" (Position 1)
2: "Blatt B" (Position 2) 

User 2 reordered the sheets: 
---------------
2: "Sheet B" (Position 1) 
1: "Sheet A" (Position 2)
```

```c#
public void Main()
{
     using var wb = new XLWorkbook(_fileName);
     var ws1 = wb.Worksheet("Sheet A"); // This will fail for User 1
     var ws2 = wb.Worksheet(1);         // This will fail for User 2 
}
```

[GetSheetById_Template.xlsx](https://github.com/ClosedXML/ClosedXML/files/8554302/GetSheetById_Template.xlsx)
[GetSheetById_User_1.xlsx](https://github.com/ClosedXML/ClosedXML/files/8554303/GetSheetById_User_1.xlsx)
[GetSheetById_User_2.xlsx](https://github.com/ClosedXML/ClosedXML/files/8554304/GetSheetById_User_2.xlsx)

I hope the idea is getting clearer now. 

If you don't mind I can make a PR to handle this. 

Happy coding 
Tim

timunie:
This PR adds two functions to get a worksheet by it's ID. This helps if the name or position can change, but the ID not. 

`public IXLWorksheet GetWorksheetById(int id)` ► Gets the worksheet by ID or throws if no sheet with the given ID was found
`public bool TryGetWorksheetById(int id, out IXLWorksheet result)`  ► Doesn't throw, but requires a prameter with our.

Closes #1786

### DIFF
--- a/ClosedXML.Tests/Excel/Worksheets/XLWorksheetTests.cs
+++ b/ClosedXML.Tests/Excel/Worksheets/XLWorksheetTests.cs
@@ -1242,5 +1242,33 @@ namespace ClosedXML.Tests
                 Assert.AreEqual(XLWorksheetVisibility.Visible, wb.Worksheets.First().Visibility);
             }
         }
+
+        [Test]
+        public void GetWorksheetById()
+        {
+            using (var wb = new XLWorkbook())
+            {
+                wb.Worksheets.Add("Sheet 1");
+                wb.Worksheets.Add("Sheet 2");
+
+                // Get sheet number 2 and assert that the name is Sheet 2
+                Assert.IsTrue(wb.Worksheets.GetWorksheetById(2).Name == "Sheet 2");
+
+                // Test if an exception is thrown if sheet with the given ID is not found
+                Assert.Throws<InvalidOperationException>(() => wb.Worksheets.GetWorksheetById(3));
+
+                // Test if TryGetWorksheetById returns true if a sheet is found and otherwise false
+                Assert.IsTrue(wb.Worksheets.TryGetWorksheetById(2, out var result1) && result1 != null);
+                Assert.IsTrue(!wb.Worksheets.TryGetWorksheetById(3, out var result2) && result2 == null);
+
+                // Rename sheet 1 and check if the name is still valid
+                wb.Worksheet(1).Name = "New Name";
+                Assert.IsTrue(wb.Worksheets.GetWorksheetById(1).Name == "New Name");
+
+                // Reorder the sheets and test again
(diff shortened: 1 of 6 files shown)


igitur:
Please add appropriate unit tests.

timunie:
@igitur If you like the idea, sure. I can add these tests. Thank you for the fast feedback.

timunie:
It was a good point to add unit tests. Turned out I have not done it correctly in the first step. Therefore added WIP for now.

timunie:
Points out that we have the following issue here: 

In Open XML `SheetId` can be different from `r:Id`, but `SheetId` must be unique I think. But ClosedXml seems to handle this differently. Each added sheet has `SheetId=0` by default. When you change this by increment a counter, It works for the demo App but many many tests fail as they compare to sometimes different IDs. 

from the [ECMA-376](https://www.ecma-international.org/publications-and-standards/standards/ecma-376/):
image
 
Is this little function worth the effort? 

I'll push my recent changes, so you can check on your own. A short feedback if I should keep on working on this would be great.

jahav:
I am closing this a WON'T FIX.

I don't think `SheetId` is something that should be exposed to user, it's an internal id. ClosedXML can just renumber sheetIds as it sees fit (which might happen, I plan on using them for dealing with sheet renaming). There are several other reasons e.g. ids are a mess and exposing it makes everything worse, it's not even exposed to VBA and I am more judicial about increasing API surface.

timunie:
@jahav understood, I totally understand this. Thank you for getting back to me and letting me know. 💐

PR #1800, opened 2022-05-26
Implement PMT function

### LINKED ISSUE #1777: Function PMT was not recognised.

- [ ] Bug
- [x] Feature
- [ ] Question

- [x] Yes
- [ ] No

**Version of ClosedXML**

e.g. 0.95.3

Complete this.

Complete this.

Regressions get higher priority. Test against the latest build of the previous minor version. For example, if you experience a problem on v0.95.3, check whether it the problem occurred in v0.94.2 too. 

## Reproducibility

```c#
public void Main()
{
    // Code standards:
    // - Fully runnable. I should be able to copy and paste this code into a 
    //   console application and run it without having to edit it much.
    // - Declare all your variables (this follows from the previous point)
    // - The code should be a minimal code sample to illustrate issue. The code 
    //   samples on the wiki are good examples of the terseness that I want. Don't
    //   post your full application.
}
```

jahav:
Implement PMT function to calculate to calculate loan payments. This is my first contribution.

I am slightly confused about behavior of parameters, spec (Part 1 18.17.7.253) says 
> Howver, if type is any number other than 0 or 1, #NUM! is returned.

But when I put number other than 0/1 into Excel, it just accepted any number. On the other hand, the number of periods 0 returned NUM error, which is behavior not described in the spec. I used excel as the correct source.

The implementation was done using EMI (e.g. http://rmathew.com/2006/calculating-emis.html) and by comparing it to the excel values.

Resolves #1777

### DIFF
--- a/ClosedXML.Tests/Excel/CalcEngine/StatisticalTests.cs
+++ b/ClosedXML.Tests/Excel/CalcEngine/StatisticalTests.cs
@@ -342,6 +342,65 @@ namespace ClosedXML.Tests.Excel.CalcEngine
         }
 
         [Test]
+        public void Median_CellRangeOfNonNumericValues_ThrowsApplicationException()
+        {
+            //Arrange
+            var ws = workbook.Worksheets.First();
+
+            //Act - Assert
+            Assert.Throws<ApplicationException>(() =>
+            {
+                ws.Evaluate("AVERAGE(D3:D45)");
+            });
+        }
+
+        [Test]
+        public void Median_EvenCountOfCellRange_ReturnsAverageOfTwoElementsInMiddleOfSortedList()
+        {
+            //Arrange
+            var ws = workbook.Worksheets.First();
+
+            //Act
+            double value = ws.Evaluate("MEDIAN(I3:I10)").CastTo<double>();
+
+            //Assert
+            Assert.AreEqual(244.225, value, tolerance);
+        }
+
+        [Test]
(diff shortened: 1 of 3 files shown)


ClosedXML/Excel/CalcEngine/Functions/Financial.cs
```diff
@@ -60,5 +64,31 @@ public static void Register(CalcEngine ce)
             // YIELDDISC Returns the annual yield for a discounted security; for example, a Treasury bill
             // YIELDMAT Returns the annual yield of a security that pays interest at maturity
         }
+
+        private static object Payment(List<Expression> p)
```
igitur, line 68: Is it possible to name this method `Pmt` to align with the Excel function name?

ClosedXML/Excel/CalcEngine/Functions/Financial.cs
```diff
@@ -60,5 +64,31 @@ public static void Register(CalcEngine ce)
...
+        {
+            double rate = 1.0 + p[0];
```
igitur, line 70: I'd prefer `double rate = p[0];` with adjustments below accordingly.

igitur: And then when `rate == 0.0` none of the `Math.Pow()` functions would be needed.

igitur (changes requested):

ClosedXML/Excel/CalcEngine/Functions/Financial.cs
jahav: Haven't noticed it. Thanks.

PR #1802, opened 2022-05-28
Unify error handling in expressions.

### LINKED ISSUE #1763: VLOOKUP doesn't work correctly
- [x] Bug
- [ ] Feature
- [ ] Question

- [x] Yes
- [ ] No

**Version of ClosedXML**

0.95.4

When reading the Cell.Value, in C#, from a valid .xlsx file I receive a "No matches found" exception message.

Cell.Value should contain the valid string value as displayed in the worksheet.

Don't believe so.

## Reproducibility
The Excel cell formula that is causing the issue is:
```Excel
=IF(ISNA(VLOOKUP($A2,Sheet2!$A$2:$B$10,1,FALSE)),"Default",IF(B2>VLOOKUP($A2,Sheet2!$A$2:$B$10,2,FALSE),"Found","Not Found"))
```

```c#
public void Main()
{
    DataTable dt = null;
    string filePath = "No Match Found - Sample.xlsx";  //Add path to Excel Spreadsheet here

    if (File.Exists(filePath)) {
        using (var fileStream = File.OpenRead(filePath)) {
            using (var workbook = new XLWorkbook(fileStream)) {
                var ws = workbook.Worksheet("Sheet1");
                    dt = ToDataTable(ws);
            }
        }
    }
}

public static DataTable ToDataTable(IXLWorksheet worksheet, int headerRow = 1) {
    var dt = new DataTable(worksheet.Name);
    //Range for reading the cells based on the last cell used.
    var maxCol = 1;
    worksheet.RowsUsed().ForEach(r => maxCol = Math.Max(r.LastCellUsed().Address.ColumnNumber, maxCol));
    var readRange = $"1:{maxCol}";
    worksheet.RowsUsed().ForEach(row => {
        if (row.RangeAddress.FirstAddress.RowNumber == headerRow) { //If reading the First Row (used) then add them as column name
            row.Cells(readRange).ForEach(cell => {
                var uniqueColumnNum = 1;
                while (cell.Address.ColumnNumber > dt.Columns.Count + 1) {
                    dt.Columns.Add("Untitled_Column_" + (dt.Columns.Count + 1).ToString());
                }
                var headerName = cell.Value.ToString().Replace('\n', ' ').Trim();
                while (dt.Columns.Contains(headerName)) {
                    headerName = cell.Value.ToString().Replace('\n', ' ').Trim() + "_" + uniqueColumnNum++;
                }
                dt.Columns.Add(headerName);
            });
            while (maxCol >= dt.Columns.Count + 1) {
                dt.Columns.Add("Untitled_Column_" + (dt.Columns.Count + 1).ToString());
            }
        } else if (row.RangeAddress.FirstAddress.RowNumber > headerRow) {
            //Adding a Row in datatable
            dt.Rows.Add();
            var cellIndex = 0;
            //Updating the values of datatable
            row.Cells(readRange).ForEach(cell => {
                dt.Rows[dt.Rows.Count - 1][cellIndex] = cell.Value.ToString();  //This is where the exception occurs, "No Matches Found"
                cellIndex++;
            });
        }
    });
    return dt;
}

```

[No Match Found - Sample.xlsx](https://github.com/ClosedXML/ClosedXML/files/8047212/No.Match.Found.-.Sample.xlsx)

jahav:
Streamline error handling in formulas. Errors should be always thrown as CalcEngineException exception instead of an return of an error values. That is in line with section 18.17.3, that basically says if an anything in expression causes this, the result of expression is an error (with exception of methods that work on error values).

Also, it makes various function implementation far easier, no need to check for error values anywhere other than functions that should handle it.

I only worked on error handling functions other in Informations.cs, remainder is out of scope of this PR. I have significant doubts about others (obvious errors and even typos in function names, e.g ERRORTYPE instead of ERROR.TYPE). I made some small changes to TYPE so it work with errors, but I would lump it with the others.

### Why exceptions
Error handling is rather rather incosistent.
* Most of code uses exceptions derived from `CalcEngineException`. That is true for both function implementation and CalcEngine processing of operations
* Some functions (mostly from Information.cs) return an error enum `ExpressionErrorType`
* Some functions return other exceptions (e.g. FIND returns `ArgumentException` when it doesn't find the text, MEDIAN returns `ApplicationException` on empty).

This PR basically says that CalcEngineException is the correct state and functions should use it and fixes error checking functions. 
Resolved #1763

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Functions/Information.cs
+++ b/ClosedXML/Excel/CalcEngine/Functions/Information.cs
@@ -10,13 +10,13 @@ namespace ClosedXML.Excel.CalcEngine.Functions
         public static void Register(CalcEngine ce)
         {
             //TODO: Add documentation
-            ce.RegisterFunction("ERRORTYPE", 1, ErrorType);
-            ce.RegisterFunction("ISBLANK", 1, int.MaxValue, IsBlank);
-            ce.RegisterFunction("ISERR", 1, int.MaxValue, IsErr);
-            ce.RegisterFunction("ISERROR", 1, int.MaxValue, IsError);
+            ce.RegisterFunction("ERROR.TYPE", 1, ErrorType);
+            ce.RegisterFunction("ISBLANK", 1, 1, IsBlank);
+            ce.RegisterFunction("ISERR", 1, 1, IsErr);
+            ce.RegisterFunction("ISERROR", 1, 1, IsError);
             ce.RegisterFunction("ISEVEN", 1, IsEven);
             ce.RegisterFunction("ISLOGICAL", 1, int.MaxValue, IsLogical);
-            ce.RegisterFunction("ISNA", 1, int.MaxValue, IsNa);
+            ce.RegisterFunction("ISNA", 1, 1, IsNa);
             ce.RegisterFunction("ISNONTEXT", 1, int.MaxValue, IsNonText);
             ce.RegisterFunction("ISNUMBER", 1, int.MaxValue, IsNumber);
             ce.RegisterFunction("ISODD", 1, IsOdd);
@@ -27,34 +27,51 @@ namespace ClosedXML.Excel.CalcEngine.Functions
             ce.RegisterFunction("TYPE", 1, Type);
         }
 
-        static IDictionary<ErrorExpression.ExpressionErrorType, int> errorTypes = new Dictionary<ErrorExpression.ExpressionErrorType, int>()
+        private static object ErrorType(List<Expression> p)
         {
-            [ErrorExpression.ExpressionErrorType.NullValue] = 1,
-            [ErrorExpression.ExpressionErrorType.DivisionByZero] = 2,
-            [ErrorExpression.ExpressionErrorType.CellValue] = 3,
-            [ErrorExpression.ExpressionErrorType.CellReference] = 4,
-            [ErrorExpression.ExpressionErrorType.NameNotRecognized] = 5,
-            [ErrorExpression.ExpressionErrorType.NumberInvalid] = 6,
-            [ErrorExpression.ExpressionErrorType.NoValueAvailable] = 7
-        };
+            try
+            {
+                p[0].Evaluate();
+            }
+            catch (NullValueException)
+            {
+                return 1;
+            }
+            catch (DivisionByZeroException)
+            {
+                return 2;
+            }
+            catch (CellValueException)
+            {
+                return 3;
+            }
+            catch (CellReferenceException)
+            {
+                return 4;
+            }
+            catch (NameNotRecognizedException)
+            {
+                return 5;
+            }
+            catch (NumberException)
+            {
+                return 6;
+            }
+            catch (NoValueAvailableException)
+            {
+                return 7;
+            }
 
-        static object ErrorType(List<Expression> p)
-        {
-            var v = p[0].Evaluate();
-
-            if (v is ErrorExpression.ExpressionErrorType)
-                return errorTypes[(ErrorExpression.ExpressionErrorType)v];
-            else
-                throw new NoValueAvailableException();
+            throw new NoValueAvailableException();
         }
 
-        static object IsBlank(List<Expression> p)
+        private static object IsBlank(List<Expression> p)
         {
-            var v = (string) p[0];
+            var v = (string)p[0];
             var isBlank = string.IsNullOrEmpty(v);
 
-
-            if (isBlank && p.Count > 1) {
+            if (isBlank && p.Count > 1)
+            {
                 var sublist = p.GetRange(1, p.Count);
                 isBlank = (bool)IsBlank(sublist);
             }
@@ -62,33 +79,48 @@ namespace ClosedXML.Excel.CalcEngine.Functions
             return isBlank;
         }
 
-        static object IsErr(List<Expression> p)
+        private static object IsErr(List<Expression> p)
         {
-            var v = p[0].Evaluate();
-
-            return v is ErrorExpression.ExpressionErrorType
-                && ((ErrorExpression.ExpressionErrorType)v) != ErrorExpression.ExpressionErrorType.NoValueAvailable;
+            try
+            {
+                p[0].Evaluate();
+                return false;
+            }
+            catch (NoValueAvailableException)
+            {
+                return false;
+            }
+            catch (CalcEngineException)
+            {
+                return true;
+            }
         }
 
-        static object IsError(List<Expression> p)
+        private static object IsError(List<Expression> p)
         {
-            var v = p[0].Evaluate();
-
-            return v is ErrorExpression.ExpressionErrorType;
+            try
+            {
+                p[0].Evaluate();
+                return false;
+            }
+            catch (CalcEngineException)
+            {
+                return true;
+            }
         }
 
-        static object IsEven(List<Expression> p)
+        private static object IsEven(List<Expression> p)
         {
             var v = p[0].Evaluate();
             if (v is double)
             {
-                return Math.Abs((double) v%2) < 1;
+                return Math.Abs((double)v % 2) < 1;
             }
             //TODO: Error Exceptions
             throw new ArgumentException("Expression doesn't evaluate to double");
         }
 
-        static object IsLogical(List<Expression> p)
+        private static object IsLogical(List<Expression> p)
         {
             var v = p[0].Evaluate();
             var isLogical = v is bool;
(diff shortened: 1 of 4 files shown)


ClosedXML/Excel/CalcEngine/Functions/Information.cs
```diff
@@ -27,95 +27,136 @@ public static void Register(CalcEngine ce)
...
+                return 5;
+            }
+            catch (NumberException)
+            {
+                return 6;
+            }
+            catch (NoValueAvailableException)
+            {
+                return 7;
+            }
```
Pankraty, line 63: Not insisting, but this can be shortened:
```suggestion
            catch (CalcEngineException ex)
            {
                return ex switch
                {
                    NullValueException _ => 1,
                    DivisionByZeroException _ => 2,
                    CellValueException _ => 3,
                    CellReferenceException _ => 4,
                    NameNotRecognizedException _ => 5,
                    NumberException _ => 6,
                    NoValueAvailableException _ => 7,
                    _ => throw new NoValueAvailableException()
                };
            }
```

Pankraty (approved):

ClosedXML/Excel/CalcEngine/Functions/Information.cs
jahav: Thanks, I didn't realize that. The shorter version is far more readable.

Pankraty (approved):

igitur:
Thanks for this, @jahav. 

So you hit a lot of legacy issues here, and just judging from the whitespace fixes, it seems we haven't touch these files in a long time.

I think I introduced `ErrorExpression.ExpressionErrorType` specifically to avoid `try`/`catch` blocks, which, I believe, can negatively impact performance. I'm open to being corrected here. I agree that `try`/`catch` simplifies the code a lot and the handling currently inconsistent, but I'm not yet convinced the favouring `try`/`catch` over `ErrorExpression.ExpressionErrorType` is the way to go.

I did related work in #1404 where I wanted to replace the enum value with exception classes, but instead of throwing them, I return them as references, but @Pankraty felt it was too esoteric. I still like that approach, because [everywhere](https://stackoverflow.com/a/52322/179494) I [read](https://stackoverflow.com/a/39820555/179494) people claim that unwinding the stack during a `catch` has performance issues. And I prefer returning a class rather than an enum value because we can benefit from properties that add information about the context.

Let's discuss the direction we should go before working on the actual PR, if that's OK.

Thanks for dragging us back into the thick of things ;-)

jahav:
The PR was mostly to unify the behavior that was inconsistently used, but I am all for setting a better, canonical standard.

# Performance

I've run a simple [benchmark](https://gist.github.com/jahav/2d621605230a54a23a84019bac5e2de4) comparing `ABS(-1)/ABS(0)` against `ABS(-1)/ABS(1)`. One of formulas returns an exception for #DIV/0! and other just number. Results:

|        Method |         Mean |      Error |     StdDev |
|-------------- |-------------:|-----------:|-----------:|
|   ReturnValue |     61.06 ns |   1.304 ns |   1.281 ns |
| ExceptionPath | 31,077.00 ns | 371.773 ns | 347.757 ns |

When I used XLWorkbook.EvaluateExpr instead of XLCell.Value, the values were [pretty similar](https://gist.github.com/jahav/677328404774d5551ce1174373da3237) without cell overhead.

Then I made a more [pure benchmark](https://gist.github.com/jahav/3b7584b3722ce0e131196f4224fe28be) without ClosedXml comparing returning an error value with exception (every method public, so hopefully not optimized into oblivion).

|         Method |          Mean |      Error |     StdDev |
|--------------- |--------------:|-----------:|-----------:|
|    ReturnValue |      1.104 ns |  0.0499 ns |  0.0442 ns |
| ExceptionValue | 10,727.865 ns | 79.3920 ns | 74.2633 ns |


Yeah... not great. Although errors shouldn't be that frequent, it's a reasonably normal occurance (e.g. `VLOOKUP` return #N/A if no match found). 31ms for a single error formula seems too much (imagine VLOOKUP for 10k rows where half lookups won't find anything -> exception = 155 seconds). I would agree with @igitur that exceptions are not very perspective idea from performance standpoint.

# Musing about current structure of CalcEngine

My view (being immature in all things OOXML and ClosedXML) is mostly directed by spec.
> #### 18.17.2.6 Types and Values:
> Each expression has a type. SpreadsheetML formulas support the following types: array, error, logical, number,
and text.

As far as spec is concerned, error is a type that can have several values, not an exception.

Second thing is that ClosedXML should support all type, including arrays, as input/output. There are [several functions](https://support.microsoft.com/en-us/office/excel-functions-that-return-ranges-or-arrays-7d1970e2-cbaa-4279-b59c-b9dd3900fc69) that return range/array and more that take array as input.

Calc engine should probably also perform array calculation for scalar functions by itself. The function shouldn't deal with that (e.g. SQRT takes a number and outputs number, if formula is SQRT({1;2;3;4}), output is an array, but SQRT shouldn't be responsible for anything other than scalar number-scalar number). Edge cases sucks, but they are important to think in the beginning.

Handling input/output should be pretty straightforward, so anyone implementing a function can implement it without diving deep into ClosedXML. I think that should be a major concern.

Currently, each function gets a list of expressions. I feel that we are missing an abstraction of an excel value that can represent possible input/output. Functions operate over expressions, not values. Maybe it's just me, but that seems so wrong, two responsibilities in one (expression is a recipe, result value is the cake). 

Let's take `TYPE ` function for example: 
|  Name |  Type |  Description   |
|-------|-------|----------------|
| value | any   | The value whose type is to be determined. No conversion shall take place on an argument passed to this function. |

That little note about _no conversion_ is not possible with current API (because there is no abstraction of a value, Expression is always converted to a value) without resorting to something like this (or GetArray in matrix multiplication):

```
        private static object SumIf(List<Expression> p)
        {
            // get parameters
            var range = p[0] as IEnumerable;
            var sumRange = p.Count < 3 ?
                p[0] as XObjectExpression :
                p[2] as XObjectExpression;   // At least expression are mostly hidden, but sometimes it just spills out
```

So I think in order to facilitate easy implementation of functions, we should make something like `XLValue` to represent (=errors, number, string, ranges, arrays...) and use them as input/output of functions instead of expression.

It also limits things like array calculation or possible features like differential calculation where only things that need to be recalculated are recalculated and so on.

## Back to topic exceptions vs return values
Spec says that 
> 18.17.3 Error values
> The evaluation of an expression can result in an error having one of a number of error values. These error values
are:
> [Note: A number of functions operate on error values: They include ERROR.TYPE (§18.17.7.110), ISERR
(§18.17.7.175), ISERROR (§18.17.7.176), and ISNA (§18.17.7.179). end note]

To rephrase (as I understand it): If anything in the expression tree results in an error, whole tree evaluates into the error, unless there is one of specific function.

I would modify CalcEngine so that instead of expressions being evaluated in functions directly, the engine would evaluate expressions by itself and pass the resulting values to the function (probably through some transparent proxy, so we don't evaluate arguments that don't need to be evaluated - but it seems that all parameters are always needed, so maybe this is useless).

Function declarations would get some flags, like accept errors, is scalar only (so calc can handle array calculations)...

When it would detect an error, it would stop and went up the tree to find one the error-aware functions. If there was one, it would replace whole subtree with the error and continued. If no such function was in the tree, just replace whole expression with error value.

This way, functions would return error values not as an exception, but as errors. Unless function would ask for error inputs through flags.

I would probably use pattern similar to this in functions to convert values and it would be used across functions.
```
XLValue Sin(List<XLValue> p)
{
  // Since CalcEngine would stop processing on error, the original error
  // wouldn't be replaced by this one up in the tree on failed conversion.
  if (!p[0].TryConvert(out double angle))
     return XLError.NoValueAvailable; // #VALUE!, probably implicit cast to XLValue
  return Math.Sin(angle);
}
```

Basically
* CalcEngine - responsible for orchestration
* XLValue - responsible for holding and converting values.
* function - responsible only for calculation, operates on values.

Pankraty:
If CalcEngine is being redesigned, we must take into account evaluation context: we've been asked so many times to add support of `ROW` and `COLUMN` function, and this requires the function to be aware of where it is executed. For now we only need these two fields (current row and current column) but if we ever add support for table-style refences, the context will need to be extended. 

For not generating heavy traffic of the context objects that need to be GCed, we can think of declaring it as `ref readonly struct`, and it is not possible for some reason, experiment with polling: at any given time we may need as many `context` objects, as deep is recursion in formulas, which is usually just a few levels. So why not reuse the same entities after computation chain is over? 

For inputs/outputs the problem with large traffic of objects also exists. It becomes worse as we are packing numerals, bools and dates into `object`. Some time ago a came up with possible solution allowing to have variables of a "primitive value or string" without boxing/unboxing and without wasting too much memory (if weights as `double`, as I remember) - see https://github.com/pankraty/NoBox. However, I did not make nature experiments that would prove the solution is viable.

jahav:
Hi @igitur, @Pankraty,

I think it's clear that CalcEngine has severe problems and requires a significant change to support possible features of OOXML. Since it's hard to decide on something without a some kind of proposal to critique, I made one ^_^. 

I don't think it belongs into the PR discussion, so I made an issue #1812 for it and wrote it there, please check it out. It's a first draft, so I expect a lot of improvements and fixes.

Idea is to reach a consensus on a general idea of a design and data structures. The key requirement is that design is able to support all things related to formula calculation (whether they are/will be implemented implemented is a different question, but it should be possible to add them without change of basic design).

This PR is not going to make it because is not a good solution to the problem of error propagation, so I am closing it.

I will probably concentrate on the System.Drawing.Common while we discuss the way forward for CalcEngine.

Issue #1805, opened 2022-06-08
Remove System.Drawing.Common dependency


jahav:
**tl;dr;** It's a quagmire. The issues is font measurement. Options:
1. Either dump net40/net46 target and use SixLabors.Fonts
2. Use SkiaSharp wrapper (no net40, need to upgraed to net462) over native Skia library (with all baggage of various packages for OS/arch and no Blazor)
3. Combine SixLabors.Fonts measurement for netstandard2.0 and System.Drawing.Common for net40/net46 (yay for two choices of one feature that need pixel perfect precision that must work with tests).
4. Use simplified DIY font measurement mechanism that uses glyph typographic information (bad idea (tm))

Option 3. is seems like the best choice. I made an attempt and it seems to work reasonably well. Tests already ignore dimensions for linux (default Calibri Windows fonts are not available), will use some percentage for allowed difference (e.g. 1-3%).

As for images, library is an option. There are several libraries, but we only need such a limiting subset that it might be better to just parse dimension information from images in ClosedXML directly (width, heigh, dpi).

# Problem
ClosedXML is using System.Drawing for working with XLPicture. `System.Drawing.Common` is dead and buried, [since NET6](https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/system-drawing-common-windows-only), it only works on Windows, leaving other platforms, from Linux through Android to Blazor incapable of working with images. It is a source of several reported issues.

# System.Drawing.Common usage
ClosedXML is using System.Drawing.Common
* to determine format and size of images (`Image.FromStream()`)
  * XLPictureFormat - for storing content type of Image part (15.2.14) that represents the image in workbook.
  * Width and height - for storing extent of the image and its position (20.5.2.30, 19.3.1.53) in the drawing part.
* Measuring bounding box of a text 
  * `Font` - read the specification of font
  * `FontStyle` - style of how to render font
  * MeasureString to autosize content
  * DpiX, DpiY taken from GraphicsUtils to determine Dpi and measure stuff. This seems wrong and should be an part of image specification/option. Determining EMU from DPI of screen is not ideal
* Translate colors from HTML to standard Color (`ColorTranslator.FromHtml`)

# Research
There are several ways to solve this issue:
[...]

PR #1806, opened 2022-06-08
WIP: System.Drawing.Common font measurement replacement

jahav:
This PR is an attempt to decrease dependency on System.Drawing.Common by using two different ways for font measurement:
* net40,net46 - keep System.Drawing.Common, it's windows only and will continue to work
* netstandard2.0 - use SixLabors.Fonts library to determine font measure.

I am not happy with having two different implementation, but it seems like best of bad options. Opinions are welcome.

PR is a part of of changes for #1805


igitur:
If we drop support for net40 and net461 and replace with net472, would that solve the issue?  If so, we should have a discussion about which frameworks we want to support going forward.

LeaFrock:
As net40/net461 is EOL, I think it's not bad to drop them in order to make the maintenance easier. Of course it should be a change of major version(`1.0`?). People who stay on old NFX versions can still use `0.9x`.

`net standard 2.0` is compatible to net462 and above. IMO, `net standard 2.1` , which has more exposed APIs,  is optional but more suitable for future development of this repo.

If so, `System.Drawing.Common` can be totally replaced by `ImageSharp`.

igitur:
> Of course it should be a change of major version(`1.0`?). People who stay on old NFX versions can still use `0.9x`.

Assuming we follow semver, even versions < 1.0 are allowed breaking changes.

jahav:
I am closing this PR in favor of #1866. The other PR is more capable (e.g. EMF file format) and it is more in tune with [Cell dimensions](https://github.com/ClosedXML/ClosedXML/wiki/Cell-Dimensions).

PR #1821, opened 2022-06-21
WIP Pivot subtotal styles

jahav:
Add an ability to remove a subtotal from a pivot table and fix the styles. ClosedXML should be clear that default subtotal is in the field. ASIS - empty Subtotals mean default, TOBE - empty subtotals means no subtotal, XLSubtotalFunction.Automatic means default, any additional subtotal function overrides the default.

It modifies the API, from `XLPivotField.AddSubtotal(XLSubtotalFunction)` to `XLPivotField.SetSubtotal(XLSubtotalFunction, bool)`, but it is either this or additional functions like `RemoveSubtotal`. Since it is not possible to affect the order of subtotals, this seems better.

Second thing is styling - it's not possible to style other subtotal function other than default one. even this is a partial solution, because it doesn't style label of a subtotal, but at least it is an improvement.


igitur:
Hi, I definitely want to review this one before merging, please.

jahav:
@igitur It is missing ability to load saved file that has subtotals in some fields and and not in others. I had some problems with XML comparing, so I worked on it. It is mostly to solve issue that I can't style functions other than [default subtotal](https://github.com/ClosedXML/ClosedXML/wiki/Pivot-Table-Styles#subtotals-at-the-bottom-of-groups).

PR #1824, opened 2022-06-23
Xml compare

jahav:
Compare two XML files using XML structure and not a string. The key issue for this is that DocumentFormat.OpenXml was generating attribtues in a different order. Also, regexp are seriously not a good to manipulate xml (or any other nonregular language).


igitur:
Hi, thanks for this. Please use a Rebase merge where possible. We try to keep the commit history linear. We make exceptions (Squash & Rebase or normal Merge commit) where it makes sense, but the default is rebase.

jahav:
> Please use a Rebase merge where possible. We try to keep the commit history linear. 

Ok, thanks for info.

PR #1825, opened 2022-06-24
Refactor/simplify .csproj files

igitur:
Refactor and simplify .csproj files. Move common properties into its own `Directory.Build.props` file that is included automatically in the build process.

Documentation about Directory.Build.props: https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022

### DIFF
--- a/ClosedXML.Examples/ClosedXML.Examples.csproj
+++ b/ClosedXML.Examples/ClosedXML.Examples.csproj
@@ -2,15 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.1;net40;net46</TargetFrameworks>
-    <LangVersion>8.0</LangVersion>
     <OutputType>Exe</OutputType>
-    <Version>0.95.4</Version>
-    <NoWarn>$(NoWarn);NU1605</NoWarn>
-    <Configurations>Debug;Release</Configurations>
-    <SignAssembly>true</SignAssembly>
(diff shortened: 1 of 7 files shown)


igitur:
Maybe we should call it `common.props` because there's already more than just the version number in the new file. 

And if I recall, there's a special name you can give the file so that it's automatically included in the project builds, so that we don't have to add it to the .csproj file explicitly.

Ah, here it is: https://blog.johnnyreilly.com/2021/07/14/directory-build-props-c-sharp-9-for-all

igitur:
@jahav I want to get this in before bumping the version number and releasing 0.96.0

igitur:
Better documentation about `Directory.Build.props`: https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2022

igitur:
@Pankraty Do you have capacity to add a review?

Pankraty (approved):
Looks good for me. I wish we were able to move TargetFrameworks there too, but this does not seem feasible

igitur:
> I wish we were able to move TargetFrameworks there too, but this does not seem feasible

Do you mean e.g. drop support for net40 and targeting higher versions for .NET Core? I've been thinking about that too, but let's address this in a separate PR. i want to get this one in for 0.96.0.

Pankraty:
Nothing this radical, I just meant, this would be nice to have it unified across the projects, but I understand the reasons to why they differ

PR #1829, opened 2022-06-30
Add more Protect() methods

igitur:
In preparation for the removal of the obsolete `Protect()` methods, I realised we can do with more of the newer style `Protect()`, specifically those that don't require a password, but do accept the other options.

### DIFF
--- a/ClosedXML/Excel/Protection/IXLWorkbookProtection.cs
+++ b/ClosedXML/Excel/Protection/IXLWorkbookProtection.cs
@@ -1,6 +1,15 @@
-﻿namespace ClosedXML.Excel
+﻿// Keep this file CodeMaid organised and cleaned
+using System;
+using static ClosedXML.Excel.XLProtectionAlgorithm;
+
+namespace ClosedXML.Excel
 {
     public interface IXLWorkbookProtection : IXLElementProtection<XLWorkbookProtectionElements>
     {
+        IXLWorkbookProtection Protect(XLWorkbookProtectionElements allowedElements);
+
+        IXLWorkbookProtection Protect(Algorithm algorithm, XLWorkbookProtectionElements allowedElements);
+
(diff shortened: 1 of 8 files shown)


ClosedXML/Excel/Protection/IXLWorkbookProtection.cs
```diff
@@ -1,6 +1,13 @@
...
+using System;
+using static ClosedXML.Excel.XLProtectionAlgorithm;
+
+namespace ClosedXML.Excel
 {
     public interface IXLWorkbookProtection : IXLElementProtection<XLWorkbookProtectionElements>
     {
+        IXLWorkbookProtection Protect(Algorithm algorithm, XLWorkbookProtectionElements allowedElements);
+
+        IXLWorkbookProtection Protect(String password, Algorithm algorithm = DefaultProtectionAlgorithm, XLWorkbookProtectionElements allowedElements = XLWorkbookProtectionElements.Windows);
```
Pankraty, line 13: I was thinking that instead of having multiple overloads with different numbers of parameters, we should probably move toward having a single methdod with options object (e.g. `Protect(XLProtectOptions options)`). This will make adding additional parameters easier, as well as this would better support backward compatibility between different versions of the library. As a drawback, the default values will be a bit more obscure. 
What do you think?

igitur: Maybe later? At this stage I just want to remove the obsolete methods, but to get a like-for-like replacement we need the new Protect methods.

Pankraty: Ah, I see now that there still would be multiple overloads of `Protect` method because of the inherited interfaces. Nevermind then.

Pankraty (approved):

PR #1830, opened 2022-06-30
Remove obsoleted methods

igitur:
Remove obsoleted methods from #1171 and #1029

### DIFF
--- a/ClosedXML.Tests/Excel/Misc/XLWorkbookTests.cs
+++ b/ClosedXML.Tests/Excel/Misc/XLWorkbookTests.cs
@@ -359,7 +359,7 @@ namespace ClosedXML.Tests.Excel
             using (var wb = new XLWorkbook())
             {
                 var ws = wb.Worksheets.Add("Sheet1");
-                wb.Protect(true, false);
+                wb.Protect(XLWorkbookProtectionElements.Windows);
                 Assert.IsTrue(wb.LockStructure);
                 Assert.IsFalse(wb.LockWindows);
                 Assert.IsFalse(wb.IsPasswordProtected);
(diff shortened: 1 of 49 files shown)


ClosedXML/Excel/Ranges/IXLRangeBase.cs
```diff
@@ -118,20 +109,18 @@ public interface IXLRangeBase : IXLAddressable
         IXLCell FirstCellUsed();
 
         /// <summary>
-        ///   Returns the first cell with a value of this range.
+        /// Returns the last cell with a value of this range.
```
Pankraty: Looks like this should be `first`
```suggestion
        /// Returns the first cell with a value of this range.
```

ClosedXML/Excel/Ranges/IXLRangeBase.cs
```diff
@@ -118,20 +109,18 @@ public interface IXLRangeBase : IXLAddressable
...
+        /// <param name="options">The options to determine whether a cell is used.</param>
         IXLCell FirstCellUsed(XLCellsUsedOptions options);
 
         IXLCell FirstCellUsed(Func<IXLCell, Boolean> predicate);
 
-        [Obsolete("Use the overload with XLCellsUsedOptions")]
-        IXLCell FirstCellUsed(Boolean includeFormats, Func<IXLCell, Boolean> predicate);
-
+        /// <summary>
+        /// Returns the last cell with a value of this range.
```
Pankraty: Same here

ClosedXML/Excel/Ranges/IXLRangeBase.cs
Pankraty: Also, the comment may be a bit misleading - cells considered non-empty does not necessarily have a value. Maybe 
```suggestion
        /// Returns the first non-empty cell of this range.
```
?

ClosedXML.Tests/Excel/Misc/XLWorkbookTests.cs
```diff
@@ -359,7 +359,7 @@ public void WbProtect2()
             using (var wb = new XLWorkbook())
             {
                 var ws = wb.Worksheets.Add("Sheet1");
-                wb.Protect(true, false);
```
jahav, line 362: The protection API is rather confusing API. The original parameters true and false were for lockStructure and lockWindows.... but the Protect(XLWorkbookProtectionElements.Windows) indicates which elements are allowed... I would expect it to be opposite.

* `wb.Protect(XLWorkbookProtectionElements.None)` means "nothing is allowed to changes", but it reads as protect no element
* `wb.Protect(XLWorkbookProtectionElements.Everything)` means "everything is allowed to changes", but it reads as protect everything.

jahav (approved):
It is not directly related with the PR, but the API for protection is inversion of what would be expected.

Issue #1812, opened 2022-06-12
CalcEngine redesign


jahav:
# CalcEngine Design

Proposal to modify the CalcEngine in order to

* Be able to support all formula features (either alredy implemented or future ones)
* Enable simple implementation of all predefined function (356 in the spec, 550 in current online documentation)
* Streamline error handing
* Achieve a good performance
* Achieve a better architecture for easier improvements (e.g. we could easily add an ability to calculate functions of large ranges through multiple threads)

The main changes
* Introduce a value type to hold the values that are passed from one expression to other
* Introduce execution context so calculation can be modified based on various conditions (e.g. array formula `=COLUMN()` for `B3:B10` has to know the cell that is being calculated)
* Move the evaluation from expressions to CalcEngine. CalcEngine will evaluate expressions and pass values to other expressions.

# Names
Area = rectangular range

# Unified value type
The type for passing values through calc engine (internal types, not a public API). All values that can be input or output from an expression will be stored in an `XLValue` type. `XLValue` will be able to contain exactly one of
* error - represented as an `enum` (probably ExpressionErrorType from #1404). I think enum is fine, excel doesn't provide additional info about error info either and we can store additional info in CalcContext itself (like a place where error happened). Though I don't have a problem with a different representation.
* logical value - represented by a `bool`
* numeric value - represented by a `double`
* text - represented by a `string`
* array - a rectangular 1D/2D array of `XLValue`s. There can't be an array in an array.
* reference - a reference to a cell. The reference is a binary tree-like structure, where leafs are rectangular sheet areas and nodes are unions/intersections of areas. Reference can also represent 3D reference range. References won't by itself contain data, it will only point to places with data. Reference operations will work even if operands are in different worksheets.

`XLValue` will contain a an enum property to indicate the type within and ability to get the value. The `XLValueType` enum will be different from `XLDataType`. Date/time won't have it's own type and will be represented by a number in serial date/time fashion.

> Sidenote: I am using XL as a prefix for some types, but does ClosedXml use [Smurf Naming Convention](https://blog.codinghorror.com/new-programming-jargon/#21)? Fine for external API, not sure about internal stuff.

[...]

PR #1836, opened 2022-07-14
Replace handcrafted parser with XLParser

jahav:
Our handcrafted parser is no good and we should use more standard compliant parser. XLParser has been selected by consensus. It is a small parser with less than 100 terms (see [official grammar](https://docs.microsoft.com/en-us/openspecs/office_standards/ms-xlsx/3d025add-118d-4413-9856-ab65712ec1b0) for comparison).

*tl;dr;*
* Handcrafted parser has been replaced with XLParser. Drop in replacement without new functionality.
* Attempt at separation of formula parsing and evaluation.
* XLParser produces CST and it is transformed to AST based on *Expression classes. Difference between CST and AST: CST looks like [this](https://i.stack.imgur.com/SyonV.png), AST looks like [this](https://i.stack.imgur.com/dhd3v.png).
* Added few more AST nodes
* Functions weren't modified.
* Covered by tests

# Scope
Goal of the PR is to replace old parser. Goal is not to have new binding for functions or to add formula values. No new features. As long as we have AST, XLParser and test pass, that is good. This is supposed to be a first stage of of CalcEngine redesign #1812.
It doesn't fix any of bad issues of current parser, e.g.:
* Missing separation of parsing and evaluation (=throws on nonexisting ranges and functions)
* Function arguments are complete clusterfuck
* Range operations and many others don't work.

# Abstract Syntax Tree
XLParser produces concrete syntax tree (CST), but that is not suitable for evaluation. CST has to be transformed into abstract syntax tree (AST). [These images sum up the different pretty well](https://stackoverflow.com/a/10176731). ClosedXML is mostly interested in formula evaluation, not formula syntax.

I build sensible AST and then use a visitor to replace some reference nodes with compatibility node XObjectExpression. XObjectExpression will be deleted once we have formula values and sensible evaluation.

Most AST nodes will describe some kind of transformation that outputs a value, though some don't.
* AstNode - base class of all AST nodes
* Expression : AstNode - node that can be evaluated
* ScalarNode : Expression - node that contains a const scalar value.
* ReferenceNode : Expression that evaluates
* PrefixNode : Specifies a file/sheet/3Dsheets for reference node or a function from different sheet.
* ....

Sidenote: I kept name of Expression nodes because they are used everywhere, new ones have suffix `Node`. Will be fixed when values are implemented and term `Expression` is not everywhere.

## Node factories
Irony doesn't have official documentation. I tried to use an implementation style from their samples (e.g. [Variable](https://github.com/IronyProject/Irony/blob/master/Languages/Refal/Ast/Variable.cs)). it was a disaster, full of unclear switches, conditions and other horror. The `AstNodeFactory` is increadible more readable. Microoptimizatins (e.g. using a swicth because it can use binary algorithm to find correct case instead of going through each condition one by one) are overrated.

## State of functions
I knew it was bad, but the state of excel functions in ClosedXml is rather horriyfying. I truly believe we will have to fix/reimplement most of them. 

I made some small changes due to AST changes (enough to make test pass), but it is not possible to fix the functions at this time (=for now, it is a waste of time). We have to wait for next stages of CalcEngine redesign. We will have to develop some kind of fuzzer to compare edge cases of functions with the real result from Excel.

# Things to note:
## License
XParser is under MPL2.0, copyleft file based license. As long as we only use package, we there should be no problem.
Irony is under MIT license.
There is no problem from license standpoint.

## State o Irony
Irony is in bad state. There are several forks that are carried over from codeplex migration (e.g. XLParser is using Irony.NetCore, although there is an official package from IronyProject). 

Documentation is scarce. I had to look through code to find out how it works and what should I do.

Development has stalled. Even though the most annoying bug I have encountered (NRE if CST node doesn't specify some kind of factory for AST node) has been fixed, there is no new release (+fix is only in the official Irony repository that is not used by XLParser).

## NET40
Not sure what to do about ClosedXML target net40, [Irony.NetCore](https://www.nuget.org/packages/Irony.NetCore/1.0.11) and [XLParser](https://www.nuget.org/packages/XLParser) has a lowest version net452, though [Irony](https://www.nuget.org/packages/Irony) has net40 as lowest version. Suppress warnings, ask XLParser to switch to Irony package + target last, remove the target. I am leaning toward the drop of target net40 (ancient, out of official support since 2016, doesn't have IReadOnly collection interfaces, #if defs for Fudy in the codebase for `Dispose` methods). 

## XLParser development
Grammar in XLParser is still changing and upgrading the dependency won't be simple. E.g. since last release 1.5.2, the terms describing columns of structured reference have changed. Extra care and test coverage is a must.


jafin:
+1 for removing net40 support. net462 is I understand the minimum supported framework from Microsoft.

igitur:
> +1 for removing net40 support. net462 is I understand the minimum supported framework from Microsoft.

Yup, but some people complained when we removed support for net35, even though support for that was dropped long ago. Just saying. I still think it's time net40 to go. Maybe netstandard2.0 only going forward?

jahav:
Resources are scarce. Time of developers is valuable. Should we allocate it for a long dead target with incredibly easy update path? We will still target net46.

If they don't update framework, they can keep using old version.

jahav:
@igitur Hi, some time you asked about performance and I made some guesses that weren't very accurate. I made some benchmarks (https://github.com/ClosedXML/ClosedXML/pull/1848#issuecomment-1206784192) and the results are troublesome. XLParser is two orders of magnitude slower than the original parser.

I looked at why is it so slow and what can be done about it. Findings:
* XLParser doesn't specify prefixes for regexps. Irony has a heuristic that looks at next char and limits token checks only to terminals that can even match ([here is the lookup setup](https://github.com/IronyProject/Irony/blob/a77ef332c3d6868ce5326b126efba9d5eeba53c2/Irony/Parsing/Data/Construction/ScannerDataBuilder.cs#L74) and [here is lookup used](https://github.com/IronyProject/Irony/blob/a77ef332c3d6868ce5326b126efba9d5eeba53c2/Irony/Parsing/Scanner/Scanner.cs#L200)). That alone shaved more than 60% of time even in my super quick & dirty & nonrepresentative tests.
* Grammar (thus regexps) is set as case insensitive. Small, but measurable difference. 
* Regexps are also marked as possibly capturing, but it doesn't need it and capturing is [againts best practices](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices#capture-only-when-necessary).

Irony overhead itself (without next token matching, just shift-reduce) took ~50ms per 1000 parsings, so even if we decrease next token matching to zero, it will still take 40-50 μs per single parsing (20'000 parsings per second per thread, depending o n hw) compared to 3-4μs the current parser. 

I am looking into it. I think we can get up to 10x better result than we get now from XLParser, but that is about the limit of Irony. The comparison is rather unfair, since the original parser can handle only very limited subset of formula and any improvements to the handwritten parser would make numbers more even (+we would get a lot of bugs). I am fine with 10-15x slower results than now for a good parser, because we don't evaluate formulas it on load and I plan to use multi threading for evaluation in the future.

Thoughts?

jahav:
My perf. concerns were based on a small sample and a little overblown.

XLParser has a large collection of excel formulas from Enron (yea, that Enron). Based on that collection, formulas are parsed at 49.97 μs/formula (47.295 s to parse 946 320 formula, Ryzen 5500U, NET6).

I made a PR to XLParser to improve performance, it should be decreased to 28 μs/formula (26.496).

Based on this info, I think it is fine to merge, even if https://github.com/spreadsheetlab/XLParser/issues/161 won't be accepted. We still have to we fix a cache, so it actually works, but that is another topic.

PR #1842, opened 2022-07-25
Retarget ClosedXML to netstandard2.0

igitur:
- Target only netstandard2.0 for ClosedXML.
- Upgrade net46 to net461 for peripheral projects
- Upgrade SourceLink

Required for #1836


igitur:
@jahav Please review and if you're happy, rebase #1836 on this.

igitur:
The upgrade to 461 was necessary to support netstandard20. I also had to add a new explicit reference to System.IO.Packaging.

Directory.Build.props
```diff
@@ -25,6 +25,8 @@
     <RepositoryUrl>https://github.com/ClosedXML/ClosedXML</RepositoryUrl>
     <RepositoryType>git</RepositoryType>
 
+    <!-- Sourcelink: https://github.com/dotnet/sourcelink/ -->
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
```
jahav, line 29: I am not very familiar with making nuget packages. Do we need this for some reason? I looked an it seems that if we set RepositoryUrl (two lines above), it is redundant.

[PublishRepositoryUrl vs RepositoryUrl settings](https://github.com/dotnet/sourcelink/issues/659#issuecomment-711288537)

jahav (approved):

ClosedXML.Examples/ClosedXML.Examples.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;net40;net46</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net461</TargetFrameworks>
```
jafin, line 4: net461 is EOL (Apr 26, 2022), would it be preferable to target supported frameworks, being `net462` ?

ClosedXML.Sandbox/ClosedXML.Sandbox.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;net40;net46</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net461</TargetFrameworks>
```
jafin, line 4: net461 is EOL (Apr 26, 2022), would it be preferable to target supported frameworks, being net462 ?

ClosedXML.Examples/ClosedXML.Examples.csproj
igitur: @jafin These are internal projects, not libraries released to the public. I'd prefer to set it to the lower possible target in order to ensure everything works, tests pass, etc. Even though net461 is EOL, it's not as if Microsoft is going to come to users' PCs and physically delete the framework.

Directory.Build.props
igitur: At the moment we build a .nupkg and a .snupkg file. I have been uploading only the .nupkg file to NuGet thus far, with the idea of switching soon. I meant to do it for v0.96, but at that time I thought I could upload both, which is false. So the `RepositoryUrl` would affect the .nupkg build and the `PublishRepositoryUrl` would override that in the .snupkg build (not 100% sure, but I think that's true). I think this is minor enough to review later.

ClosedXML.Examples/ClosedXML.Examples.csproj
igitur: @jahav @Pankraty Your opinion on keeping the internal projects on net461? net461 vs net462 is a minor issue to me, so we can increase it if there is enough support. I'd say once my default Visual Studio installation on a new PC doesn't have net461 anymore, I'd be inclined to increase it. I wouldn't want to install custom frameworks just to get ClosedXML to build.

jahav: I would go for the lowest version, i.e. net461. It is most likely to show any problems, since it is earliest one. I agree with you.

LeaFrock: > It is most likely to show any problems

If `net461` is, so is `net462` too. 

> I wouldn't want to install custom frameworks just to get ClosedXML to build.

I want to vote for `net462` but this is the point which convinces me finally. Currently keeping `net461` is at least acceptable 🤣 .

Issue #1843, opened 2022-07-26
Duplicate key error when copying sheets from different workbooks


Gaurav-Puri101187:
**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.95.3

**What is the current behavior?**

I have 2 excel files with some custom formatting
ExcelTarget has some number custom formats as:
**<numFmts count="5"><numFmt numFmtId="164" formatCode="_(* #,##0_);_(* \(#,##0\);_(* &quot;-&quot;??_);_(@_)"/><numFmt numFmtId="165" formatCode="#,##0.0"/><numFmt numFmtId="167" formatCode="0.0%"/><numFmt numFmtId="168" formatCode="0.000"/><numFmt numFmtId="169" formatCode="0.0"/>**

Excel Source has number custom format as 
**<numFmts count="5"><numFmt numFmtId="164" formatCode="#,##0.0"/><numFmt numFmtId="165" formatCode="0.0%"/><numFmt numFmtId="166" formatCode="0.0"/><numFmt numFmtId="167" formatCode="0.000"/><numFmt numFmtId="169" formatCode="dd\-mm\-yyyy"/></numFmts>**

Now after copying some sheets from source to destination we get Duplicate key 169 can't be inserted error while saving.

I went through the source code and i believe XLWorkBook_Save.cs -> ResolveNumberFormats method doesn't handle the gaps between the style ids. 

**What is the expected behavior or new feature?**

File should be saved successfully.

**Is this a regression from the previous version?**

No.

## Reproducibility
Mentioned above.

```

- [ ] I attached a sample spreadsheet.  (You can drag files on to this issue)

jahav:
Duplicate of #1679

Issue #1679, opened 2021-06-15
Style.NumberFormat.Format - throws {"An item with the same key has already been added. Key: 171"}


Pitterling:
**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**
0.95.4

**What is the current behavior?**
 
range.Style.NumberFormat.Format = "#,##0.00;-#,##0.00;0;@";    // <-- causing Exception
range.Style.NumberFormat.NumberFormatId = 40;                       // <-- works

```console
System.ArgumentException
  HResult=0x80070057
  Message=An item with the same key has already been added. Key: 171
  Source=System.Private.CoreLib
  StackTrace:
   at System.ThrowHelper.ThrowAddingDuplicateWithKeyArgumentException[T](T key)
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at ClosedXML.Excel.XLWorkbook.GenerateWorkbookStylesPartContent(WorkbookStylesPart workbookStylesPart, SaveContext context)
   at ClosedXML.Excel.XLWorkbook.CreateParts(SpreadsheetDocument document, SaveOptions options)
   at ClosedXML.Excel.XLWorkbook.CreatePackage(String filePath, SpreadsheetDocumentType spreadsheetDocumentType, SaveOptions options)
   at ClosedXML.Excel.XLWorkbook.SaveAs(String file, SaveOptions options)
   at ClosedXML.Excel.XLWorkbook.SaveAs(String file, Boolean validate, Boolean evaluateFormulae)
   at ClosedXML.Excel.XLWorkbook.SaveAs(String file)
   at ClosedXML_FormatBug.Program.Main(String[] args) in C:\xxx\ClosedXML-FormatBug\Program.cs:line 25
```

**What is the expected behavior or new feature?**

No Exception.

**Is this a regression from the previous version?**

Regressions get higher priority. Test against the latest build of the previous minor version. For example, if you experience a problem on v0.95.3, check whether it the problem occurred in v0.94.2 too. 

## Reproducibility
**Code to reproduce problem:**
```c#
public void Main()
{
[...]

PR #1845, opened 2022-07-26
Issue #1843 Cannot save excel workbook due to duplicate keys in styles.

Gaurav-Puri101187:


jahav:
This was a PR for issue #1843 that by was a duplicate of #1679. 

Nonsequential format ids were resolved in PR #1940 and thus it is not needed. Closing.

Issue #1676, opened 2021-06-09
Cell.Value = "Infinity"


ddrogahn-venga:
**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [ ] Yes
- [x] No

**Version of ClosedXML**
0.95.4.0 (nuget.org)

**What is the current behavior?**
Cell().Value = seems to parse "Infinity" and "-Infinity" as a number, but opening the file in Excel gives a warning:
"We found a problem with some content ... do you want to try to recover?"
"Repaired Records: Cell information from /xl/worksheets/sheet1.xml part"

https://docs.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheet.cellvalues?view=openxml-2.8.1#DocumentFormat_OpenXml_Spreadsheet_CellValues_String
Specifically the CellType/Datatype "t" is set to "n" == Number
   <x:c r="A1" s="0" t="n">
    <x:v>Infinity</x:v>
   </x:c>

"NaN" does not trigger this problem.
The string for infinity may change with the culture? unless InvariantCulture is used.
https://stackoverflow.com/questions/18398690/how-should-i-parse-infinity-with-double
(I can't seem to code Decimal.Parse to take "Infinity" or "∞"...)

**What is the expected behavior or new feature?**
Excel doesn't have a representation for infinity: =-1*LOG10(0) returns #NUM!   =1/0 returns #DIV/0!
So "Infinity", "-Infinity" and localizations should be interpreted as a string.

**Is this a regression from the previous version?**
No

## Reproducibility

**Code to reproduce problem:**
```c#
public void Main()
{
	using (var wb = new XLWorkbook()) {
		IXLWorksheet ws = wb.Worksheets.Add("Sheet 1");
		ws.Cell(1, 1).Value = "Infinity";
		wb.SaveAs("Infinity.xlsx");
	}
}
```

- [x] I attached a sample spreadsheet.  (You can drag files on to this issue)
attached
[Infinity.xlsx](https://github.com/ClosedXML/ClosedXML/files/6626502/Infinity.xlsx)

igitur:
That's a fairly rare case. But why don't you attempt a PR to fix this?
Thanks

sylvaincaillot:
Same problem, I am just trying to store "Infinity" which is the name of a product in my Excel file and it gives me an error "We found a problem with some content ... do you want to try to recover?"

[...]

PR #1847, opened 2022-07-29
Add specific handling for cell value of "Infinity".

everythingajw:
I was recently the victim of issue #1676, so I patched it by adding special handling for `Infinity` and `-Infinity`. The underlying issue results from .NET's `double` type parsing `Infinity` and `-Infinity` into actual `double` values (of `double.PositiveInfinity` and `double.NegativeInfinity` respectively). This is fine for most use cases where values of positive and negative infinity are acceptable, but not for Excel since Excel has no representation of infinity. ClosedXML writes these values as numeric types to Excel, but since Excel lacks a representation for infinity, this causes Excel to think the workbook is corrupted. I'm not 100% sure, but I imagine this is because Excel tries to read the string `Infinity` as a number (since ClosedXML marked the cell's data type as numeric), but it cannot do so since `Infinity` is a string of characters.

In my tests, Excel will see this issue and fix it if you ask it to do so (when it prompts for repair, select "yes"). However, this isn't suitable for a production environment where workbooks are being prepared programmatically and delivered to customers. Handling these cases specifically will generate a workbook without any prompts for repair.

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -62,6 +62,11 @@ namespace ClosedXML.Excel
 
         private static readonly Regex utfPattern = new Regex(@"(?<!_x005F)_x(?!005F)([0-9A-F]{4})_", RegexOptions.Compiled);
 
+        /// <summary>
+        /// Collection of values that should not be parsed to a number.
+        /// </summary>
+        private static readonly string[] DisallowedDoubleValues = new string[] { "NaN", "Infinity", "-Infinity" };
+
         #region Fields
 
         private string _cellValue = String.Empty;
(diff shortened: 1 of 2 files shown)


igitur:
Please add unit tests.

everythingajw:
I added unit tests for my changes.

I ran the unit tests, and `MiscTests.CellValues` and `MiscTests.DataTypes` both fail, even if I remove my changes. I fought with OLE DB, but I couldn't make it work for whatever reason, so I couldn't run the OLE DB tests. Other than that, everything passes.

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -305,6 +305,8 @@ private string DeduceCellValueByParsing(string value, XLStyleValue style)
                 this.Style.SetIncludeQuotePrefix();
             }
             else if (!string.Equals(value.Trim(), "NaN", StringComparison.OrdinalIgnoreCase) &&
+                     !string.Equals(value.Trim(), "Infinity", StringComparison.OrdinalIgnoreCase) &&
```
igitur: I think the list of 'disallowed' strings for parsing to `Double` should stored in a local static array.

igitur:
What happens in the case of `cell.SetValue(Double.PositiveInfinity)`? Does ClosedXML accept it and is the resulting file valid?

ClosedXML/Excel/Cells/XLCell.cs
everythingajw: Very good point. I was thinking the same thing and was going to bring it up. I'll get on that.

everythingajw:
D'oh! Can't believe I didn't think of checking that. Yes, ClosedXML accepts it just fine and Excel doesn't complain at all. `double.PositiveInfinity` gets represented as `∞` and `double.NegativeInfinity` gets represented as `-∞`. In both cases, the cell data type is `XLDataType.Text`. 

I see there's a unit test relating to positive infinity; it uses `1.0 / 0.0` instead of `double.PositiveInfinity`. I think tests explicitly using `double.PositiveInfinity` and `double.NegativeInfinity` should be added so we're using the actual values instead of relying on an arithmetic rule. Plus it's more self-documenting - it's more obvious what `double.PositiveInfinity` is compared to `1.0 / 0.0` (particularly since dividing by zero in the math world is undefined). Thoughts?

jahav:
@everythingajw Hi, I am the new maintainer. I am thankful that you took the time to investigate the issue and submit the PR, but the whole situation with storing a value as a string and a type is just untenable. It is completely unclear semantic for users of the library and isn't not how the the Excel works (=a lot of problems in other areas).

I am working on #1894 (builds on #1848 and #1899) to represent the values as a type and there won't be automatic conversion that causes all these problems anymore (as well as a lot of other problems). That is the single thing I want to have in the next version (other are optional). No conversions (unless such a method will be used) = no Infinity or NaN (though it will throw for `double.NaN` or infinity values) or screwed DateTime parsing.

I won't be merging this PR, sorry about that. It improves on the situation that is just wrong and will be untenable and will be fixed soon.

Listbd:
Thank you @jahav for addressing the underlying issue. Looking forward to your fix.
And thank you so much @everythingajw for trying to fix this for us in the meantime and helping me find the work-around.

PR #1860, opened 2022-09-12
Pass null for worksheet when instantiating XLColumns/XLRows from inside of ColumnsUsed/RowsUsed

### LINKED ISSUE #1859: Worksheet's ColumnWidth/RowHeight are reassigned when ColumnsUsed/RowsUsed are changed

- [x] Bug
- [ ] Feature
- [ ] Question

- [x] Yes
- [ ] No

**Version of ClosedXML**

`develop`

If we change `Width` for columns returned by `IXLWorksheet.ColumnsUsed()` the worksheet property `ColumnWidth` is also affected.
Same happens with the `RowHeight` property when we change `Height` of `IXLWorksheet.RowsUsed()`.

* `ColumnWidth` and `RowHeight` remain unchanged in the described scenario
* They should be chaged though when `IXLWorksheet.Columns().Width` or `IXLWorksheet.Rows().Height` are updated (work OK).

Does not seem so.

## Reproducibility

```c#
public void Main()
{
            var wb = new XLWorkbook();
            var ws = wb.AddWorksheet();

            ws.Cells("C3:E5").Value = "Test";

            ws.ColumnsUsed().Width = 30;
            ws.RowsUsed().Height = 30;

            wb.SaveAs("Actual.xlsx");
}
```

**Expected:**
[Expected.xlsx](https://github.com/ClosedXML/ClosedXML/files/9547774/Expected.xlsx)
image

**Actual:**
[Actual.xlsx](https://github.com/ClosedXML/ClosedXML/files/9547773/Actual.xlsx)
image

A PR is coming in a minute.

Pankraty:
Fixes #1859

The documentation to `XLRows` constuctor clearly states: "If worksheet is specified it means that the created instance represents all rows on a worksheet so changing its height will affect all rows" (https://github.com/ClosedXML/ClosedXML/blob/develop/ClosedXML/Excel/Rows/XLRows.cs#L23). Same for [`XLColumns`](https://github.com/ClosedXML/ClosedXML/blob/develop/ClosedXML/Excel/Columns/XLColumns.cs#L21). 

In `ColumnsUsed`/`RowsUsed`, the worksheet argument was initialized with the current worksheet which caused the described behavior.

### DIFF
--- a/ClosedXML.Tests/Excel/Columns/ColumnTests.cs
+++ b/ClosedXML.Tests/Excel/Columns/ColumnTests.cs
@@ -247,5 +247,45 @@ namespace ClosedXML.Tests.Excel
 
             Assert.IsFalse(column.RangeAddress.IsValid);
         }
+
+        [Test]
+        public void AssignWorksheetColumnWidthWhenAllColumnsChanged()
+        {
+            var ws = new XLWorkbook().AddWorksheet();
+            var columns = ws.Columns();
+
+            columns.Width = 100;
+
+            Assert.AreEqual(100, ws.Column("G").Width, XLHelper.Epsilon);
+            Assert.AreEqual(100, ws.ColumnWidth, XLHelper.Epsilon);
+        }
+
+        [Test]
+        public void PreserveWorksheetColumnWidthWhenNotAllColumnsChanged()
+        {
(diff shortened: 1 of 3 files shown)


jahav (approved):
Looks great. Thank you.

PR #1866, opened 2022-09-27
Remove System.Drawing.Common dependency

### LINKED ISSUE #1805: Remove System.Drawing.Common dependency

jahav:
A PR to replace System.Common.Drawing dependency from ClosedXML.

Default graphic engine is rather simplified, but should work for most simple cases.
* Should read dimensions of jpg, png, emf pictures 
* Adjust-to-content that roughly works, but since it doesn't use actual font, it is inherently inaccurate

Documentation is at https://github.com/ClosedXML/ClosedXML/wiki/Graphic-Engine.

~~TODO~~
* ~~Finish SixLabors engine~~
* ~~Tests + coverage.~~
* ~~Check color parser, what are actual allowed values per spec~~
* ~~Think about disposing. I pretty much avoided it, since it is pointless for SixLabors (fonts stored in static collection, never released, default engine is few kb)~~

I probably won't add a graphic engine for System.Drawing.Common. Since we stopped supporting net40, everywone can use netstandard2.0.

Will resolve #1805.


igitur:
What do you think about making the default engine even simpler. Store a map of character codes to a width/height tuple (for a default font size (11pt?)). We could generate the data manually from one of the common fonts, Calibri? Calibri 11pt is the default font & size in a new instance of my Excel 2019 workbook. This default engine should be smaller in size than the current embedded NotoSansDisplay.fon.

jahav:
We can't use Calibri or pretty much any other Windows font because it is a copyrighted with restricted rights (and quite frankly most of the other others).

> Apart from the document embedding rights described previously, you may not redistribute the Windows fonts. You may not copy them to other computers or servers, and you may not convert them to other formats, including bitmap formats, or modify them.
[MS fonts faq: Redistribution and extended rights](https://learn.microsoft.com/en-us/typography/fonts/font-faq#redistribution-and-extended-rights)

The NotoSansDisplay.fon file is not a standard format, it only contains minimal required layout information in font design units - I made the file by loading the font and writing out basic metric and width for codepoints = it is already simplified. I don't want to implement font loading, it is rather complex and the reason for XILGraphicEngine  interface. 2 bytes per codepoint and it has about 6k characters for various alphabets (no chinese characters, but herbew and quite a few others). We could dump everything other than ASCII and it would be a smaller one, but make it a smaller (few KiB), It doesn't really work well with characters from languages that require ore complex layout processing (e.g. Arabic, see StyleFont.xlsx test, that is the worst one).

I originally wanted to include a layout information from Calibri as you suggest, but font licenses are not very friendly for transformative work (see the _convert them to other formats_ above) and are rather limited even for redistribution. Open Font License seems to be the only option.

igitur:
Surely storing width and height metrics doesn't fall under `convert them to other formats`? That seems more about converting to bitmaps. Or should we rather just avoid any gray area?

jahav:
Actual licensing of windows font is pretty complicated (even getting license is annoying), but we definitely shouldn't go this way. Most licenses use a term "derivative work" (it probably has a more well-defined legal judicature) and taking a layout info from a font would definitely qualify as a derivate work, it wouldn't exist without the original.

jahav:
@igitur I checked out the TTF file format yesterday and it is not so complicated, if I only want to read single font (=I know what features it supports). I will try to do it later today. The end result will be two fonts for default engine (will have to declare public), Calibri and NotoSansDisplay. The default will be Calibri reading one from %SystemRoot%/Fonts.

Do you think it is better to use a fallback Calibri -> NotoSansDisplay or throw an exception font not found, install it/use the NotoSans one? I think exception is better.

jahav:
Ok, loading of TTF layout in 250 lines :). Obviously very limited, good enough for Calibri. The test files were fixed because of
* character replacement that actual font does through GSUB table (replaces a glyph for a slash with a glyph for a fraction glyph U+2044)
* bold text, obviously bold text is slightly wider than regular
* slight height changes of AdjustToContent of row, but that is pretty broken anyway. The code provides a reasonably close number, so good enough.

I spent a massive amount of time trying to figure out how the AdjustToContent and widths/height in general work and I still don't know. Spec is incorrect, rounding doesn't make sense and it is all around strange. [The Raster Tragedy at Low Resolution](http://rastertragedy.com/RTRCh4.htm#Sec21) was enlightening, but not a solution. I found a lot of information on the interwebz about it, nearly all wrong.

igitur:
> I spent a massive amount of time trying to figure out how the AdjustToContent and widths/height in general work and I still don't know. Spec is incorrect, rounding doesn't make sense and it is all around strange. [The Raster Tragedy at Low Resolution](http://rastertragedy.com/RTRCh4.htm#Sec21) was enlightening, but not a solution. I found a lot of information on the interwebz about it, nearly all wrong.

Agree, I also spent lots of time on this, including peeking at the LO codebase. Last time I got stuck at https://polymathprogrammer.com/2010/01/18/calculating-column-widths-in-excel-open-xml/ but didn't get round to playing with that.

igitur:
Is your idea that eventually non-naive implementations become separate NuGet packages that users can optionally install and use? That would make sense to me.

jahav:
@igitur I would like to keep ClosedXml reasonable dependency free, but in this case there is also a significant problem: no good dependency
* System drawing is broken on linux
* Skia is a native binary (no blazor for sure), would require multi targeting
* SixLabors - preview package of a group that changed it's license several times ( apache to AGPL to current "split semi-commercial" license) in last 4 years

With no good dependency, pluggable component with a basic in-house engine that is probably the best.

elitastic:
We have just updated to the latest version and successfully implemented our own GraphicEngine so that we can use SkiaSharp  for measuring text width. It works great, thank you all!

Now we're able to use "Calibri" (by manually loading a custom typeface) on our alpine linux containers.

igitur:
@elitastic Fantastic, are you considering open-sourcing your implementation?

elitastic:
@igitur, yes of course:

Some notes: 
- Add the font resources (Calibri ttfs) gzipped as embedded resources and adjust path in `LoadTypeface(...)` method.
- I've only implemented `GetTextWidth()` method, as I only use the GraphicEngine for `worksheet.Columns().AdjustToContents();`

Startup.cs
```
public async Task ConfigureServicesAsync(IServiceCollection services)
{
    XLWorkbook.DefaultStyle.Font.FontName = "Calibri";
    LoadOptions.DefaultGraphicEngine = new SkiaSharpXLGraphicEngine();
}

public void Configure(IApplicationBuilder app, IWebHostEnvironment env, IHostApplicationLifetime applicationLifetime, IServiceProvider serviceProvider, ILoggerFactory loggerFactory)
{
    // Dispose cached typefaces
    applicationLifetime.ApplicationStopping.Register(() =>
    {
        SkiaSharpXmlGraphicEngine.DisposeTypefaceCache();
    });
}
```

SkiaSharpXLGraphicEngine.cs
```
using System;
using System.Collections.Concurrent;
using System.IO;
using System.IO.Compression;
using ClosedXML.Excel;
using ClosedXML.Graphics;
using SkiaSharp;

namespace Namespace;

public class SkiaSharpXLGraphicEngine : IXLGraphicEngine
{
    private static readonly ConcurrentDictionary<string, SKTypeface> SkTypefaces = new ConcurrentDictionary<string, SKTypeface>();

    public XLPictureInfo GetPictureInfo(Stream imageStream, XLPictureFormat expectedFormat)
    {
        throw new NotImplementedException("Get Picture Info");
    }

    public double GetTextHeight(IXLFontBase font, double dpiY)
    {
        throw new NotImplementedException("Get Text Height");
    }

    public double GetTextWidth(string text, IXLFontBase font, double dpiX)
    {
        var fontName = font.Bold ? "Calibri-Bold.ttf.gz" : "Calibri-Regular.ttf.gz";
        var typeface = LoadTypeface(fontName);

        using (var paint = new SKPaint(new SKFont(typeface, (float)font.FontSize)))
        {
            var width = paint.MeasureText(text);

            return PointsToPixels(width, dpiX);
        }
    }

    private static SKTypeface LoadTypeface(string fontName)
    {
        return SkTypefaces.GetOrAdd(fontName, name =>
        {
            using (var resource = typeof(SkiaSharpXLGraphicEngine).Assembly.GetManifestResourceStream($"[ENTER_EMBEDDED_RESOURCE_PATH]{name}"))
            {
                using (var fontStream = new GZipStream(resource, CompressionMode.Decompress))
                {
                    var typeFace = SKTypeface.FromStream(fontStream);
                    if (typeFace == null)
                    {
                        throw new InvalidOperationException($"Could not load font: {name}");
                    }

                    return typeFace;
                }
            }
        });
    }

    public double GetMaxDigitWidth(IXLFontBase font, double dpiX)
    {
        throw new NotImplementedException("Get Max Digit Width");
    }

    public double GetDescent(IXLFontBase font, double dpiY)
    {
        throw new NotImplementedException("Get Descent");
    }

    // From https://github.com/ClosedXML/ClosedXML/blob/develop/ClosedXML/Graphics/DefaultGraphicEngine.cs
    private static double PointsToPixels(double points, double dpi) => points / 72d * dpi;

    public static void DisposeTypefaceCache()
    {
        foreach (var typeface in SkTypefaces.Values)
        {
            typeface.Dispose();
        }

        SkTypefaces.Clear();
    }
}
```

Issue #1886, opened 2022-10-25
Error when evaluating VALUE function on an empty cell


cjundt:
## Read and complete the full issue template

Do not randomly delete sections. They are here for a reason.

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.97.0

**What is the current behavior?**

Throws System.FormatException: Input string was not in a correct format.

**What is the expected behavior or new feature?**

Function should return 0 (zero) in case of empty cell as in Excel.

**Is this a regression from the previous version?**

No

## Reproducibility

**Code to reproduce problem:**
```c#
using ClosedXML.Excel;

public void Main()
{
	using( var wb = new XLWorkbook( ) ) {
		var ws = wb.Worksheets.Add("Sample Sheet");
		ws.Cell( "A1" ).Value = "";
		ws.Cell( "A10" ).FormulaA1 = "=VALUE(A1)";

		Console.WriteLine( $"A10={ws.Cell( "A10" ).Value}" );
	}
}
```

- [ ] I attached a sample spreadsheet.  (You can drag files on to this issue)

PS : Pull request will follow

PR #1889, opened 2022-10-25
#1886 - Error when evaluating VALUE function on an empty cell

cjundt:
Fixes issue #1886


jahav:
Closed in favor of #1895. This function doesn't even parse currency or date.

PR #1907, opened 2022-11-03
Add a read the docs

jahav:
A testing PR to add a read-the-docs to the repo.  
* ClosedXML needs a better documentation, 
* ever PR should include docs about feature/changes => need docs in the repo .
* Github markdown is very limited
* Github wiki doesn't do versioning by release


jahav:
For future reference
* Sphix is a RST/MkDoc to other formats generator (in our case RST to HTML)
* Breathe-doc is a generator that takes XML from Doxygen and converts it to csharp documentation. Breathe is in a "minimal active work" mode ([source](https://github.com/breathe-doc/breathe/issues/862#issuecomment-1284562375))
* `sphinx_csharp` is a domain extension for Breathe-doc to generate csharp elements (classes, methods, structs), e.g. parses method signature and outputs nodes to write to RST.
* The Doxygen generation must be done through Python `subprocess.call` in `conf.py`, otherwise ClosedXML would need to store XML files in repo (not acceptable). See [Running on Read the Docs - Generating Doxygen XML Files](https://breathe.readthedocs.io/en/latest/readthedocs.html)
* Breathe-doc added support for C# in this PR https://github.com/breathe-doc/breathe/pull/550
* Breathe-doc uses an extensions to generate the markup from Doxygen XML. Extension must be registered in the Sphinx.
* There are several extensions, a [dead one](https://github.com/djungelorm/sphinx-csharp) (has a obsolete package on PyPI that throws errors), [maintained fork](https://github.com/rogerbarton/sphinx-csharp) that doesn't have a package on PyPI, [a not-so-pretty one](https://github.com/madTeddy/sphinxsharp) where I don't know how well it is integrated to Breathe.
* Breathe `::member::` option shows even public methods from non-public methods. That kind of make sense, why should non-public class have public methods, but not something we want to show. EXTRACT_PRIVATE in Doxygen should have stopped it, but doesn't... not really a problem.

conf.py
```python
# Need to have breathe to generate xml to rst and sphinx_csharp (note the underscore) 
# to generate pretty Doxygen documentation
extensions = ['sphinx_csharp','breathe']

import subprocess, os

read_the_docs_build = os.environ.get('READTHEDOCS', None) == 'True'

if read_the_docs_build:
     subprocess.call('doxygen', shell=True)
     subprocess.call('pip install git+https://github.com/rogerbarton/sphinx-csharp.git', shell=True)    
```

PR #1938, opened 2022-12-18
Reimplement VLOOKUP using bisection.

### LINKED ISSUE #1763: VLOOKUP doesn't work correctly

jahav:
Bisection should be significantly faster and VLOOKUP is one of the most used function. It is now also more compliant with the excel behavior.

This PR together with #1933 resolves #1763.

### DIFF
--- a/ClosedXML.Tests/Excel/CalcEngine/LookupTests.cs
+++ b/ClosedXML.Tests/Excel/CalcEngine/LookupTests.cs
@@ -1,9 +1,9 @@
 // Keep this file CodeMaid organised and cleaned
 using ClosedXML.Excel;
 using ClosedXML.Excel.CalcEngine;
-using ClosedXML.Excel.CalcEngine.Exceptions;
 using NUnit.Framework;
 using System;
+using System.Linq;
 
 namespace ClosedXML.Tests.Excel.CalcEngine
 {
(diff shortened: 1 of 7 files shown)


jahav:
Two lessons from this PR:
1. Don't copy text from spec PDF to the code. If it contains unusual characters (e.g. eclippse or en dash), ReadTheDocs generation can fail and it is pretty hard to find exact place.

2. VLOOKUP approximate search is a bisection with a twist:
* Ignore different type
* Don't find a value, but "if an exact match is not found, the next largest value that is less than lookup-value is returned"
* If there are multiple values, return the one with the highest row number

PR #1944, opened 2022-12-22
fix #1896 "Pictures and Shapes display order is not respected after save"

### LINKED ISSUE #1896: Pictures' and Shapes' display order is not respected after save
- [x] Bug
- [ ] Feature
- [ ] Question

- [X] Yes
- [ ] No

**Version of ClosedXML**

0.97.0

When saving a workbook containing stacked pictures and shapes, the assets display order is scrambled. Pictures always get in front of shapes.

Images and shapes should keep the disposition they had in opened file.

This is not a regression. Bug was present in 0.92.0 already.

## Reproducibility

```c#
public void Main()
{
    // Code standards:
    // - Fully runnable. I should be able to copy and paste this code into a 
    //   console application and run it without having to edit it much.
    // - Declare all your variables (this follows from the previous point)
    // - The code should be a minimal code sample to illustrate issue. The code 
    //   samples on the wiki are good examples of the terseness that I want. Don't
    //   post your full application.
	var filePath = "C:\Tests\Images-Shapes.xlsx"
	var workbook = new XLWorkbook(filePath);
	workbook.SaveAs(filePath + ".test.xlsx");
}
```

I found a possible cause and fix for this issue in ClosedXML.Excel.XLWorkbook.AddPictureAnchor method.
This comes from the position the new OpenXml elements describing the picture are inserted in WorksheetDrawing part.
Since they are appended, they seem to be drawn last. Because other types of graphics (i.e. shapes) are ignored, these end-up being sent to back.
By keeping track of the original part's ordering, this issue can be resolved.

[Images-Shapes.xlsx](https://github.com/ClosedXML/ClosedXML/files/9888765/Images-Shapes.xlsx)

nakamura2000:
When saving picture original code deletes current anchor of image and adds new anchor of image.
This results in different display order .
I fix to use ReplaceChild function to keep picture position.


ClosedXML.Tests/Excel/ImageHandling/PictureTests.cs
```diff
@@ -470,5 +471,26 @@ public void CanCopyEmfPicture()
...
+            var relId = xlImage1.RelId;
+            int pos1 = OpenXmlHelper.FindImageRelIdDrawingOrder(stream,ws1.Name, relId);
+
+            using var ms = new MemoryStream();
+            wb.SaveAs(ms);
+            stream.Close();
+            int pos2 = OpenXmlHelper.FindImageRelIdDrawingOrder(ms, ws1.Name, relId);
+            ms.Close();
+
+            Assert.AreEqual(pos1, pos2);
```
jahav: Since there isn't an API for Z-order, it is better to just compare files. In this case, we are basically checking cycle load-save and file comparison is a standard solution for that. Test shouldn't deal with xlsx xml structure directly.

ClosedXML/Excel/IO/WorksheetPartWriter.cs
```diff
@@ -1714,7 +1712,14 @@ private static void AddPictureAnchor(WorksheetPart worksheetPart, Drawings.IXLPi
...
 
-                    worksheetDrawing.Append(absoluteAnchor);
+                    if (existingAnchor != null)
+                    {
+                        worksheetDrawing.ReplaceChild(absoluteAnchor, existingAnchor);
+                    }
+                    else
+                    {
+                        worksheetDrawing.Append(absoluteAnchor);
+                    }
```
jahav: Repeated code. Replace with a method.

ClosedXML/Utils/OpenXmlHelper.cs
```diff
@@ -81,6 +82,52 @@ public static XLColor ToClosedXMLColor(this X14.ColorType openXMLColor, IDiction
...
 
+        #region Internal Methods
+        /// <summary>
+        /// Find image position in drawing.
+        /// </summary>
+        /// <param name="stream">elsx stream</param>
+        /// <param name="sheetName">sheet name</param>
+        /// <param name="relId">picture relId</param>
+        /// <returns>position index in drawing part</returns>
+        internal static int FindImageRelIdDrawingOrder(System.IO.Stream stream, string sheetName, string relId)
```
jahav: This method shouldn't be in the ClosedXML project, because it is only used for tests. Tests also shouldn't deal with xml structure directly, unless there is a good reason, see comment near the test.

jahav (approved):
The test file has also been in the Examples directory. That directory is reserved for files created by various examples in the ClosedXML.Tests.Examples namespace (I think they are also used on wiki). Since we can't create file with mixed z-order and there is no example for it, it belongs to other directory.

I want to do a release this weekend, so I just squashed your commits and added changes from the review.

PR #1946, opened 2022-12-31
Renamed XLPivotValues parameters

### LINKED ISSUE #1945: Methods' parameters' names in XLPivotValues

- [x] Bug
- [ ] Feature
- [ ] Question

- [x] Yes
- [ ] No

**Version of ClosedXML**

0.97

Parameters' names in majority of methods (i.e. [Remove](https://github.com/ClosedXML/ClosedXML/blob/a0b0bb25fc63571af27e19ebd391b1106c3ef754/ClosedXML/Excel/PivotTables/PivotValues/XLPivotValues.cs#L87)) suggest passing **SourceName** of XLPivotField when underlying dictionary uses **CustomName** since 5b5efdd8ed1b4d80c8b64952c2102c5c9c8f00d9. Also [IndexOf](https://github.com/ClosedXML/ClosedXML/blob/a0b0bb25fc63571af27e19ebd391b1106c3ef754/ClosedXML/Excel/PivotTables/PivotValues/XLPivotValues.cs#L82) will throw **ArgumentNullException** because it passes **SourceName** as key to mentioned dictionary.

Names match underlying logic.

Yes.

## Reproducibility
```c#
using ClosedXML.Excel;

var workbook = new XLWorkbook("Book1.xlsx");

var pivot = workbook.Worksheets.SelectMany(sheet => sheet.PivotTables).Single(pivot => pivot.Name == "PivotTable1");

Console.WriteLine("Source names:");
Console.WriteLine(string.Join(Environment.NewLine, pivot.Values.Select(value => value.SourceName)));

Console.WriteLine("Custom names:");
Console.WriteLine(string.Join(Environment.NewLine, pivot.Values.Select(value => value.CustomName)));

pivot.Values.Remove("Field3");

// Will throw exception.
var index = pivot.Values.IndexOf(pivot.Values.ElementAt(1));

// Values are the same.
workbook.SaveAs("Book1 - Copy.xlsx");
```

[Book1.xlsx](https://github.com/ClosedXML/ClosedXML/files/10327373/Book1.xlsx)

0MG-DEN:
Fixes #1945

### DIFF
--- a/ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotValueStyleFormat.cs
+++ b/ClosedXML/Excel/PivotTables/PivotStyleFormats/XLPivotValueStyleFormat.cs
@@ -24,7 +24,7 @@ namespace ClosedXML.Excel
 
         public IXLPivotValueStyleFormat ForValueField(IXLPivotValue valueField)
         {
-            FieldReferences.Add(new PivotValueFieldReference(valueField.SourceName));
+            FieldReferences.Add(new PivotValueFieldReference(valueField.CustomName));
             return this;
         }
 
--- a/ClosedXML/Excel/PivotTables/PivotValues/IXLPivotValues.cs
+++ b/ClosedXML/Excel/PivotTables/PivotValues/IXLPivotValues.cs
@@ -12,18 +12,18 @@ namespace ClosedXML.Excel
 
         void Clear();
 
-        Boolean Contains(String sourceName);
+        Boolean Contains(String customName);
 
         Boolean Contains(IXLPivotValue pivotValue);
 
-        IXLPivotValue Get(String sourceName);
+        IXLPivotValue Get(String customName);
 
         IXLPivotValue Get(Int32 index);
 
-        Int32 IndexOf(String sourceName);
+        Int32 IndexOf(String customName);
 
         Int32 IndexOf(IXLPivotValue pivotValue);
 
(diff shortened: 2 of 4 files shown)


jahav (changes requested):
Interface `IXLPivotValues` also needs to have signatures changed.

`PivotValueFieldReference` needs to have a diferent value in ctor. It wuld fail later in some edge cases, because it uses `pt.Values.IndexOf(Value.ToString())` where `Value` is a `SourceName` (at least when I looked through call graph).

The `AbstractPivotFieldReference.Match` method of the is later used for pivot table table styling IIRC, https://github.com/ClosedXML/ClosedXML/wiki/Pivot-Table-Styles#cross-of-values-and-axis-fields

0MG-DEN:
Thanks for the guidance @jahav
Sorry for not making these changes right away

jahav (approved):

jahav:
Thanks for the contribution. Pivot tables will likely get some TLC in 0.102. That `IndexOf` should probably be removed from public API (position randomly changes on modification of dictionary).

igitur:
Pivot tables do need some work, but have also come a long way in a short time, before dev paused.

I'd suggest the next big step is to factor out common pivot table data sources, because multiple pivot tables can reference a single source and changes to a source affect all dependent pivot tables.

Also see related issue at https://github.com/OfficeDev/Open-XML-SDK/issues/612

PR #1950, opened 2023-01-04
Update some dependencies

sbeca:
I was recently trying to track down a bug with a spreadsheet that I have where loading then saving the spreadsheet causes validation issues to be thrown. I was randomly hoping that an update to the OpenXml dependency would help but it didn't, but I thought that I might as well upstream my dependency updates anyway.

I'm just including dependency updates in this PR that should hopefully not include any breaking changes.

For reference, I tried the following dependency updates but they failed:
1. I tried updating `DocumentFormat.OpenXml` to v2.19.0 but that appears to have some breaking changes that caused some image tests in ClosedXML to start failing, so I only updated to v2.18.0 which allows all tests to pass.
2. I tried updating `XLParser` to v1.6.2 but that has a bunch of breaking changes that caused a bunch of tests in ClosedXML to start failing so I left that dependency alone.

### DIFF
--- a/ClosedXML.Examples/ClosedXML.Examples.csproj
+++ b/ClosedXML.Examples/ClosedXML.Examples.csproj
@@ -16,7 +16,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\ClosedXML\ClosedXML.csproj" />
-    <PackageReference Include="DocumentFormat.OpenXml" Version="2.16.0" />
+    <PackageReference Include="DocumentFormat.OpenXml" Version="2.18.0" />
     <PackageReference Include="morelinq" Version="2.10.0" />
   </ItemGroup>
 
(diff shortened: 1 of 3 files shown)


jahav:
I am sorry but I am not going to merge it. 
* Either is is a important dependency (DocumentFormat.OpenXml), where I need to go through changes myself to see what has changed. 2.19 is also a minor change, but something failed. I recently had a bug related to a change that landed in 2.19 ( https://github.com/OfficeDev/Open-XML-SDK/pull/1240) for Unity.
* It is an unstable dependency (SixLabors.Fonts), where anything can change. I don't want to change version, unless it is needed for something or it is out of beta.
* I am pretty sure I don't actually need Microsoft.CSharp and it should be removed.

Basically I don't think my coverage is enough and I need to review changes of dependencies (at least cursory review) and it is not enough that test passed.

igitur:
In the past we were very conservative and would leave the dependencies on their lowest possible version so that we don't force consumers of ClosedXML to upgrade and thereby possibly introducing conflicts or other problems. 

The consumer still has the option to reference a later version of the dependencies directly if they wish to upgrade.

twsouthwick:
@jahav can you elaborate on what the bug you hit in 2.19?

Issue #1917, opened 2022-11-18
NotContains filter turns into Contains filter after reapplying filters


mrboring:
## Read and complete the full issue template

Do not randomly delete sections. They are here for a reason.

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.96.0
0.97.0
0.97.1-preview

**What is the current behavior?**

* `NotContains` filter turns into a `Contains` filter.
* The filter displays correctly until reapplying filters in Excel (Ctrl+Alt+L). This changes the filter to a `Contains` filter hiding the wrong rows.

I filtered the first column to not contain `String3`.

Before reapplying filters it's working. Row 4 is hidden as expected:
![image](https://user-images.githubusercontent.com/5991977/202725067-f8c1e6fd-3dda-428e-99c1-c64acb62ebca.png)

After, it's changed to a contains filter. Row 4 is visible and the other rows are hidden:
![image](https://user-images.githubusercontent.com/5991977/202725230-ed2a2cea-ab9c-40a2-8f5a-a310419eb208.png)

**What is the expected behavior or new feature?**

The filter should have remained the same and the number of filtered rows should appear in the Excel status bar.

**Is this a regression from the previous version?**

Regressions get higher priority. Test against the latest build of the previous minor version. For example, if you experience a problem on v0.95.3, check whether it the problem occurred in v0.94.2 too. 

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

Without a code sample, it is unlikely that your issue will get attention. Don't be lazy. Do the effort and assist the developers to reproduce your problem. Code samples should be [minimal complete and verifiable](https://stackoverflow.com/help/mcve). Sample spreadsheets should be attached whenever applicable. Remove sensitive information.

**Code to reproduce problem:**

Sorry, I'm not a good C# programmer. Here's my F# code:

```F#
#r "nuget: ClosedXML"

open System
open ClosedXML.Excel
let x = new XLWorkbook()

let savePath = sprintf @"%s\temp\Test_DoesNotContain_%s.xlsx"
[...]

PR #1953, opened 2023-01-05
Fixes issue #1917 not contains filter

cjundt:
Added unit tests and code fix in order to handle the "Not" string filters family as expected.


jahav:
Resolves #1917

jahav:
@cjundt I didn't like how the condition has sometimes had to be inverted, so I changed to code so it always returns true when value satisfies the filter.

Be warned that autofilter is going to be revamped, likely in 0.101, I already researched how it actually work and done some work, but it just didn't make it. You have seen it, conditions are duplicated across load code, XLFilteredColumn and XLCustomFilteredColumn, filters don't work correctly (no wildcard support, no correct coercion for comparison filters, equals operator applying filter on filter value, while not equal uses normal textual value and the rest).

PR #1954, opened 2023-01-06
Calculate default column width from default font.

### LINKED ISSUE #837: Consider baseColWidth when reading the XLSX file

- [x] Bug
- [ ] Feature

**Version of ClosedXML**

e.g. 0.93.0-beta2

The issue #829 revealed another problem. Default column width may be specified in two different ways: using `defaultColWidth` and `baseColWidth` (see [documentation](https://msdn.microsoft.com/en-US/library/office/documentformat.openxml.spreadsheet.sheetformatproperties.aspx)). ClosedXML only considers `defaultColWidth` and if column width is specified using `baseColWidth` it is simply ignored. On saving or copying a worksheet this information is lost, and new columns are created with the default width (8.43).

In case `defaultColWidth` is not specified and `baseColWidth` is set we should set `XLWorksheet.ColumnWidth` accordingly (`baseColumnWidth + {margin padding (2 pixels on each side, totalling 4 pixels)} + {gridline (1pixel)}`)

No

## Reproducibility

```c#
public void Main()
{
            using (var wb = new XLWorkbook(@"829_sample.xlsx"))
            {
                var ws = wb.Worksheet(1);
                Console.WriteLine("Expected {0}, actual {1}", 11.17, ws.ColumnWidth); // actual 8.43
                Console.WriteLine("Expected {0}, actual {1}", 11.17, ws.Column(1).Width);  // actual 8.43
            }

            Console.WriteLine("Press any key to continue");
            Console.ReadKey();
}
```

[829_sample.xlsx](https://github.com/ClosedXML/ClosedXML/files/1977595/829_sample.xlsx)

nakamura2000:
Calculate default workbook column width from default font.
when loading existing excel file default column width is fixed to 8.43, but when you change default font of excel settings (ex Arial), default column width changes to different value.

calculation of column width is described here. 
https://learn.microsoft.com/ja-jp/dotnet/api/documentformat.openxml.spreadsheet.column?view=openxml-2.8.1

I also refer to sample code from this blog.
http://www.ericwhite.com/blog/precisely-calculating-cell-sizes-in-open-xml-spreadsheetml/

1. calculate workbook column width from default font.
2. calculate worksheet column width from openxml DefaultColumnWidth and BaseColumnWidth without magic number.

this pr fix #837


jahav:
@nakamura2000 I will at the PR over the weekend, it's not being ignored.

PR #1968, opened 2023-01-12
Change ExpressionCache to use ConditionalWeakTable

lahma:
We see in our CI runs from time to time following exception:

```
Error Message:
 System.NullReferenceException : Object reference not set to an instance of an object.
Stack Trace:
 at ClosedXML.Excel.CalcEngine.CalcEngine.Evaluate(String expression)
 at ClosedXML.Excel.XLWorksheet.Evaluate(String expression)
 at ClosedXML.Excel.XLCell.RecalculateFormula(String fA1)
 at ClosedXML.Excel.XLCell.Evaluate(Boolean force)
 at ClosedXML.Excel.XLCell.get_Value()
```

As our test runs are memory-intensive, my guess would be that there is a race condition with current `WeakReference` handling in `ExpressionCache`. Basically reference's target could be cleaned during the logic. Instead of adding more safeguards, I'm proposing to switch to using `ConditionalWeaktable` as it makes code more clear and less error-prone.

`CalcEngine` referencing the cache has very limited lifetime, scoped to method what I've studied the code so I don't see a reason to have such complex cleaned logic and artificial 100 item threshold either.

### DIFF
--- a/ClosedXML/Excel/CalcEngine/ExpressionCache.cs
+++ b/ClosedXML/Excel/CalcEngine/ExpressionCache.cs
@@ -1,5 +1,4 @@
-using System;
-using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 
 namespace ClosedXML.Excel.CalcEngine
 {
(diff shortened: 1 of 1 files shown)


jahav (approved):
Thanks. ConditionalWeakTable will be better solution for time being.

jahav:
I try to do release every few (two) month, so if you need it now, pick nuget from appveyor. https://github.com/ClosedXML/ClosedXML/wiki/Development-Builds

lahma:
Great thanks for the tip. A bit out of this topic, but would you be interested in having a PR adding [NUKE](https://www.nuke.build/) based parallel build infrastructure that would also build the project on GitHub? I've set it up with multiple projects, [like Quartz.NET](https://github.com/quartznet/quartznet/tree/main/build) and it's been great.

jahav:
@lahma Thanks for the offer, but not at this time. I have taken over after about a year of hiatus with a significant backlog and anything else is an overhead I don't want to spent time on. I know how current build system works if I need something.

lahma:
All good, thanks for working on and improving the library!

PR #1982, opened 2023-01-20
Add a new XLCellValue.FromObject() method

0xced:
Also add support for `float` to  `XLCellValue` conversion.


jahav (approved):

jahav:
Thanks for the contribution. That will be useful if you get object from ADO.NET or some other API without a concrete type..

Issue #1835, opened 2022-07-12
The number in this cell is formatted as text  / addIgnoredErrors() is missing


georgheimltwat:
## Read and complete the full issue template

Do not randomly delete sections. They are here for a reason.

**Do you want to request a *feature* or report a *bug*?**
- [ ] Bug
- [x] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.96.0

**What is the current behavior?**

when opening the attached excel, created with the sample code excel displays the warning "The number in this cell is formatted as text..."

**What is the expected behavior or new feature?**

supporting the built in feature of excel "addIgnoredErrors(...)" would solve the issue

**Is this a regression from the previous version?**

No

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

Without a code sample, it is unlikely that your issue will get attention. Don't be lazy. Do the effort and assist the developers to reproduce your problem. Code samples should be [minimal complete and verifiable](https://stackoverflow.com/help/mcve). Sample spreadsheets should be attached whenever applicable. Remove sensitive information.

**Code to reproduce problem:**
```c#
public void Main()
{
    // Code standards:
    // - Fully runnable. I should be able to copy and paste this code into a 
    //   console application and run it without having to edit it much.
    // - Declare all your variables (this follows from the previous point)
    // - The code should be a minimal code sample to illustrate issue. The code 
    //   samples on the wiki are good examples of the terseness that I want. Don't
    //   post your full application.
    var wb = new ClosedXML.Excel.XLWorkbook();
    var ws = wb.AddWorksheet("Table1");
    ws.Row(1).Cell(1).SetValue("1234");

    wb.SaveAs($"C:/TEMP/TestClosedXml_NumberAsText.xlsx");
}
```

- [x] I attached a sample spreadsheet.  (You can drag files on to this issue)
[TestClosedXml_NumberAsText.xlsx](https://github.com/ClosedXML/ClosedXML/files/9090184/TestClosedXml_NumberAsText.xlsx)

igitur:
>  ws.Row(1).Cell(1).SetValue("1234");

You're passing a string into `SetValue`, which then explicitly sets the cell's data type as string. 

[...]

PR #1987, opened 2023-01-26
added ignored errors

georgheimltwat:
solves https://github.com/ClosedXML/ClosedXML/issues/1835


BenPH:
This would be great if added. @jahav would you be able to take a look?

jahav:
I have little interest in new features while ClosedXML is sinking under the
weight of technical debt from its existing, half working features.

redbearsam:
I'm requiring basically this very same addition. I don't suppose 

a) someone with relevant permissions would be happy to fix the conflicts, or 
b) @georgheimltwat you'd be okay with it if I were to reproduce this work against current develop in a new PR?

georgheimltwat:
> you'd be okay with it if I were to reproduce this work against current develop in a new PR?

Yes, of course

PR #1998, opened 2023-02-01
Improve performance when opening workbook

lahma:
I'm offering here couple performance optimizations that we made to improve loading of large workbooks. I'll add some comments per change.

### DIFF
--- a/ClosedXML/Excel/Cells/XLCells.cs
+++ b/ClosedXML/Excel/Cells/XLCells.cs
@@ -71,6 +71,7 @@ namespace ClosedXML.Excel
 
         private IEnumerable<XLCell> GetUsedCells()
         {
+            var visitedCells = new HashSet<XLAddress>();
             var groupedAddresses = _rangeAddresses.GroupBy(addr => addr.Worksheet);
             foreach (var worksheetGroup in groupedAddresses)
             {
@@ -82,14 +83,13 @@ namespace ClosedXML.Excel
                     .OrderBy(cell => cell.Address.RowNumber)
                     .ThenBy(cell => cell.Address.ColumnNumber);
 
-                var visitedCells = new HashSet<XLAddress>();
+                visitedCells.Clear();
                 foreach (var cell in cells)
                 {
-                    if (visitedCells.Contains(cell.Address)) continue;
-
-                    visitedCells.Add(cell.Address);
-
-                    yield return cell;
+                    if (visitedCells.Add(cell.Address))
+                    {
+                        yield return cell;
+                    }
                 }
             }
         }
@@ -105,12 +105,11 @@ namespace ClosedXML.Excel
             var maxColumn = normalizedAddress.LastAddress.ColumnNumber;
 
             var cellRange = worksheet.Internals.CellsCollection
-                .GetCells(minRow, minColumn, maxRow, maxColumn, _predicate)
-                .Where(c => !c.IsEmpty(_options));
+                .GetCells(minRow, minColumn, maxRow, maxColumn, _predicate);
 
             foreach (var cell in cellRange)
             {
-                if (_predicate(cell))
+                if (!cell.IsEmpty(_options) && _predicate(cell))
                     yield return cell;
             }
 
@@ -131,6 +130,11 @@ namespace ClosedXML.Excel
         {
             var candidates = Enumerable.Empty<XLSheetPoint>();
 
+            if (_options == XLCellsUsedOptions.AllContents)
+            {
+                return candidates;
+            }
+
             if (_options.HasFlag(XLCellsUsedOptions.MergedRanges))
                 candidates = candidates.Union(
                     worksheet.Internals.MergedRanges.SelectMany(r => GetAllCellsInRange(r.RangeAddress)));
@@ -148,11 +152,12 @@ namespace ClosedXML.Excel
 
         public IEnumerator<XLCell> GetEnumerator()
         {
-            var cells = (_usedCellsOnly) ? GetUsedCells() : GetAllCells();
-            foreach (var cell in cells)
-            {
-                yield return cell;
-            }
+            return GetCells().GetEnumerator();
+        }
+
+        private IEnumerable<XLCell> GetCells()
+        {
+            return _usedCellsOnly ? GetUsedCells() : GetAllCells();
         }
 
         #endregion IEnumerable<XLCell> Members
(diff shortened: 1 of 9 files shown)


ClosedXML/Excel/Cells/XLCells.cs
```diff
@@ -71,6 +71,22 @@ private IEnumerable<XLSheetPoint> GetAllCellsInRange(IXLRangeAddress rangeAddres
 
         private IEnumerable<XLCell> GetUsedCells()
         {
+            if (_rangeAddresses.Count == 1)
```
lahma, line 74: no need to do expensive calculations in this case

ClosedXML/Excel/Cells/XLCells.cs
```diff
@@ -71,6 +71,22 @@ private IEnumerable<XLSheetPoint> GetAllCellsInRange(IXLRangeAddress rangeAddres
...
+                return GetUsedCellsInRange(_rangeAddresses[0], ws, usedCellsCandidates);
+            }
+
+            // need to do a bit more complex aggregation
+            return GetUsedCellsByWorksheetGroup();
+        }
+
+        private IEnumerable<XLCell> GetUsedCellsByWorksheetGroup()
+        {
+            var visitedCells = new HashSet<XLAddress>();
```
lahma, line 74: reusing same HashSet instead of always creating a new one in loop, there were a lot of them created even when one can be used if cleared between iterations

ClosedXML/Excel/Cells/XLCells.cs
```diff
@@ -105,12 +120,11 @@ private IEnumerable<XLCell> GetUsedCellsInRange(XLRangeAddress rangeAddress, XLW
             var maxColumn = normalizedAddress.LastAddress.ColumnNumber;
 
             var cellRange = worksheet.Internals.CellsCollection
-                .GetCells(minRow, minColumn, maxRow, maxColumn, _predicate)
-                .Where(c => !c.IsEmpty(_options));
```
lahma, line 109: this was a capturing closure (memory allocations), faster to do inside next loop as extra check

ClosedXML/Excel/Cells/XLCells.cs
```diff
@@ -131,6 +145,11 @@ private IEnumerable<XLSheetPoint> GetUsedCellsCandidates(XLWorksheet worksheet)
         {
             var candidates = Enumerable.Empty<XLSheetPoint>();
 
+            if (_options == XLCellsUsedOptions.AllContents)
```
lahma, line 133: next steps won't be required as AllContents doesn't match

ClosedXML/Excel/Cells/XLCells.cs
```diff
@@ -148,11 +167,12 @@ private IEnumerable<XLSheetPoint> GetUsedCellsCandidates(XLWorksheet worksheet)
 
         public IEnumerator<XLCell> GetEnumerator()
         {
-            var cells = (_usedCellsOnly) ? GetUsedCells() : GetAllCells();
-            foreach (var cell in cells)
-            {
-                yield return cell;
-            }
+            return GetCells().GetEnumerator();
```
lahma, line 155: reusing single implementation that doesn't require enumerator instances wrapping enumerators

ClosedXML/Excel/Ranges/XLRangeBase.cs
```diff
@@ -563,7 +563,14 @@ public virtual Boolean IsEmpty()
 
         public virtual Boolean IsEmpty(XLCellsUsedOptions options)
         {
-            return CellsUsed(options).Cast<XLCell>().All(c => c.IsEmpty(options));
```
lahma, line 566: capturing closure here too, was allocating a lot

ClosedXML/Excel/Rows/XLRowCollection.cs
```diff
@@ -20,9 +20,7 @@ public void Add(int key, XLRow value)
         {
             if (key > MaxRowUsed) MaxRowUsed = key;
 
-            if (Deleted.ContainsKey(key))
-                Deleted.Remove(key);
-
+             Deleted.Remove(key);
```
lahma, line 23: Remove can be called regardless whether dictionary contains something, no need for double lookups

ClosedXML/Excel/Style/IXLAlignment.cs
```diff
@@ -2,14 +2,14 @@
 
 namespace ClosedXML.Excel
 {
-    public enum XLAlignmentReadingOrderValues
+    public enum XLAlignmentReadingOrderValues : byte
```
lahma, line 5: field alignment uses less memory for small enums

ClosedXML/Excel/Style/XLStylizedBase.cs
```diff
@@ -104,10 +104,21 @@ void IXLStylized.ModifyStyle(Func<XLStyleKey, XLStyleKey> modification)
             }
         }
 
-        private IEnumerable<XLStylizedBase> GetChildrenRecursively(XLStylizedBase parent)
+        private static HashSet<XLStylizedBase> GetChildrenRecursively(XLStylizedBase parent)
         {
-            return new List<XLStylizedBase> { parent }
-                   .Union(parent.Children.Where(child => child != parent).SelectMany(child => GetChildrenRecursively(child)));
+            void Collect(XLStylizedBase root, HashSet<XLStylizedBase> collector)
```
lahma, line 109: faster without LINQ, returning concrete type so no interface dispatch etc required with result

ClosedXML/Excel/XLWorksheet.cs
```diff
@@ -934,10 +934,12 @@ public IXLRows RowsUsed(XLCellsUsedOptions options = XLCellsUsedOptions.AllConte
         {
             var rows = new XLRows(worksheet: null, StyleValue);
             var rowsUsed = new HashSet<Int32>();
-            Internals.RowsCollection.Keys.ForEach(r => rowsUsed.Add(r));
-            Internals.CellsCollection.RowsUsed.Keys.ForEach(r => rowsUsed.Add(r));
-            foreach (var rowNum in rowsUsed)
+            foreach (var rowNum in Internals.RowsCollection.Keys.Concat(Internals.CellsCollection.RowsUsed.Keys))
             {
+                if (!rowsUsed.Add(rowNum))
```
lahma, line 939: now capturing closures, if not handled yet, then do the actual logic

jahav:
Thank you. Performance is always nice.

PR #2000, opened 2023-02-01
Use space as culture group separator in AutoFilterTests

lahma:
This test case was failing locally for me when running under full framework. My locale is fi-FI. And parsing target was "10 000.00"


jahav (approved):

jahav:
Thanks for the fix.

PR #2004, opened 2023-02-09
Update DocumentFormat.OpenXML to 2.19.0

PrzemyslawKlys:
Unfortunetly I am not sure why the tests fail, if someone tells me why they fail I could maybe try and fix them. It would be cool to have OpenXML SDK updated to newest version

### DIFF
--- a/ClosedXML.Examples/ClosedXML.Examples.csproj
+++ b/ClosedXML.Examples/ClosedXML.Examples.csproj
@@ -16,7 +16,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\ClosedXML\ClosedXML.csproj" />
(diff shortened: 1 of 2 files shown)


jahav:
Build fails after update to DocumentFormat.OpenXml 2.19 because images are not saved with the correct extension (`.jpg` or `.png`), but with a `.bin` extension. I tracked it down to an reversed condition in `if (TryGetValue)` and there is already an [issue](https://github.com/dotnet/Open-XML-SDK/issues/1305) and a PR to fix.

Therefore I will skip the 2.19, because I like having correct image extensions. Hopefully, it will be fixed in 2.20, so I am closing the PR without merging and will wait for for the 2.20. Will think about 2.18.

PR #2009, opened 2023-02-13
ClosedXML recent packages' updates

DonPatch74:
Firstly, I wanted to say say huge thank you for the package that you have provided. Very great piece of software that helps with spreadsheets.

Secondly, I noticed that security solution (SNYK) I use for my application pointed out few issues, which are only related to packages used by ClosedXML. I have pulled the branch and updated everything just to make sure that will be resolved and it occured that this simple procedure fixed everything.

Please check the updated packages and if it is okay merge it, so I could keep using the ClosedXML as nuget package instead of exporting it ito bunch of dll files.

Kindest regards,
Ben

### DIFF
--- a/ClosedXML.Examples/ClosedXML.Examples.csproj
+++ b/ClosedXML.Examples/ClosedXML.Examples.csproj
@@ -16,8 +16,8 @@
 
   <ItemGroup>
     <ProjectReference Include="..\ClosedXML\ClosedXML.csproj" />
-    <PackageReference Include="DocumentFormat.OpenXml" Version="2.16.0" />
-    <PackageReference Include="morelinq" Version="2.10.0" />
+    <PackageReference Include="DocumentFormat.OpenXml" Version="2.19.0" />
+    <PackageReference Include="morelinq" Version="3.3.2" />
   </ItemGroup>
 
 </Project>
--- a/ClosedXML.Tests/ClosedXML.Tests.csproj
+++ b/ClosedXML.Tests/ClosedXML.Tests.csproj
@@ -16,10 +16,10 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CSharp" Version="4.4.1" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.8.3" />
-    <PackageReference Include="NUnit" Version="3.13.1" />
-    <PackageReference Include="NUnit3TestAdapter" Version="3.17.0" />
+    <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.4.1" />
+    <PackageReference Include="NUnit" Version="3.13.3" />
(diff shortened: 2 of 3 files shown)


jahav:
Please be more specific about the issues. Just blanket update won't work (e.g. every change in XLParser is also change in grammar, i.e. breaking change) and as a library, I generally go for lower versions, see also https://github.com/ClosedXML/ClosedXML/pull/1950#issuecomment-1371689496

DonPatch74:
Sure thing, this is the list of the security vulnerabilities found by our security provider:

- Regular Expression Denial of Service (ReDoS)
Introduced through ClosedXML@0.100.3 > XLParser@1.5.2 > Irony.NetCore@1.0.11 > NETStandard.Library@1.6.1 > System.Xml.XDocument@4.3.0 > System.Xml.ReaderWriter@4.3.0 > System.Text.RegularExpressions@4.3.0

- Information Exposure
Introduced through ClosedXML@0.100.3 > XLParser@1.5.2 > Irony.NetCore@1.0.11 > NETStandard.Library@1.6.1 > System.Net.Http@4.3.0

- Privilege Escalation
Introduced through ClosedXML@0.100.3 > XLParser@1.5.2 > Irony.NetCore@1.0.11 > NETStandard.Library@1.6.1 > System.Net.Http@4.3.0

- Improper Certificate Validation
Introduced through ClosedXML@0.100.3 > XLParser@1.5.2 > Irony.NetCore@1.0.11 > NETStandard.Library@1.6.1 > System.Net.Http@4.3.0

- Denial of Service (DoS)
Introduced through ClosedXML@0.100.3 > XLParser@1.5.2 > Irony.NetCore@1.0.11 > NETStandard.Library@1.6.1 > System.Net.Http@4.3.0

- Authentication Bypass
Introduced through ClosedXML@0.100.3 > XLParser@1.5.2 > Irony.NetCore@1.0.11 > NETStandard.Library@1.6.1 > System.Net.Http@4.3.0

Please let me know if there is anything else to provide.

jahav:
These are all minimal versions from NETStandard.Library. 

Netstandard is only a specification of API, not an implementation. At least for core, application doesn't use minimal version, but the current implementation from .NET Core of the application (e.g. 7 for net7).

When I look at the discussion in .NET Standard repo, they say basically same thing (aka false positive): https://github.com/dotnet/standard/issues/1786

> System.Text.RegularExpressions never applies a vulnerable binary on .NETFramework. It applies a facade dll that typeforwards to System.dll where all this code lives. The facade dll is not vulnerable as it does not contain the code. System.Text.RegularExpressions also does not apply its binary on .NETCore2.0 and later. There the implementation is provided by the shared framework. This package only exists for delivering the implementation to older frameworks (.netcore1.x), which are now out of support.

Also, updating packages won't solve anything, because they will still reference netstandard 1.6.1 (thus transitively minimal 4.3.0 versions).

This is only security issue, if you run retired framework/core (e.g. 4.5 or core2), so I consider it to be a false positive.

DonPatch74:
Thank you for the explanation. Then, I am gonna use our upgraded version and keep on checking for any updates.

Issue #1096, opened 2018-12-07
Incorrect deletion of the image when you add the same image again


v0idname:
**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.93.0, 0.93.1

**What is the current behavior?**

If you add several images, delete one of them and add a new one with the same name as the deleted image, a red cross will be displayed instead of the deleted image (excel 2016).
For example, excel 2007 does not open the resulting file at all.

**What is the expected behavior or new feature?**

Do not display the deleted image.

**Did this work in previous versions of our tool?  Which versions?**

No

## Reproducibility

**Code to reproduce problem:**
```c#
public void Main()
{
            // delete excel file
            string fileName = "D:\\example.xlsx";
            if (File.Exists(fileName)) File.Delete(fileName);

            // create book
            var book = new XLWorkbook();
            var sheet = book.Worksheets.Add("Sheet1");

            // add pictures
            var picName = "pic_name";
            var image = Image.FromFile("D:\\picture1.png");
            ImageConverter imgConverter = new ImageConverter();
            byte[] arr = (byte[])imgConverter.ConvertTo(image, typeof(byte[]));
            MemoryStream stream = new MemoryStream();
            stream.Write(arr, 0, arr.Length);
            sheet.AddPicture(stream, picName).MoveTo(sheet.Cell(3, 3).Address);

            var picName2 = "pic_name2";
            sheet.AddPicture(stream, picName2).MoveTo(sheet.Cell(3, 10).Address); // code is working fine without this line

            // save book
            book.SaveAs(fileName);

            // open same book
            var fileStream = new FileStream(fileName, FileMode.Open);
            book = new XLWorkbook(fileStream);
            sheet = book.Worksheet(1);

            // delete picture
            sheet.Pictures.Delete(picName);
            book.Save();

            // adding picture with same name
            sheet.AddPicture(stream, picName).MoveTo(sheet.Cell(17, 3).Address);

            // save and close book
            book.Save();
            fileStream.Close();
}
```
- [x] I attached a sample spreadsheet.
[example.xlsx](https://github.com/ClosedXML/ClosedXML/files/2656078/example.xlsx)
- [x] used image
![picture1](https://user-images.githubusercontent.com/14361976/49631069-af6f0500-fa01-11e8-8257-40e62a2e8553.png)
- [x] screenshot of the resulting file in excel 2016
[...]

PR #2038, opened 2023-03-24
merge fix issue #1096 that caused by incomplete deletion of image anchor.

xcosmos6:
Fixed issue #1096 
It was fixed on release 0.95.4 and then merged to develop. Not sure if this would work as it seems that there are a lot more difference than the 3 lines of code that I added. Please let me know if I should have done this differently and I can re-do the request. Thank you!


xcosmos6:
Now I see how different this file is in you latest version. So this issue might has already been fixed in the newest release.  Please let me know if this is the case so I can close this request. Thank you!

henriw88:
The issue still persists in version 0.97. 
We are using the .Delete property, which leads to the incomplete deletion.
This is some sample code we use:

`var pictureToDelete = workSheet.Pictures.FirstOrDefault(pic => pic.Name == "VesselGA");
            pictureToDelete?.Delete();
`
The result looks like this:
excel picture

xcosmos6:
> The issue still persists in version 0.97. We are using the .Delete property, which leads to the incomplete deletion. This is some sample code we use:
> 
> `var pictureToDelete = workSheet.Pictures.FirstOrDefault(pic => pic.Name == "VesselGA"); pictureToDelete?.Delete(); ` The result looks like this: excel picture

I am not sure what to on this as nobody has responded to my pull request or my question. I assume you can also add those 3 lines of code into version 0.97?

igitur:
Pull requests without unit tests will likely not be accepted.

xcosmos6:
> Pull requests without unit tests will likely not be accepted.

I am sorry that I actually don't know how to add the test. But I only changed 3 lines of code, so, maybe you can make changes accordingly on the develop branch when you work on the code next time? 

I did pull down the latest develop branch and made the change to it too yesterday.  It's in the `ClosedXML/Excel/IO/WorksheetPartWriter.cs` file, function `GetWorksheetDom()`, in the `foreach` loop after line 1321.
`

                foreach (var removedPicture in xlPictures.Deleted)
                {
                    // Clear existing anchor of the deleted picture
                    var existingAnchor = GetAnchorFromImageId(worksheetPart.DrawingsPart, removedPicture);
                    if (existingAnchor != null)
                        worksheetPart.DrawingsPart.WorksheetDrawing.RemoveChild(existingAnchor);

                    worksheetPart.DrawingsPart.DeletePart(removedPicture);
                }
`
So, maybe you can disregard this pull request?
Thanks!

PR #2039, opened 2023-03-24
Added OpenSettings to LoadOptions

rafael-f:
Added OpenSettings to LoadOptions so we can customize how the sheet is opened with the XLWorkbook constructor.

This allows us to use the RelationshipErrorHandlerFactory on the OpenSettings with new XLWorkbook(stream, options).

Otherwise there's no way to handle malformed uris graciously. This complements the solution for this issue: https://github.com/ClosedXML/ClosedXML/issues/249

### DIFF
--- a/ClosedXML/Excel/LoadOptions.cs
+++ b/ClosedXML/Excel/LoadOptions.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Drawing;
 using ClosedXML.Graphics;
+using DocumentFormat.OpenXml.Packaging;
 
 namespace ClosedXML.Excel
 {
@@ -36,5 +37,15 @@ namespace ClosedXML.Excel
             get => _dpi;
             set => _dpi = value.X > 0 && value.Y > 0 ? value : throw new ArgumentException("DPI must be positive");
         }
+
+        /// <summary>
+        /// Represents the settings when opening a document.
+        /// </summary>
+        public OpenSettings OpenSettings { get; set; }
+
+        /// <summary>
+        /// In ReadWrite mode. False for Read only mode.
+        /// </summary>
+        public bool IsEditable { get; set; }
     }
 }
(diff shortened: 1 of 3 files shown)


jahav:
@rafael-f I think that RelationshipErrorHandlerFactory has been removed recently (see https://github.com/dotnet/Open-XML-SDK/pull/1322). It is also part of a [changelog](https://github.com/dotnet/Open-XML-SDK/blob/main/CHANGELOG.md) for 3.0.0.

jahav:
According to the OpenXmlSdk [roadmap](https://github.com/dotnet/Open-XML-SDK/blob/main/docs/ROADMAP.md) (made few months ago), v3.0 should have preview in March and release June 2023 (3 months) and v3.0 has solved the malformed uri issue by wrapping whole Package into its own IPackage interface, I won't be merging the PR. 

I think it is better to use the solution in the underlaying library than introducing an API that would be dropped in half a year.

Sidenote: ClosedXml uses it's own types that are later translated to mostly OpenXmlSdk type. In theory, OpenXmlSdk is a dependency that can be dropped, though it will likely never happen.

rafael-f:
> @rafael-f I think that RelationshipErrorHandlerFactory has been removed recently (see [dotnet/Open-XML-SDK#1322](https://github.com/dotnet/Open-XML-SDK/pull/1322)). It is also part of a [changelog](https://github.com/dotnet/Open-XML-SDK/blob/main/CHANGELOG.md) for 3.0.0.

Interesting, the RelationshipErrorHandlerFactory was just an example though (and the one I need most right now), there are more settings on OpenSettings that we currently can't use. The future removal of the RelationshipErrorHandlerFactory will not affect this PR either. I believe there are still benefits to allowing us to use OpenSettings with the XLWorkbook constructor.

Plus there doesn't seem to be a date for v3 release of OpenXML. Currently there's no way to open an excel with a broken uri, using the RelationshipErrorHandlerFactory is the easiest way to fix it for anyone who used the XLWorkbook constructor.

PR #2042, opened 2023-03-30
Fix Stack Overflow exception and optimize for bigger datasets

JakubTracz:
`CurrentRegion` property on `XLCell` was throwing a Stack Overflow exception for bigger datasets (starting at c.a. 2000 rows) and was noticeably slow. I've changed the approach into a non-recursive one to fix the issue and provide a significantly better performance.

### DIFF
--- a/ClosedXML/Excel/Cells/XLCell.cs
+++ b/ClosedXML/Excel/Cells/XLCell.cs
@@ -2130,31 +2130,88 @@ namespace ClosedXML.Excel
             }
         }
 
-        public IXLRange CurrentRegion
+        public IXLRange CurrentRegion => Worksheet.Range(FindCurrentRegion());
+
+        private IXLRangeAddress FindCurrentRegion()
         {
-            get
+            var sheet = Worksheet;
+
+            var minRow = _rowNumber;
+            var minCol = _columnNumber;
+            var maxRow = _rowNumber;
+            var maxCol = _columnNumber;
+
+            bool hasRegionExpanded;
+
+            do
             {
-                return this.Worksheet.Range(FindCurrentRegion(this.AsRange()));
+                hasRegionExpanded = false;
+
+                var borderMinRow = Math.Max(minRow - 1, XLHelper.MinRowNumber);
+                var borderMaxRow = Math.Min(maxRow + 1, XLHelper.MaxRowNumber);
+                var borderMinColumn = Math.Max(minCol - 1, XLHelper.MinColumnNumber);
+                var borderMaxColumn = Math.Min(maxCol + 1, XLHelper.MaxColumnNumber);
+
+                if (minCol > XLHelper.MinColumnNumber &&
+                    !IsVerticalBorderBlank(sheet, borderMinColumn, borderMinRow, borderMaxRow))
+                {
+                    hasRegionExpanded = true;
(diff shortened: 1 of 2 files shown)


ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -2130,31 +2130,55 @@ public IXLRangeAddress FormulaReference
...
-                return FindCurrentRegion(this.Worksheet.Range(grownRangeAddress));
+            do
+            {
+                hasRegionExpanded = false;
+
+                for (var row = Math.Max(minRow - 1, 1); row <= Math.Min(maxRow + 1, 1048576); row++)
+                {
+                    for (var col = Math.Max(minCol - 1, 1); col <= Math.Min(maxCol + 1, 16384); col++)
+                    {
+                        if (row >= minRow && row <= maxRow && col >= minCol && col <= maxCol)
```
jahav: I think it would be better to just check the grown areas (aka borders). The loop is happening n^2 times, but has to happen only for the new borders (4xn times).

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -2130,31 +2130,55 @@ public IXLRangeAddress FormulaReference
...
-                return this.Worksheet.Range(grownRangeAddress).RangeAddress;
-            else
-                return FindCurrentRegion(this.Worksheet.Range(grownRangeAddress));
+            do
+            {
+                hasRegionExpanded = false;
+
+                for (var row = Math.Max(minRow - 1, 1); row <= Math.Min(maxRow + 1, 1048576); row++)
+                {
+                    for (var col = Math.Max(minCol - 1, 1); col <= Math.Min(maxCol + 1, 16384); col++)
```
jahav: Please use `XLHelper.MaxRowNumber` and `XLHelper.MaxColumnNumber`

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -2130,31 +2130,55 @@ public IXLRangeAddress FormulaReference
...
+
+                        minRow = Math.Min(minRow, row);
+                        minCol = Math.Min(minCol, col);
+                        maxRow = Math.Max(maxRow, row);
+                        maxCol = Math.Max(maxCol, col);
+                    }
+                }
+
+                if (!hasRegionExpanded)
+                    continue;
```
jahav: I think break is more appropriate at this place than continue. Continue jumps to the loop condition that immediately ends.

The if condition is likely redundant, when the currentRangeAddres is moved out of the loop.

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -2130,31 +2130,55 @@ public IXLRangeAddress FormulaReference
...
+
+                currentRangeAddress = new XLRangeAddress(Worksheet, $"{Worksheet.Cell(minRow, minCol).Address}:{Worksheet.Cell(maxRow, maxCol).Address}");
```
jahav: Please move this after the loop, variable is not used. It is a useless allocation in the loop.

jahav (changes requested):

jahav (approved):

jahav:
Thank you.

Issue #1712, opened 2021-08-23
System.InvalidOperationException: 'Table name cannot be a valid Cell Address 'T1'.' but Excel currently allows that name


NikiLentz:
## Read and complete the full issue template

Do not randomly delete sections. They are here for a reason.

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [ ] Yes
- [ ] No

**Version of ClosedXML**

0.95.4

**What is the current behavior?**

Currently creating a workbook with a worksheet named "T1" throws an error. IMO this is wrong, since Excel 365 allows me to do this

**What is the expected behavior or new feature?**

No Exception thrown anymore for "Table name cannot be a valid cell adress" or at least a possibility to willfully ignore it.

**Is this a regression from the previous version?**

Regressions get higher priority. Test against the latest build of the previous minor version. For example, if you experience a problem on v0.95.3, check whether it the problem occurred in v0.94.2 too. 

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

Without a code sample, it is unlikely that your issue will get attention. Don't be lazy. Do the effort and assist the developers to reproduce your problem. Code samples should be [minimal complete and verifiable](https://stackoverflow.com/help/mcve). Sample spreadsheets should be attached whenever applicable. Remove sensitive information.

**Code to reproduce problem:**
```c#
```
public void Main()
{
 XLWorkbook wb = new XLWorkbook();
                    
                wb.Worksheets.Add(dataTable, "T1");
}
```
```

- [ ] I attached a sample spreadsheet.  (You can drag files on to this issue)

danreg:
Is there any plan to fix this? It was also occuring for worksheet named TSS1234

NickNack2020:
@jahav this is a problem because sheetName is used when inserting the table, and Table names can't be named like a cell reference like T1.

Code that causes the issue
```
       public IXLWorksheet Add(DataTable dataTable, String sheetName)
        {
            var ws = Add(sheetName);
            ws.Cell(1, 1).InsertTable(dataTable, sheetName);
            return ws;
        }
```
I propose we add an overload

```
 public IXLWorksheet Add(DataTable dataTable, String sheetName,  String tableName){
            var ws = Add(sheetName);
            ws.Cell(1, 1).InsertTable(dataTable, tableName);
            return ws;
        }
```

[...]

PR #2044, opened 2023-03-31
Add ability to set table name when adding a worksheet with datatable …

NickNack2020:
Fixes Issue #1712  - Add ability to set table name when adding a worksheet with datatable.


ClosedXML/Excel/XLWorkbook.cs
```diff
@@ -857,11 +852,41 @@ public void AddWorksheet(IXLWorksheet worksheet)
             worksheet.CopyTo(this, worksheet.Name);
         }
 
+        /// <summary>
+        /// Add a worksheet with a table at Cell(row:1, column:1). The datatable's name is used for the
+        /// worksheet name and the table name.
+        /// </summary>
+        /// <param name="dataTable">Datatable to insert</param>
+        /// <returns>Inserted Worksheet</returns>
+        public IXLWorksheet AddWorksheet(DataTable dataTable)
```
jahav, line 855: Please place comments to the interface, not an implementation. That way, users will see them through IntelliSense.

ClosedXML/Excel/XLWorksheet.cs
```diff
@@ -1551,25 +1551,6 @@ private void CheckRangeNotOverlappingOtherEntities(XLRange range)
                 throw new InvalidOperationException($"The range {range.RangeAddress.ToStringRelative(includeSheet: true)} overlaps with the worksheet's autofilter.");
         }
 
-        private string GetNewTableName(string baseName)
```
NickNack2020, line 1554: Refactored out into a reusable class - See TableNameGenerator

ClosedXML/Excel/XLWorksheets.cs
```diff
@@ -160,9 +160,14 @@ public IXLWorksheet Add(DataTable dataTable)
         }
 
         public IXLWorksheet Add(DataTable dataTable, String sheetName)
+        {
+            return Add(dataTable, sheetName, TableNameGenerator.GetNewTableName(_workbook));
```
NickNack2020, line 164: It uses the new table name generator class and provides a generated table name, as suggested in the ticket.

ClosedXML/Excel/XLWorkbook.cs
NickNack2020: done

NickNack2020:
@jahav I believe I made the suggested changes. Let me know if there is anything else.

jahav:
Resolves #1712

Issue #934, opened 2018-06-29
sheet.Rows.AdjustToContents() has no effect


weskerjiang:
## Read and complete the full issue template

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature

**Version of ClosedXML**

0.93

**What is the current behavior?**

The IXLWorkSheet.Rows().AdjustToContens() does not work.

**What is the expected behavior or new feature?**

I want the row's height can change when the cell's content is too long.

**Did this work in previous versions of our tool?  Which versions?**

No! It does not work in 0.92.1

## Reproducibility
**Code to reproduce problem:**
```c#
public void Main()
{
    XLWorkbook wb = new XLWorkbook();
    IXLWorksheet sheet = wb.Worksheets.Add("data");
    sheet.Column("A").Width = 6;
    sheet.Cell("A1").Value = "This is a very long content to test the AdjustToContents function.";
    sheet.Row(1).Style.Alignment.WrapText = true;
    sheet.Row(1).AdjustToContents();    // does not work

    wb.SaveAs(@"C:\demo.xlsx");
}
```
- [x] I attached a sample spreadsheet.  (You can drag files on to this issue)
[demo2.xlsx](https://github.com/ClosedXML/ClosedXML/files/2148177/demo2.xlsx)

igitur:
Complete the full issue template.

igitur:
Did this work in previous versions of our tool? Which versions?

igitur:
Also, I can't run your code sample. Read the instructions in the issue template again. I'll reopen the issue when you've completed it fully.

weskerjiang:
I've changed the demo code.

igitur:
Still can't run it.

igitur:
> I attached a sample spreadsheet. (You can drag files on to this issue)

See this part?

weskerjiang:
So I've attached a xlsx file.

igitur:
You should be calling
```c#
sheet.Column(1).AdjustToContents();    // does not work
```

Pankraty:
Ehm, adjusting row heights and column width are completely different operations, aren't they?

weskerjiang:
I think sheet.Column(1).AdjustToContents()   is wrong because my row's width is a fixed value like 10 or 5.
If I use sheet.Column(1).AdjustToContents(),the row's width will change.

igitur:
@weskerjiang Sorry for being rude earlier.

[...]

PR #2048, opened 2023-04-03
Take word wrapping into account when adjusting row height

machekku:
Hello,
This is my initial attempt to add text wrapping support when auto adjusting row height (https://github.com/ClosedXML/ClosedXML/issues/934).

Please note: This pull request is not yet intended to be merged. It serves as a base to discuss the solution. Once architectural concerns are resolved, I will add unit tests and polish the core algorithm itself.

Here's outline of my solution:
1. When text wrapping is enabled, Excel will break long lines so that text is not wider then the cell.
2. Lines can be wrapped on white space characters.
3. Content width is calculated in XLRow.CalculateMinRowHeight() method as Column Width - Pixel Padding.
4. Cell content height is calculated in XLRow.GetContentHeight(). As noted above, let's not focus on the algorithm itself this time yet. General idea is:
- scan glyphs and accumulate them into clusters (a cluster is either a word-like object or a group of white space characters),
- control line width and introduce line breaks when needed, moving current word into next line.
6. To find line break opportunities, XLRow.GetContentHeight() method needs to know which glyphs are white space.
7. GlyphBox class is extended to include IsWhiteSpace property, which is set via constructor.
8. DefaultGraphicEngine.GetGlyphBox() determines if a glyph box represents white space by checking if any of grapheme code points is a white space code point. (I am really not sure if this assumption is correct)

Please let me know if this is the right approach.

### DIFF
--- a/ClosedXML/Excel/Rows/XLRow.cs
+++ b/ClosedXML/Excel/Rows/XLRow.cs
@@ -306,6 +306,8 @@ namespace ClosedXML.Excel
             var glyphs = new List<GlyphBox>();
             XLStyle cellStyle = null;
             var rowHeightPx = 0;
+            var workbookMdw = (int)Math.Round(engine.GetMaxDigitWidth(Worksheet.Workbook.Style.Font, dpi.X));
+
             foreach (var cell in Row(startColumn, endColumn).CellsUsed().Cast<XLCell>())
             {
                 // Clear maintains capacity -> reduce need for GC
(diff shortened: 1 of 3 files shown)


jahav:
Hi, the structure looks good to me.

I am not sure about adding a flag to the `GlyphBox`. Maybe it would be better to create a new structure (e.g. equivalent of `internal readonly record struct Grapheme(GlyphBox Box, bool IsSpace)` and return that from `XLCell`). Graphic engine should be kept mostly as a very thin wrapper and shouldn't deal with fonts/images and this is an information for Excel-specific layout. 

I haven't realized that we would need to know if a glyph box was for space and I am pretty sure there will be other other layout related things that will need to be changed and I don't want to break public API (`IXLGraphicEngine`+`GlyphBox`) every time layout will have to be fixed.

As for 8., I would just check that grapheme is only 1 codepoint long and it is a space (only space - I have tried to what Excel does with other whitespaces, e.g. En Quad U+2000 and it behaves like a normal character). Multi-codepoint graphemes are used only for emoji as far as I know (e.g. see 43 https://unicode.org/emoji/charts/full-emoji-list.html).

The important thing here is scope, because there there are four places that must work in sync:
* row - adjust to content - wrapped
* row - adjust to content - non-wrapped
* column - adjust to content - wrapped
* column - adjust to content - non-wrapped

One place can't behave differently than others for the feature (in this case non-vertical non-rotated text).

This is a breaking change for a feature used in pretty much every worksheet that sets a text with a newline and uses adjust to content.

Issue #1716, opened 2021-09-02
The identifier `LARGE` was not recognised


eryatish:
## Read and complete the full issue template

Do not randomly delete sections. They are here for a reason.

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [x] Question

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**version tested on 0.95.4**

**On reading range in my excel sheet it showing error 'The identifier `LARGE` was not recognised'. I have also tried to read the first column but it gives me blank, also I have read all the merged cells value as list/array but it is giving me array of null value**

Complete this.

**It should give the calculated value of the cell of range**

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

Without a code sample, it is unlikely that your issue will get attention. Don't be lazy. Do the effort and assist the developers to reproduce your problem. Code samples should be [minimal complete and verifiable](https://stackoverflow.com/help/mcve). Sample spreadsheets should be attached whenever applicable. Remove sensitive information.

**Code to reproduce problem:**
```c#
public void Main()
{
using (var workbook = new XLWorkbook(workbookPath))
                {
// below code is giving me blank value
                    var worksheet = workbook.Worksheet(1);
    worksheet.Range("E42:I42")
                     .Cells()
                     .Select(c => c.CachedValue)
                     .Where(c => c != null)
                     .Where(c => c != "")
                     .ToList();

// I have used the below code just to fetch the first cell value of the range, coz the value is stored in first cell in a range, but got blank
var range1 = worksheet.Cell("E42").GetFormattedString();

// below code gives error of 'The identifier `LARGE` was not recognised'
worksheet.RecalculateAllFormulas();

//below code gives error of 'Cannot convert 'NEW - Prototype'!E42's value to System.String'
var range1 = worksheet.Range("E42:I42");
var cell1 = range1.Cell(1, 1);
cell1.InvalidateFormula();
String cellValue = cell1.GetValue<string>();
}
}
```

- [ ] I cannot attach the spreadsheet as it is highly confidential. I am sharing the formula that is used in range

[...]

PR #2050, opened 2023-04-04
Implement Large Formula - Targets #1716

NickNack2020:
I took a pass at implementing the "Large" formula/function. 

See https://support.microsoft.com/en-us/office/large-function-3af0af19-1190-42bb-bb8b-01672ec00a64; if this implementation meets ClosedXml standards, it should be trivial also to implement small. I will open a PR for Small if this is accepted.


jahav:
The functions that return object are slated for rewrite. Please use the the typed functions that return AnyValue.

It's important that function behaves as close as possible to Excel otherwise it only"mostly" work. Also try to minimize number of allocations on heap, that is problematic when evaluating 1mil of cells. 

https://github.com/ClosedXML/ClosedXML/wiki/Calc-Engine#how-to-develop-a-new-function

PR #2051, opened 2023-04-06
Fix NRE during normal style loading

lahma:
Version used [0.101.0-rc](https://www.nuget.org/packages/ClosedXML/0.101.0-rc)

`WorkbookStylesPart` has been annotated with nullable attribute and when we tried the latest RC, a NRE is being thrown due to files not having styles part:

```text
System.NullReferenceException : Object reference not set to an instance of an object.
   at ClosedXML.Excel.XLWorkbook.LoadSpreadsheetDocument(SpreadsheetDocument dSpreadsheet) in C:\Work\OpenSource\ClosedXML\ClosedXML\Excel\XLWorkbook_Load.cs:line 192
   at ClosedXML.Excel.XLWorkbook.LoadSheets(Stream stream) in C:\Work\OpenSource\ClosedXML\ClosedXML\Excel\XLWorkbook_Load.cs:line 50
   at ClosedXML.Excel.XLWorkbook.Load(Stream stream) in C:\Work\OpenSource\ClosedXML\ClosedXML\Excel\XLWorkbook_Load.cs:line 38
   at ClosedXML.Excel.XLWorkbook..ctor(Stream stream, LoadOptions loadOptions) in C:\Work\OpenSource\ClosedXML\ClosedXML\Excel\XLWorkbook.cs:line 707
   at ClosedXML.Excel.XLWorkbook..ctor(Stream stream) in C:\Work\OpenSource\ClosedXML\ClosedXML\Excel\XLWorkbook.cs:line 698
```
This PR adds a null check which makes our test suite green again. I tried to make use consistent by using same variable that was already checked for null earlier so should be more obvious that it can be nullable. In ideal work of course the `XLWorkbook_Load.cs` would have nullable enabled, but that would be a bigger change.

### DIFF
--- a/ClosedXML/Excel/XLWorkbook_Load.cs
+++ b/ClosedXML/Excel/XLWorkbook_Load.cs
@@ -170,17 +170,11 @@ namespace ClosedXML.Excel
                     Properties.Manager = efp.Properties.GetFirstChild<Manager>().Text;
             }
 
-            Stylesheet s = null;
-            if (dSpreadsheet.WorkbookPart.WorkbookStylesPart != null &&
-                dSpreadsheet.WorkbookPart.WorkbookStylesPart.Stylesheet != null)
-            {
-                s = dSpreadsheet.WorkbookPart.WorkbookStylesPart.Stylesheet;
-            }
-
-            NumberingFormats numberingFormats = s == null ? null : s.NumberingFormats;
-            Fills fills = s == null ? null : s.Fills;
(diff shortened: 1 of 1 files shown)


jahav:
Thanks for the test and bugfix. 

We need to start enabling nullable everywhere, at least `<nullable>true</nullable>` in csproj + `#nullable disable` in every file, so it can be enabled file by file.

PR #2056, opened 2023-04-09
Enable nullables in main project

lahma:
This PR sets `<Nullable>enabled</Nullable>` in main ClosedXml.csproj and annotates the "easy wins" from source code. I left `#nullable disabled` for files that would have cause too much noise or I was otherwise unsure of the correct annotations.


ClosedXML/ClosedXML.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
+    <Nullable>enable</Nullable>
```
jahav: Netstandard2.0 has to use annotations nullable context, because otherwise we will get false positive warnings due to lack of nullable attributes (e.g. `string.IsNullOrWhitespace` and other methods from BCL is not taken into account).

jahav (changes requested):
The enable attribute in csproj adds it everywhere. All files (at the very least all public API files) must be marked with a `#nullable disable` or checked. I favor gradual change, so first add attribute, then check & remove.

As an example, `IXLCell.cs` is not marked as a disabled. That means it is compiled with nullable attributes that declare everything as never null, but some methods/props do return null (e.g. `IXLCell.MergedRange`). Null-aware client code would use it, expecting never to be null and would later fail on NRE because of it.

lahma:
Thanks for the review, I easily get sidetracked and now I'm fixing things upstream 😉 : https://github.com/dotnet/Open-XML-SDK/pull/1399 . I'll try to find time to revisit this with your recommendations.

As a side note, There's again a NRE when using the latest `DocumentFormat.OpenXml` (2.20.0) when reading Excel file, now when trying to access `SpreadsheetDocument.WorkbookPart` which has can-be-null annotation and for some reason now is null (maybe memory optimization, who knows).

jahav (approved):

jahav:
I have kept `#nullable disable` anywhere I had doubts. There are still 366 files with `#nullable disable`, but that is fine with a gradual approach I prefer.

There are several places, where marking a member as a nullable is not a suitable solution, e.g. I am pretty sure all colors from `IXLTheme` are required, `IXLSparline.Location` should never be null.

In some cases, the warning is an indication that API should be cleaner (e.g. `IXLTableField.TotalsRowFormulaA1` throws on disabled totals row, but `IXLTableField.TotalsRowLabel` return null).

I also want to to use multi targeting and nullable attributes for `Try*` method. I will likely just use [Nullable](https://github.com/manuelroemer/Nullable) package, though I have to check it it will work correctly with annotation nullable context. I want to do that first and then continue in removal of `#nullable disabled` I put into every file. That will limit the amount use null forgiving operator.

Sidenote:
As you wrote about the operators that were missing nullable arguments (which caused things like `openXMLColor.Indexed! != null!`), it's possible just to use `openXMLColor.Indexed is not null`, which bypasses custom operators (+should execute less instructions).

lahma:
Thanks for completing this one and sorry that I dropped the ball a bit here.

PR #2061, opened 2023-04-12
SUMIFS multiple dimensions

### LINKED ISSUE #2041: SUMIF doesn't consider only first column

- [x] Bug
- [ ] Feature
- [ ] Question

- [x] Yes
- [ ] No

**Version of ClosedXML**

0.100.3
(Installed via nuget.org using NuGet package manager)

When reading a complex file with lots of formulas and links, it gives wrong formula calculation results

Complete this.

No

## Reproducibility

No special actions are needed: just open the attached file and take the value of cell (13, 9) from the first sheet as an example

Value in file:
image

The value produced by the code:
image

```c#
using (var fs = new FileStream(@"E:\Temp\Расчёт гидравлики СНТ Вишня.xlsx", FileMode.Open, FileAccess.Read))
{
    var workbook = new XLWorkbook(fs);
    var worksheet = workbook.Worksheet(1);
    var xlCell = worksheet.Cell(13, 9);
    xlCell.InvalidateFormula();
    MessageBox.Show(xlCell.Value.ToString());
}
```

[Расчёт гидравлики СНТ Вишня.xlsx](https://github.com/ClosedXML/ClosedXML/files/11112243/default.xlsx)

igitur:
Fixes #2041

Add support and applicable errors for input ranges having multilpe dimensions (more than 1 column)

### DIFF
--- a/ClosedXML.Tests/Excel/CalcEngine/MathTrigTests.cs
+++ b/ClosedXML.Tests/Excel/CalcEngine/MathTrigTests.cs
@@ -2062,6 +2062,16 @@ namespace ClosedXML.Tests.Excel.CalcEngine
             }
         }
 
+        [TestCase("SUMIFS(D1:E5,A1:B5,\"A*\",C1:C5,\">2\")")]
+        [TestCase("SUMIFS(H1:I3,A1:B3,1,D1:F2,2)")]
+        [TestCase("SUMIFS(D:E,A:B,\"A*\",C:C,\">2\")")]
+        public void SumIfs_ReturnsErrorWhenRangeDimensionsAreNotSame(string formula)
+        {
+            using var wb = new XLWorkbook();
+            var ws = wb.AddWorksheet();
+            Assert.AreEqual(XLError.IncompatibleValue, ws.Evaluate(formula));
+        }
+
         [Test]
         public void SumProduct()
         {
--- a/ClosedXML.Tests/Excel/CalcEngine/MathTrigTests.cs
+++ b/ClosedXML.Tests/Excel/CalcEngine/MathTrigTests.cs
@@ -4,7 +4,6 @@ using NUnit.Framework;
 using System;
 using System.Globalization;
 using System.Linq;
-using ClosedXML.Excel.CalcEngine;
 
 namespace ClosedXML.Tests.Excel.CalcEngine
 {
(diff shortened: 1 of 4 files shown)


ClosedXML/Excel/CalcEngine/CalcEngineHelpers.cs
```diff
@@ -134,12 +134,26 @@ internal static bool ValueIsBlank(object value)
...
+            return (long)columnCount * (long)rowCount;
+        }
+
+        /// <summary>
+        /// Get dimensions of the specified range without initializing them all
+        /// (which might cause serious performance issues on column-wide calculations).
+        /// </summary>
+        /// <param name="rangeExpression">Expression referring to the cell range.</param>
+        /// <returns>A tuple of column and row counts.</returns>
+        internal static (int columnCount, int rowCount) GetRangeDimensions(XObjectExpression rangeExpression)
```
jahav: Names of properties of named tuples should be in PascalCase.

ClosedXML.Tests/Excel/CalcEngine/MathTrigTests.cs
```diff
@@ -2063,6 +2063,17 @@ public void SumIfs_ReturnsCorrectValues_ReferenceExampleForSumIf1FromMicrosoft(i
             }
         }
 
+        [Test]
```
jahav: `[Test]` is not required if there is a `[TestCase]`

ClosedXML.Tests/Excel/CalcEngine/MathTrigTests.cs
```diff
@@ -1783,6 +1782,26 @@ public void SumDateTimeAndNumber()
...
```
jahav: `[Test]` is not required if there is a `[TestCase]`

jahav (approved):
Looks good. Thanks for investigating, good work. 

I wonder about other functions, SUMIF has a sibling functions AVERAGEIF/S and COUNTIF/S. We don't implement AVERAGEIF, but COUNTIF has a similar problem.

I think that rather than fixing old range functions, it would be better to reimplement them. The `XObjectExpression` adapter works, but I would like to remove it. 

It would require a new method to validate criteria (outlined [here](https://github.com/ClosedXML/ClosedXML/pull/736#issuecomment-1507067695)).

igitur:
> I wonder about other functions, SUMIF has a sibling functions AVERAGEIF/S and COUNTIF/S. We don't implement AVERAGEIF, but COUNTIF has a similar problem.

I can have a look at the sibling functions and fix them as a short-term goal.

 
> I think that rather than fixing old range functions, it would be better to reimplement them. The `XObjectExpression` adapter works, but I would like to remove it.
> 
> It would require a new method to validate criteria (outlined [here](https://github.com/ClosedXML/ClosedXML/pull/736#issuecomment-1507067695)).

Don't think I have the capacity for that now, sorry. Hence the goal of this PR is a short term fix.

PR #2093, opened 2023-05-27
Use spare arrays to store cell values.

### LINKED ISSUE #1969: Optimize cell collection data structure
ClosedXML has massive memory consumption. This issue contains a proposal for change of data structures that contain cells. Current focus is on fast saving, but this is something for 0.103 or 0.104 (if not bad design).

Key points:
* Each XLCell consumes memory to inclue all possible information in a cell, despite not needing them.
* Represent sheet cells in a sparse 2D-array of slices, where each slice would contain data for subset of data (slice for values, slice for formulas, slice for style, slice for rest [comments, hyperlinks...])
* Sparse array for each slice implemented through two nested lookup arrays (first for rows, the row would then contain sparse lookup table for column).
* Elements of slice wouldn't need to store the address
* XLCell would only be a proxy for data in slices. A disposable object, created whenever needed.
* Can be implemented in phases, completely hidden behind CellCollection, and then used by various features to take advanatage (first would be `InsertData`/`InsertTable`).
* PoC can be done in few days.

Benefits:
* We could store 10M rows with columns 10 each in ~150MB (+whatever text would require).
* Most API is only interested in one area. E.g. when looking for fomulas, only search formulas. 
* Can enumerate only addresses with data and skip empty ones (CellCollection currently goes over all addresses in area)
* Possibly faster data insertion (each new cell won't require an allocation new object)
Drawbacks
* XLCell would always be allocated when requested. Can lead to larger GC. Only matters is user won't use bulk and even then, it is not worse than current situation, where each cell is allocated, as long as it is accessed only once or few times.
* Features requiring data from multiple slices would require multiple lookups (likely mitigated by iterators).

# Problems
Design is intended to solve two problems:

## Cell takes up too much memory
Each cell takes up 152 bytes. That doesn't include anything extra, just each allocated `XLCell` object (assuming 64 bits). It breaks down to
* Address: 18 bytes
* Value: 33 bytes
* Formula: 42 bytes
* Styles: 8 bytes
* Rest (comments+hyperlinks): 16 bytes
* Object overhead 16 bytes
Total 131 bytes. Rest is likely alignment and something I have overseen. The 152 bytes value was calculated by dotMemory.

## CellCollection looks over empty cells

CellCollection is basically a `Dictionary<RowIndex, Dictionary<Column, XLCell>>`. When it has to look for something, it gets a range and it goes over all rows/columns of the range, wheter there is a cell or not and then it checks for data of a cell, if there is one.
* There can be many useless lookups.
* If we are looking for a specific content, all cells have to be inspect all cells for the content anyway. Example: when ClosedXML deletes a row, it has to loop over all cells to find ones that have formulas, to potentially fix the formula (if it contained reference below the deleted row, move the reference upwards)).

# Design

Design stands on two principles:
* Each dimension of the worksheet is represented as a sparse lookup table
* Different content of each cell is represented by a different slice.

## Sparse/Two-level lookup table (**2LUT**)
*This structure must have a name, but I have no idea what name, maybe jagged arrays*

The values would be a generic values and would be stored in the secondary level. Primary level would be directly indexable and it woudl start from 0 and would grow (likely double each time).

For secondary level, there are several alternatives:
* Just allocate up to 32 items (start at 1 and double size each time there isn't enough space)
* Use bitmap (UInt32) to indicate which fields are packed in the array in the order of set bits of the bitmask. The befit would mainly be lone cells, where to store info about 1 cell, we might need to allocate 32 other elements. On the other hand, it should be rare. Might save some memory, might be just a performance eating complication with insufficient benefits.

Simplest implementation of LUT
two-level-lookup2

```csharp
class Lut<TContent>
{
	private const int BottomLutBits = 5;
	private const int BottomLutMask = (1 << BottomLutBits) - 1;

	public TContent[][] _nodes = new TContent[1][];

	public KeyValuePair<bool, TContent> Get(int index) // Zero based index
	{
		var topLevelIndex = index >> BottomLutBits;
		if (topLevelIndex >= _nodes.Length)
			return new KeyValuePair<bool, TContent>(false, default);

		var node = _nodes[topLevelIndex];
		if (node is null)
			return new KeyValuePair<bool, TContent>(false, default);

		var bottomLevelIndex = index & BottomLutMask;
		if (bottomLevelIndex >= node.Length)
			return new KeyValuePair<bool, TContent>(false, default);

		return new KeyValuePair<bool, TContent>(true, node[bottomLevelIndex]);
	}

	public void Set(int index, TContent value)
	{
		var topLevelIndex = index >> BottomLutBits;
		while (topLevelIndex >= _nodes.Length)
			_nodes = DoubleArray(_nodes);

		var bottomLevelIndex = index & BottomLutMask;
		var node = _nodes[topLevelIndex];
		if (node is null)
		{
			var initialSize = 1;
			while (bottomLevelIndex >= initialSize)
				initialSize *= 2;

			_nodes[topLevelIndex] = node = new TContent[initialSize];
		}
		else
		{
			while (bottomLevelIndex >= node.Length)
				Array.Resize(ref node, node.Length * 2);
		}

		node[bottomLevelIndex] = value;
	}

	// Array.Resize still doesn't work for jagged arrays
	private static TContent[][] DoubleArray(TContent[][] originalArray)
	{
		var originalLength = originalArray.Length;
		TContent[][] doubledArray = new TContent[originalArray.Length * 2][];
		Array.Copy(originalArray, doubledArray, originalLength);
		return doubledArray;
	}
}
```
First attempt will likely be done with first approach, because it is simpler and works for large rectangular areas filled with data (=99% of large workbooks).

The exact details (2/3 lvls, compressed/not) will be decided by benchmarking. Compressed has appeal and we could behave differently during bulk insert (allocate per source)/individual manipulation. It would likely needed some switch for incorrect APIn usage (e.g. if 10k cells inserted individually and not even one InsertRange used, just disable compression).

Comparessed 2LUT
two-level-lookup

## Slice

Slice would be a 2LUT for columns in 2LUT for rows. We expect that data are filled from left-top corner, so 2LUT would grow roughly in line with how data are filled (2LUT starts at 0 and grows as necessary).

Each slice would contain a different values, e.g. there would be 
* a slice for cell values - that would be rather full
* a slice for formulas - likely far less full
* a slice for styles
* a slice for rest of data (e.g. reference to comments, hyperlinks, just overall rarely used stuff) - items would be rather large, but there should be only a few of them.

slices

Of course each slice has some overhead, so I don't think we should go over 3-4 slices.
Each slice would have a default value, i.e. value returned for missing addresses (e.g. style).

The slices would contain all data of cells, distributed across slices. Since each slice would contain only specific data, it would comparatively save a lot of memory. I.e. currently each cell has to have 8 byte for a reference to a comment, even if most workbooks have very few comments. That is a wasted memory.

Slice would have only have the it's values in itself. That means if there were only a few values, it would be faster to find them compared to current `CellCollection`.

The downside is that it might be necessary to lookup multiple slices for some operations. Though many bulk operations operate on same type of structure (e.g. `IXLRangeBase.SetValue` sets value, `IXLRangeBase.DeleteComments` deletes comments), there are some  that require lookup of multiple slices (e.g. `IsEmpty(XLCellsUsedOptions)`/`Colar(XLClearOptions)`).

Note that no slice would need to store an address in the slice. Another saving. Slice would provide 

## XLCell
XLCell would stop being the central store of info about cell, but instead would be an object that would return values from slices. It would be basically disposable, created as necessary. Inner workings of ClosedXML would work more with slices. User would of course still got XLCell, but it would be an adapter, translating calls to its method by to slices data. They would likely be a very shortlived objects, created as necessary for the user.

# FAQ
## Why not SortedDictionary/BCL?
Sorted dictionary uses red-black tree that uses [~38 bytes](https://referencesource.microsoft.com/#System/compmod/system/collections/generic/sortedset.cs,67e006fe2743f817) per node. Considering payload should be nearly always less, it would at least double amount of required memory, unlike 5%-10%. for bastardized tree. That overhead would be paid for each slice again. There is also less cache locality and less control and log(n) access time.

`List<T>` has an additional overhead of 32 bytes compared to an array. It matters when there is enough of them.

# Reference counting
One thing that should be done in conjunction with the slices is a reference counting of texts and formulas.
Following image show how it should work for strings, but the real improvements will be formulas (an equivalent principle, where formula will be stored in RC tree-form).
reference-counting

Reference counting of strings will half the memory required for storing cell value, instead of 8 byte reference, we will only store number that also stores values of number, boolean and error. I still need enum to determine content of the number, but that is just 1 bytes (likely 2/4 for memory alignment).

This of course can be done even today, but I am hesitant to invest time into modification of XLCell when it will be revamped in not-too-distant future.

# Things to do
* Think about it after some time again. Maybe it is just a wrong design. Check with other devs.
* Do a proof of concepts, data structure should be roughly correct, but operations would be limited to insert data (likely `InsertData`) and lookup of values.

jahav:
Resolves #1969


ClosedXML.Tests/Utils/StreamHelper.cs
```diff
@@ -27,6 +27,9 @@ public static class StreamHelper
             // count and uniqueCount were removed due to streaming, but they are used by every file, for now ignore difference.
             ("/xl/sharedStrings.xml", XName.Get("sst", @"http://schemas.openxmlformats.org/spreadsheetml/2006/main"), XName.Get("count")),
             ("/xl/sharedStrings.xml", XName.Get("sst", @"http://schemas.openxmlformats.org/spreadsheetml/2006/main"), XName.Get("uniqueCount")),
+            // TODO: Remove, only temporary
```
jahav: Reenable. This was only temporary, I know have dimensions.

ClosedXML/Excel/Cells/ISliceEnumerator.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace ClosedXML.Excel
+{
+    /// <summary>
+    /// Enumerator over a slice, with additional properties.
+    /// </summary>
+    internal interface ISliceEnumerator
```
jahav: Do I really need this? Perhaps `IEnumerator<XLSheetPoint>` would be more fitting.

ClosedXML/Excel/Style/XLStylizedBase.cs
```diff
@@ -15,7 +15,7 @@ internal abstract class XLStylizedBase : IXLStylized
         /// <summary>
         /// Read-only style property.
         /// </summary>
-        internal XLStyleValue StyleValue { get; private protected set; }
+        internal virtual XLStyleValue StyleValue { get; private protected set; }
```
jahav, line 20: I should reimplement IXLStylized directly in the `XLCell`, but Children and other places rely on it being `XLStylizedBase`.

ClosedXML/Excel/Columns/XLColumnCollection.cs
```diff
@@ -6,6 +6,8 @@ namespace ClosedXML.Excel
 {
     internal class XLColumnsCollection : IDictionary<Int32, XLColumn>
     {
+        internal int MaxColumn { get; private set; }
```
jahav: Only 1 usage, is it still needed? Also, `ShiftColumnRight` doesn't update max.

ClosedXML/Excel/IO/WorksheetPartWriter.cs
```diff
@@ -111,16 +111,16 @@ internal class WorksheetPartWriter
...
-                sheetDimensionReference = "A1:" + XLHelper.GetColumnLetterFromNumber(maxColumn) +
-                                          maxRow.ToInvariantString();
-            }
+//            if (xlWorksheet.Internals.CellsCollection.Count > 0)
+//            {
+//                maxColumn = xlWorksheet.Internals.CellsCollection.MaxColumnUsed;
+//                var maxRow = xlWorksheet.Internals.CellsCollection.MaxRowUsed;
+//                sheetDimensionReference = "A1:" + XLHelper.GetColumnLetterFromNumber(maxColumn) +
+//                                          maxRow.ToInvariantString();
+//            }
```
jahav: Reenable.

ClosedXML/Excel/Ranges/XLRangeBase.cs
```diff
@@ -1090,26 +1072,26 @@ private IXLRangeColumns InsertColumnsBeforeInternal(Boolean onlyUsedCells, Int32
 
             if (!onlyUsedCells)
```
jahav, line 1084: Check exactly, what does `onlyUsedCell` supposed to do here.

ClosedXML/Excel/Ranges/XLRangeBase.cs
```diff
@@ -1295,67 +1280,45 @@ private IXLRangeRows InsertRowsAboveInternal(Boolean onlyUsedCells, Int32 number
...
             }
 
-            var cellsToInsert = new Dictionary<IXLAddress, XLCell>();
-            var cellsToDelete = new List<IXLAddress>();
-            int firstRow = RangeAddress.FirstAddress.RowNumber;
-            int firstColumn = RangeAddress.FirstAddress.ColumnNumber;
-            int lastColumn = Math.Min(
-                RangeAddress.FirstAddress.ColumnNumber + ColumnCount() - 1,
-                Worksheet.Internals.CellsCollection.MaxColumnUsed);
+            foreach (var group in Worksheet.SparklineGroups)
```
jahav: This is duplicated.

ClosedXML/Excel/Ranges/XLRangeBase.cs
```diff
@@ -1476,59 +1439,91 @@ public void Delete(XLShiftDeletedCells shiftDeleteCells)
...
-                        Worksheet.Internals.CellsCollection.MaxColumnUsed);
+//                    cellsQuery = Worksheet.Internals.CellsCollection.GetCells(
+//                        RangeAddress.FirstAddress.RowNumber,
+//                        RangeAddress.FirstAddress.ColumnNumber,
+//                        RangeAddress.LastAddress.RowNumber,
+//                        Worksheet.Internals.CellsCollection.MaxColumnUsed);
+
+                    Worksheet.Internals.CellsCollection.DeleteAreaAndShiftLeft(XLSheetRange.FromRangeAddress(RangeAddress));
+
+                    foreach (var group in Worksheet.SparklineGroups)
```
jahav: Move SparklineGroup shift to a method, here and in the other places.

jahav (commented):
I fixed many things along the review, kept few reminders. 
`XLRangeBase` will need more TLC.

ClosedXML/Excel/Columns/XLColumnCollection.cs
jahav: `XLColumnCollection.MaxColumn` is no longer needed and was removed.

ClosedXML/Excel/Ranges/XLRangeBase.cs
jahav: Inserting and shifting of whole columns is rather inconsistent across the codebase. In some places, the column is shifted before this method is called and thus the we can't shift column properties again. In others, the code relies on this method.

Should be consistent, but one small steps at a time.

ClosedXML/Excel/Style/XLStylizedBase.cs
jahav: Will do later. That requires changes to stylized code structure that don't belong to this PR.

PR #2103, opened 2023-06-11
Compatibility with Linux using Sixlabors.ImageSharp

MohammadR-prog:
Compatibility with Linux using Sixlabors.ImageSharp against system.drawing.
From .NET 5.0 onwards, Linux does not support system.drawing, so it is not possible to use ClosedXml in Linux either. With these changes, this problem is solved and ClosedXml becomes cross-platform.

### DIFF
--- a/ClosedXML.Examples/ClosedXML.Examples.csproj
+++ b/ClosedXML.Examples/ClosedXML.Examples.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1;net461</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1;net461;net6.0</TargetFrameworks>
     <OutputType>Exe</OutputType>
   </PropertyGroup>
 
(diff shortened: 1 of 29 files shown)


jahav:
To be more precise the nuget package System.Drawing.Common (which as been removed ~year ago in 0.97). The rest of System.Drawing namespace (Point, Rect and so on), which is the only thing remaining to be used, are part of .NET Core and the other frameworks.

It's unlikely I am going to merge this. I took significant effort to not depend on `Sixlabors.ImageSharp`, licensing issues and lack of vector image support being most prominent. Recently, they also dropped support for netstandard2.0.

ClosedXML works on Linux, at least it did the last time I checked it and there are several issues that use Linux as an OS. If it doesn't work, provide reproducible scenario in an issue.

MohammadR-prog:
I had a project with .net 3.1 and recently upgraded it to .net 6.0 and I am also using ubuntu server.
When I was using .net 3.1 and closedxml on Linux, everything was ok, but since I upgraded, I was getting a system.drawing error in the sections where I was using ClosedXml.
I made these changes and took the output from it and used it in the project and the problem was solved.

Also I used version 2.3.1 for .netstandard and 3.0.1 for .net 6.0.

However, it may be a mistake on my part, I will check again.

I am very interested in ClosedXml and I wish you the best of luck.

PR #2111, opened 2023-06-23
Add IFS formula support

aspriddell:
Adds [_xlfn.IFS](https://support.microsoft.com/en-gb/office/ifs-function-36329a26-37b2-467c-972b-4a39bd951d45) support, along with tests and updated documentation.

I'm not 100% sure on the kind of error should be returned on the incorrect number of arguments provided as Excel blocks the function from being entered...


igitur:
I also don't know what happens if you use the wrong number of arguments, but try generating a file using ClosedXML with e.g. 3 arguments, save it, open it in Excel and let us know what Excel shows.

jahav:
Generally for built-in defined functions, excel refuses to load and repairs a sheet. The number of arguments for inbuild functions is required to be correct by a parser. In this case [MS-XLSX], 23.3
```
ifs-params = "(" argument "," argument *126(","argument "," argument) ")"
```

aspriddell:
Creating a sheet with an invalid number of params results in `#NAME?` being printed in the cell:
image

I'll update the initial check to match this behaviour

(Tested using Excel Version 2305 Build 16501.20228 so it might behave differently on older versions)

jahav:
@aspriddell  You have used incorrect name in the test. The name of a function must have prefix `_xlfn`. The `#NAME` in a screnshot indicates function not found, because you used `IFS` and not `_xlfn.IFS`. The name of a function displayed in Excel is `IFS`, but that is display only. The real name used inside excel is `_xlfn.IFS`. The `_xlfn` is a prefix used for future functions of excel.

```csharp
var ws = wb.AddWorksheet();
ws.Cell(1,1).FormulaA1 = "_xlfn.IFS(FALSE, \"A\", TRUE)";
wb.SaveAs(@"c:\temp\pulls\2111\test.xlsx");
```
image

aspriddell:
Apologies, my bad. So what error should be returned in that case?

jahav:
@aspriddell I would go with `#VALUE!`. 

ISO29500 18.17.7 says
> When the type of an argument passed to a function is incompatible with the type expected the error value
#VALUE! is returned by that function.

Type is not the same as number, but it fits reasonably well. We could also throw an exception, but I *really* don't want to throw exceptions from calc engine.

Incorrect number of arguments should be caught much earlier, e.g. when the function is being set and solve it some way there (from parameter truncation through an exception to user defined function what to do passed `LoadOptions`).

aspriddell:
> I would go with `#VALUE!`.

Sounds good, I'll redo the changes now.

igitur:
What does Libreoffice do with that file?

aspriddell:
LibreOffice returns `#N/A` when using the file generated using the script in [jahav's comment](https://github.com/ClosedXML/ClosedXML/pull/2111#issuecomment-1605553183)

image

igitur:
I'm leaning to also returning #NA, but let's hear what @jahav says.

aspriddell:
Closing due to inactivity

PR #2119, opened 2023-07-04
Bump openxml sdk

rstm-sf:


rstm-sf:
@jahav ping https://github.com/ClosedXML/ClosedXML/pull/2004#issuecomment-1483952395

igitur:
Why is this important when users can manually upgrade to a higher version of OpenXML if they want? @Pankraty, you've made some valid points about this in the past?

twsouthwick:
We recently released a v3.0 that requires dependencies to be recompiled against it. Consumers will not be able to upgrade to it without an update to ClosedXML (see https://github.com/dotnet/Open-XML-SDK/issues/1608)

mariojsnunes:
Can this go up to 3.0.0?
Seems related to https://github.com/ClosedXML/ClosedXML/issues/2220

Issue #2133, opened 2023-07-20
the StringCompatibilityExtensions.Contains polyfill should not be part of the public API for ClosedXML


Applesauce314:
StringCompatibilityExtensions added in  #2082 has the added extension public, which is colliding with the polyfill extension method that already exists in our project.
this method should probably be made internal.

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [ ] Yes
- [x] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.102.0

**What is the current behavior?**

`string.Contains` is public.

**What is the expected behavior or new feature?**

`string.Contains` is internal and package users can polyfill this overload themselves.

**Is this a regression from the previous version?**

yes, causes
 `CS0121	The call is ambiguous between the following methods or properties:  'System.StringCompatibilityExtensions.Contains(string, char)' and 'System.MyStringCompatiblityExtensions.Contains(string, char)'`

## Reproducibility

**Code to reproduce problem:**
```c#
namespace MyProgram
{
    public void Main()
    {
        string x = "";
        Console.WriteLine(x.Contains('x'));
    }
}
namespace System
{
    internal class MyStringCompatiblityExtensions
    {
#if !NETCOREAPP2_1_OR_GREATER
        /// <summary>
        /// polyfill for contains(char value) added in netcore
        /// </summary>
        /// <param name="input">the string to search</param>
        /// <param name="value">the char to search for</param>
        /// <returns></returns>
        internal static bool Contains(this string input, char value)
        {
            return input.Contains(value.ToString());
        }
#endif
    }
}
```

- [ ] I attached a sample spreadsheet.  (You can drag files on to this issue)

ojalkumar:
@jahav What is the expected timeline for this bug to be fixed. I am getting same problem while using this library with version [v0.102.2]. Please mention the version in which this issue will be patched?

I have tested with [0.103.0-beta] version and I think there is no problem in this version but I am getting an error which says
"error CS8002: Referenced assembly 'ClosedXML, Version=0.103.0.0, Culture=neutral, PublicKeyToken=null' does not have a strong name." Do you have any other working alternative to handle this problem?

PR #2134, opened 2023-07-20
update accessibility of string.Contains(char) polyfill to internal

Applesauce314:
addresses #2133


jahav:
Fixes #2133 

Thanks, I will probably use one of existing netstandard polyfills sooner or later, but it shouldn't make a mess in other projects.

PR #2138, opened 2023-07-31
Replace XLParser with ClosedParser.

jahav:
It is much faster (1 mil formulas parsed under 2 seconds.), optimized for AST instead of CST and can handle A1 and R1C1 reference styles. There is a simple visualizer https://parser.closedxml.io/

It's not perfect and needs some TLC, but is good enough (it parses whole Enron and Euses dataset plus the tests) and it's time to do something else.


igitur:
Awesome work, @jahav

PR #2153, opened 2023-08-27
Fix ROUND (and probably others) not handling the result of binary operations on refs

sbeca:
While testing the latest HEAD of ClosedXML today, I noticed that I was getting some `System.InvalidCastException: Object must implement IConvertible` exceptions.

I tracked it down to a basic calculation in my source Excel file that looked this this: `=ROUND((H18*H$16+M18*M$16)/(H$16+M$16),0)`

This didn't seem like a function call that should be too difficult or cause issues, so after diving into the ClosedXML code, I noticed that the binary operations on the references within the `ROUND` were returning 1x1 Arrays and when the `ROUND` function tried to convert that to a `double` it would throw the exception.

This PR updates the binary operation code to return a scalar instead of an array, when the binary operation is operating on 2 references that only reference 1 cell each. (EDIT: I also found the same problem with unary operations and fixed that as well)

You might have another way of fixing this issue (e.g. maybe adding an implicit conversion from a 1x1 Array to a double would be better?) and not accept my PR, which is all good with me. But this issue not triggering any test failures and just throwing exceptions at runtime doesn't seem ideal so taking my basic test and adding it to the test suite so this gets captured in the future would be awesome

### DIFF
--- a/ClosedXML.Tests/Excel/CalcEngine/MathTrigTests.cs
+++ b/ClosedXML.Tests/Excel/CalcEngine/MathTrigTests.cs
@@ -1419,10 +1419,10 @@ namespace ClosedXML.Tests.Excel.CalcEngine
 
             // References are treated differently to constant values within calculations by the calc engine,
             // so let's make sure to include a test to validate that everything keeps working in the future
-            ws.Cell("A8").FormulaA1 = "ROUND((A1*A$2+A3*A$4)/(A$5+A$6),0)"; // (1 * 2 + 3 * 4) / (5 + 3) = 1.75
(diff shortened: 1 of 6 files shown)


jahav (approved):
@sbeca Thanks for the PR. There is no real handbook how should the calculation engine work, so I generally figure it out when something doesn't work. 

I originally believed that calculation of binary operations was based on somethin I call scalar/collection semantic - the calculation should depend on a types of arguments. Later, I discovered single/multi semantic (see `TryPickSingleOrMultiValue`) that seems to be more in-line with how Excel works and should be used instead (I hope).

Of course, this may turn out to be another mistake. I have originally thought that non-element array and non-single cell area reference are always same, but `ISEVEN`/`ISODD` proven it is not always so (I really hope that it's just a bug in a function implementation).

Anyway, at least code seems to be cleaner.

sbeca:
Sounds reasonable. Thank you for the PR review and making further improvements to it

PR #2157, opened 2023-09-02
Fix incorrect logic check in Array Rescale

sbeca:
While tracking down an unexpected result from an Excel calculation I was testing, I came across what looks like a copy-paste mistake that was causing my issue. This PR fixes the bug and adds a basic test for my issue that was failing before but now passes after this fix.


jahav (approved):
LGTM. Thank you for the fix.

PR #2162, opened 2023-09-09
Fix precedent cells incorrectly accepting file references

sbeca:
Today I came across a bug that caused a stack overflow and crashed my app. One of my test spreadsheets had a cell `'10.Livestock'!X122` with a formula of `=-'[1]10.Livestock'!X122` and with the latest code of ClosedXML from Git, this caused my crash.

I know that ClosedXML doesn't currently support file references, and has previously rejected them with a `NotImplementedException`. With latest HEAD though, the new `PrecedentCells` code was ignoring the file reference prefix entirely, resulting in the cell `'10.Livestock'!X122` basically looking like it had formula of `=-'10.Livestock'!X122`, at least for the code that works out the precedent cells. With a certain setup where the value is out of date, this can cause the precedent cell code to get into an infinite recursion loop, overflowing the stack and crashing the app.

This PR updates the precedent cell code to use a different method of getting the worksheet which checks for file references and throws an exception if detected, which matches up with what the calculation code does. It also adds a new test to make sure that file references are always rejected.

### DIFF
--- a/ClosedXML.Tests/Excel/CalcEngine/PrecedentCellsTests.cs
+++ b/ClosedXML.Tests/Excel/CalcEngine/PrecedentCellsTests.cs
@@ -1,5 +1,6 @@
 ﻿using ClosedXML.Excel;
 using NUnit.Framework;
+using System;
 using System.IO;
 using System.Linq;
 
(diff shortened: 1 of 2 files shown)


jahav:
@sbeca I am currently working to replace the the current recursive dependency resolution with standard (see [Excel Recalculation](https://learn.microsoft.com/en-us/office/client-developer/excel/excel-recalculation)) dependency tree and calculation chain (#2155, #2156, #2160) and it's a [blocker](https://github.com/ClosedXML/ClosedXML/discussions/2112) for releasing 0.103. 

I definitely have to add tests to new precedent visitor, thanks.

Since the `XLCalcEngine` will be deleted that will be deleted in the before 0.103 is released, I don't think it makes sense to merge the PR and I am closing it 🙇.

PR #2174, opened 2023-09-28
Load cells from workbook using a streaming

jahav:
Originally, rows of a workbook were loaded into a memory as a mini-DOM and then converted to internal structures.

This PR changes loading code, so it doesn't create mini-DOM for each row, but instead streams each element.

It has modest improvements for a speed 
* load of 250k rows, 15 columns (10 text, 5 numbers) has decreased from 20.5s (0.102.1) to 17.5s.
* For 250k x 15 columns of numbers, it decreased from 17.8s  to 14.6s
* For comparison, Excel loads the file in 5 - 6 seconds.

Honestly, I hoped for more, but it there are no easy pickings anymore. I think replacing parser would bring improvements (not very optimal, lot of checking, doesn't use spans...), but that is just not realistic.

*dotTrace* flame graphs:

image
image

[large_file_250_000-numbers.xlsx](https://github.com/ClosedXML/ClosedXML/files/12744266/large_file_250_000-numbers.xlsx)
[large_file_250_000.xlsx](https://github.com/ClosedXML/ClosedXML/files/12744267/large_file_250_000.xlsx)

### DIFF
--- a/ClosedXML/Excel/Cells/ValueSlice.cs
+++ b/ClosedXML/Excel/Cells/ValueSlice.cs
@@ -237,7 +237,7 @@ namespace ClosedXML.Excel
             }
         }
 
-        private readonly struct XLValueSliceContent
+        private readonly struct XLValueSliceContent : IEquatable<XLValueSliceContent>
         {
             /// <summary>
             /// A cell value in a very compact representation. The value is interpreted depending on a type.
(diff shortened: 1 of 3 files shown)


jahav:
I added custom equals method to structs for cell value. Originally, it used RTTI to calculate equality (different from `readonly record struct`s that do have better equality logic baked in). It gave me another 10%, so roughly 15-20% improvement over 0.102.1.

I am pretty sure the rest is mostly XML parser. Not sure why so slow, but if a checking property takes up roughly as much time as setting a value, it's problematic. Well, all default XML parser are problematic because of lack of `Span` and heap allocation everywhere, but still....

image

Issue #2130, opened 2023-07-19
Pivot table - support groups fields


ZephyrZiggurat:
**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [ ] Yes
- [x] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

I cloned and ran locally via ClosedXML.Sandbox.

**Version of ClosedXML**

0.102.0

**What is the current behavior?**

Exception during load of a vendor provided spreadsheet.
```
System.ArgumentOutOfRangeException: The column 'xyz' does not appear in the source range. (Parameter 'sourceName')
   at ClosedXML.Excel.XLPivotFields.Add(String sourceName, String customName)
   at ClosedXML.Excel.XLPivotFields.Add(String sourceName)
   at ClosedXML.Excel.XLWorkbook.LoadSpreadsheetDocument(SpreadsheetDocument dSpreadsheet)
   at ClosedXML.Excel.XLWorkbook.LoadSheets(Stream stream)
   at ClosedXML.Excel.XLWorkbook.Load(Stream stream)
   at ClosedXML.Excel.XLWorkbook..ctor(Stream stream, LoadOptions loadOptions)
   at ClosedXML.Excel.XLWorkbook..ctor(Stream stream)
   ...
```

**What is the expected behavior or new feature?**

Spreadsheet is loaded.

**Is this a regression from the previous version?**

Yes. Works in 0.101.0

## Reproducibility

**Code to reproduce problem:**
The exception happens during the load of an existing file with a lot of pivot tables and connections. I'm sorry but a code sample isnt really helpful here.
```
new XLWorkbook("file")
```

- [ ] I attached a sample spreadsheet.  (You can drag files on to this issue)

I dont have enough Excel experience to recreate this issue myself and I cant attach the file from the vendor.

---

I did spend a bit troubleshooting in the code and noticed a couple things.

At the previous commit, there was a check that the field is available before trying to add it in several places:

https://github.com/ClosedXML/ClosedXML/blob/69e37ad5f9d936c67d259e46582374930a23dd5f/ClosedXML/Excel/XLWorkbook_Load.cs#L714-L720

I assume this is what avoided the exception that is currently being thrown in `XLPivotFields.Add`.

---

[...]

PR #2178, opened 2023-10-01
Load pivot records

jahav:
Pivot table needs 
* the pivot table itself - handles how to display pivot table cache in the sheet
* pivot table definition - contains what fields are available and some precomputed characteristics for each field, as well as shared item that are used by the pivot table
* pivot cache records - the actual values for pivot table, (duplicated)

The pivot cache records weren't loaded and it is a prerequisite for any further pivot table development.

Honestly, this is kind of brownfield development, it will require several additional PRs to clean up, but the end result should be that it will be possible to load and save pivot cache (both records and table definition) without losing data.

This PR
* Moves logic for reading pivot tables to a separate reader
* adds structures to hold both shared items and field record values
* loads the record values to pivot cache

It doesn't write them back to output yet.

Related to #2130


artsai:
Hi @jahav,
https://github.com/ClosedXML/ClosedXML/issues/2130#issuecomment-1751790757, This fix only exists on 0.104.0-preview2, will there be a plan to add this to 0.102.3 or I have to wait for 0.104.0 version?

PR #2188, opened 2023-10-09
Enable nullable in a few more classes

sbeca:
I'm looking to add some extra functionality to `XLWorksheet` that I need for my project and I wanted the safety net of the `nullable` feature, so I enabled it for `XLWorksheet` and some related classes. I though I might as well upstream the changes in case they're useful.

### DIFF
--- a/ClosedXML.Tests/Excel/NamedRanges/NamedRangesTests.cs
+++ b/ClosedXML.Tests/Excel/NamedRanges/NamedRangesTests.cs
@@ -610,7 +610,7 @@ namespace ClosedXML.Tests.Excel
             Assert.IsFalse(ws.NamedRanges.Contains("NameX"));
 
             Assert.IsNotNull(ws.NamedRange("Name"));
-            Assert.IsNull(ws.NamedRange("NameX"));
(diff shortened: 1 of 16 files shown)


jahav:
I am painfully aware that public API is ~3k methods with very inconsistent and undocumented behavior.

Adding nullability should always be combined with at least cursory API review, because it exposes this inconsistency to the consumers of the package, e.g. `ws.Cell(1, 20000)` will throw while `ws.Cell("ZZZ1")` will return `null`.

Just marking things nullable as they are is not a great strategy, because consumers that use it suddenly get nullable warnings on things that should throw an exception (or aren't actually null in first place, e.g. `IXLWorkshet.SelectedRanges`) and unless user asks for invalid values (considering XML documentation is also pretty much missing), this is blissfully left ambiguous because there is no nullability analysis because of missing context e.g. in this case
```
ws.Cell("A1").Value = "Sales 2023";
ws.Cell("A7").Style.Font.Bold = true;
```

Of course, other parts of code rely on some methods returning null, so things get rather complicated.

Keep it in mind please.

sbeca:
Thank you for your time and effort on improving this PR. I will definitely keep your points in mind if I ever touch nullability again 🙂

PR #2194, opened 2023-10-14
Update language level and use polyfill for missing functions.

jahav:
The Polyfill library is a development only dependency that has *.cs sources in the .nuget directory. The *.cs files are conditionaly compiled along the project depending on the target and thus more modern version use less polyfills.

The main benefit is ability to use struct records (that require IsExternalInit attribute for init properties), though generally it just allows a modern development even for netstandard2.0 and nullability analysis (e.g. `NotNullWhenAttribute` for `TryGet` pattern).

The polyfills are internal and thus don't pollute applications consuming ClosedXML.

### DIFF
--- a/ClosedXML/ClosedXML.csproj
+++ b/ClosedXML/ClosedXML.csproj
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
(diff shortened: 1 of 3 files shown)


jahav:
Many thanks to @SimonCropp for his work on all-in-one [Polyfill](https://github.com/SimonCropp/Polyfill) package.

PR #2201, opened 2023-10-23
Make XLCell GetFormat public

sbeca:
Very small PR: just makes `XLCell.GetFormat` public, and exposes it within the `IXLCell` interface.

This is useful in cases like mine where I have a web backend that handles reading of spreadsheets and calculating cell values, but I want to leave the formatting of the values to a web frontend for speed or culture-specific formatting reasons. In cases like that, I need to be able to read the format value in the backend to pass it to the frontend.

### DIFF
--- a/ClosedXML/Excel/Cells/IXLCell.cs
+++ b/ClosedXML/Excel/Cells/IXLCell.cs
@@ -378,6 +378,11 @@ namespace ClosedXML.Excel
         String GetFormattedString(CultureInfo culture = null);
 
         /// <summary>
+        /// Gets the format string used by <see cref="GetFormattedString"/> to format and render the numeric value of a cell.
+        /// </summary>
+        String GetFormat();
+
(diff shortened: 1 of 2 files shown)


jahav:
I am going to close it.

There are two reasons:
1. Some predefined formats can differ across culture. Predefined format are baked into the excel, except formats 14 and 22 (e.g. the short date (14 format) displays numbers using the format `dd.MM.yyyy` in cs-CZ, but as `M/d/yyyy` in en-US region). Therefore the results wouldn't even be correct. Granted, the value formatting doesn't really work correctly even now, but I don't want to make add API endpoint that is immediately buggy.
2. I don't want to add API, it's another thing to maintain. This can be done through extension method (take number format from `IXLCell.Style.NumberFormat`) plus copying the dictionary of formats.

I should have closed this long time ago instead of letting it linger.

PR #2218, opened 2023-11-20
Fixed NullReferenceException at loading workbook with empty si element

psynomorph:
Hi!

I recently came across an xlsx workbook that opens in Excel, but does not open by ClosedXml. Feature of the workbook - empty <si> element in sharedStrings.xml. It cause NullReferenceException at loading cell value.
This small pull request contains fix.

Example of workbook - [XlsxWithEmptySi.xlsx](https://github.com/ClosedXML/ClosedXML/files/13415438/XlsxWithEmptySi.xlsx)


jahav:
LGTM.

PR #2225, opened 2023-12-08
Update index.rst: Fixed the `using` code structure

francios-de:


jahav:
The PR only changes `using` declaration to `using` statement. The `using` declaration It allows for shorter snippet code without clutter and is prefered way for the documentation.

It has been available since C#8 and can be used with .NET framework (easily with SDK style projects, likely feasible with old msbuild-style projects).

As such, I am closing the PR without merging.

PR #2231, opened 2023-12-15
Fix InvalidCastException at workbook loading

lvxiao312:
I encountered this issue when trying to load a workbook containing chart sheets.


jahav:
LGTM

PR #2248, opened 2024-01-05
Update OpenXML SDK to 3.0.

jahav:
Quick update of source code to reflect changes in OpenXML SDK 3.0. It requires net462 and `System.IO.Packaging` 8.0.0.

One test fails (`SavingTests.SavesAsProperSpreadsheetDocumentType`) which was caused by https://github.com/dotnet/Open-XML-SDK/issues/1615. That should be fixed in 3.1 that should be done soon, only [one issue remains for 3.1 milestone](https://github.com/dotnet/Open-XML-SDK/milestone/10).


twsouthwick:
I should have a v3.0.1 with the fix available in a couple of days (hopefully tomorrow, but to set expectations a few days).

Let me know if you find any other issues with v3 as you roll it out - I'll try to address any regressions :)

twsouthwick:
v3.0.1 is available on nuget with the requested fix - please give it a try

twsouthwick:
>  It requires net462 and System.IO.Packaging 8.0.0.

The supported TFMs are still the same. We support .NET Framework 3.5+, .NET Standard 2.0, .NET Core/5+

twsouthwick:
You may be seeing a change in the support of System.IO.Packaging; however, on framework it never actually did anything and type forwarded to WindowsBase.dll

jahav:
@twsouthwick Thanks! Great work. 

Test passes. I am looking forward to leverage the new `IPackage` and I think I there were few malformed relationship URIs bugs over the years, so that will be really nice to check out too.

PR #2267, opened 2024-01-18
Remove outdated transitive package upgrades

### LINKED ISSUE #2265: Remove references to "System.Net.Http" and "System.Text.RegularExpressions"
- [x] Bug
- [ ] Feature
- [ ] Question

- [x] Yes
- [ ] No

**Version of ClosedXML**
`0.104.0-preview2`

Because of [this commit](https://github.com/ClosedXML/ClosedXML/commit/ce6ce6a11324dbbe25ccdc209520d66893b3d418#diff-1e6758ed2a0ed27e66cb0c73f876f6909cf3552bda17d548b01aa3f0058a8188) the dependency on [`XLParser v1.5.2`](https://www.nuget.org/packages/XLParser/1.5.2#dependencies-body-tab) has been removed. This package brought in the `NETStandard.Library v1.6`, which in turn brought in the (vulnerable versions of) `System.Net.Http` and `System.Text.RegularExpressions`.

Now that `XLParser` is gone, these explicit upgrades of these packages can also be safely removed from `ClosedXML`. 

No more dependency on `System.Net.Http v4.3.4` and `System.Text.RegularExpressions v4.3.1`.

No

Arjan321:
Fixes #2265

### DIFF
--- a/ClosedXML/ClosedXML.csproj
+++ b/ClosedXML/ClosedXML.csproj
@@ -16,8 +16,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <!-- NU5104 - SixLabors.Fonts is available only as a prerelease -->
-    <NoWarn>$(NoWarn);NU1605;NU5104;CS1591</NoWarn>
+    <NoWarn>$(NoWarn);NU1605;CS1591</NoWarn>
   </PropertyGroup>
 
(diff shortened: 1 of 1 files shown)


jahav:
LGTM. I have also updated Fody, because it depended on netstandard 1.4.

PR #2284, opened 2024-02-04
Optimise workbook loading by stopping unneeded invalidation

sbeca:
I'm working on an app that uses ClosedXML, and we've recently started testing it on some large spreadsheets, e.g. 1 test spreadsheet that I've been profiling is 27MB, has 132 worksheets, and probably has thousands of calculations.

These larger spreadsheets have been taking over 10 seconds to load in release builds (time of just doing a `new XLWorkbook(workbookFilePath)`), which we would like to get lower if possible, so I've been doing some profiling to see if there are any easy wins. First cab off the rank is this:

CpuUsage1

Having 23% of the time being spent just marking formulas as dirty seems like something that could have had some easy wins, so I dug deeper. This percent is from a debug build, and the savings in release builds turned out to be a decent amount smaller, but it is still worth improving.

The crux of the issue here is that every single time that a sheet is added, the calculations of all the previous sheets that have already been added get marked as dirty. In my test workbook, this means that this mark dirty pass gets run 132 times (although the first time it does nothing because nothing has been loaded yet). Also, as it appears to process all added calculations so far each time it runs, the amount of work done might grow exponentially, as each time it runs there might be more and more calculations to enumerate over to mark as dirty.

One potential solution here would be to delay all this mark dirty work until after all the worksheets have been loaded, and then only do it once at the end. However, I think that this also exposes a logic mistake, so this hasn't been what I've done here.

The current load code, after it was added the new worksheet, has the following code that it runs when loading individual cells of a worksheet (in `LoadCell`):

```csharp
// If the cell doesn't contain value, we should invalidate it, otherwise rely on the stored value.
// The value is likely more reliable. It should be set when cellFormula.CalculateCell is set or
// when value is missing. Formula can be null in some cases, e.g. slave cells of array formula.
if (formula is not null && !cellHasValue)
{
    formula.IsDirty = true;
}
```

This code makes it look like the original intent of the loading code was that marking formulas as dirty was only done when needed, and at some point a change was made and now all formulas always all get marked as dirty when loaded. Or, I should be a bit more correct and say that the current logic always marks all formulas dirty for all worksheets except the last worksheet loaded. This means you wouldn't notice this issue at all if your workbook only has 1 worksheet. But in any case, it looks like this current logic likely wasn't intended.

This PR has one potential solution, where through some new method overloads, `NotifyWorksheetAdded` is not called when loading a workbook from a file, but is still called when modifying a workbook after loading or after creating a new workbook in memory.

I might be missing some historical context, or maybe there is a better way to achieve this fix, so I'm happy if this gets fixed/improved another way.

In my testing this took my load times in release builds from about 16 seconds for my test spreadsheet, down to 15 seconds, or about a 6% reduction. That might not sound like a lot, but every bit helps, and this also likely fixes a logic mistake if I'm reading things correctly.

This fix could likely have further savings after the load as this might allow for better utilisation of cached calculation values written to files, whereas at the moment, most calculations in multi-worksheet workbooks will be marked as dirty, likely leading to the cached calculation values written to files not being possible to be used. I haven't looked into this though, as I'm focused on just loading of the workbook for now, so I might be wrong.

### DIFF
--- a/ClosedXML/Excel/XLWorkbook_Load.cs
+++ b/ClosedXML/Excel/XLWorkbook_Load.cs
@@ -203,13 +203,10 @@ namespace ClosedXML.Excel
                 ColumnWidth = CalculateColumnWidth(8, Style.Font, this);
             }
 
-            // We don't want to trigger a NotifyWorksheetAdded on the workbook every time we add a new worksheet during
-            // the loading of the workbook, as that invalidates all the cached caculation values unnecessarily and can
-            // cause a large amount of time to be wasted during workbook loading. Large workbooks with many worksheets
-            // and many caculations were seeing 5-15% of load time being spent just being used marking calculations as
-            // dirty over and over again before this was disabled.
-            const bool notifyWorksheetsAdded = false;
-
+            // We loop through the sheets in 2 passes: first just to add the sheets and second to add all the data for the sheets.
+            // We do this mainly because it skips a very costly calculation invalidation step, but it also make things more consistent,
+            // e.g. when reading calculations that reference other sheets, we know that those sheets always already exist.
+            // That consistency point isn't required yet but could be taken advantage of in the future.
             var sheets = workbookPart.Workbook.Sheets;
             Int32 position = 0;
             foreach (var dSheet in sheets.OfType<Sheet>())
@@ -221,7 +218,7 @@ namespace ClosedXML.Excel
                 if (string.IsNullOrEmpty(dSheet.Id))
                 {
                     // Some non-Excel producers create sheets with empty relId.
-                    var emptySheet = WorksheetsInternal.Add(sheetName, position, sheetId, notifyWorksheetsAdded);
+                    var emptySheet = WorksheetsInternal.Add(sheetName, position, sheetId);
                     if (dSheet.State != null)
                         emptySheet.Visibility = dSheet.State.Value.ToClosedXml();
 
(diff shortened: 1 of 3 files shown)


ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -202,6 +202,13 @@ private void LoadSpreadsheetDocument(SpreadsheetDocument dSpreadsheet)
                 ColumnWidth = CalculateColumnWidth(8, Style.Font, this);
             }
 
+            // We don't want to trigger a NotifyWorksheetAdded on the workbook every time we add a new worksheet during
+            // the loading of the workbook, as that invalidates all the cached caculation values unnecessarily and can
+            // cause a large amount of time to be wasted during workbook loading. Large workbooks with many worksheets
+            // and many caculations were seeing 5-15% of load time being spent just being used marking calculations as
+            // dirty over and over again before this was disabled.
+            const bool notifyWorksheetsAdded = false;
```
sbeca: This could have just been an inline `false` passed into the 2 calls to `Add` below, but I wanted somewhere shared to add my explaining comment for future devs, so hoisted it into a variable

jahav (changes requested):
@sbeca 
The root ause is one of the long-term key problems that I am slowly trying to fix: Keep workbook consistent through structure changes.

That basically means, if I delete a modify a portion of workbook, it should be updated to reflect that in a consistent manner (e.g. formulas should be updated, merge). As of now, that is mostly done through magic/unified ranges (this si the core reason why any range has be kept track of, otherwise the magic can't find all ranges).

I am not a fan of that, because it leads to invalid model and can't support various changes (e.g. if I delete few rows, the calculation chain should remove some links or shift addresses). My goal is that various objects in a workbook (e.g. table, merged range, pivot table, calculation chain, picture...) that need to be updated will be more "self-contained" and when a structural change happens, I just call a object and it should take care about changing it's internal state *independently* (or possibly reject it altogether, e.g. it shouldn't be possible to delete a row that contains a header of a table).

This of course isn't flashed out, but it is a general idea of what I want to do.

In 0.103, the calculation of formulas has been changed from recursive evaluation to calculation chain with dependency tracking. Calculation chain tracks dependencies and if there isn't a sheet, it can't track it. So when sheet is added, it checks whether there a formula (e.g. `=NotYetLoadedSheet!A5 * 2`) and if there was, it adds tracking to a dependency tree. Similar thing should happen on deletion.

Because switch to calc chain was a lot of work and I didn't have time to properly implement efficient reaction to structural changes (plus ClosedParser didn't have a way to check if a sheet is used in formula back then, it does now), I just deleted whole dependency tree and it will be recreated when needed instead of marking only fomrulas that actually need it. It also makes everything dirty, because any formula could depend on added sheet, thus potentially be now invalid (if I mark all as invalid, all potential dirty ones are for sure now dirty). That is basically what `XLCalcEngine.Purge` does in the trace you posted.

As for the original code that marks cells dirty
```
if (formula is not null && !cellHasValue)
{
    formula.IsDirty = true;
}
```

The idea is that when I load a file that contains values, I assume it's correct and nothing needs recalculation, unless value changes (or there is an array formula with AlwaysCalculateArray = true or formula contains volatile function like `NOW`/`RAND`). The fact that it is already marked as dirty is unintentional side-effect of formula dependency tracking.

Back to the potential solution: In this case, I think it would be better to first create all sheets as empty and only then fill them with data. The empty sheets won't invalidate any formula, because there are no formulas and once they are added, the invalidation won't happen. The workbook part contains element with all sheets and their names.

Probably move a portion of `foreach (var dSheet in sheets.OfType<Sheet>())` loop that adds sheets into another loop that happens before the current one.

sbeca:
@jahav: Thank you very much for the PR review and providing that background context.

I agree that your proposed solution of loading the sheets in 2 loops is better than my original solution, so I have updated this PR to do that instead.

I also added some extra nullable annotations while I was in the area to help improve the useability of the API slightly.

jahav (approved):

Issue #1686, opened 2021-06-29
Saving a document with solid color data bars changes it to gradient data bars


bchmura:
## Read and complete the full issue template

Do not randomly delete sections. They are here for a reason.

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.95.999.2326
0.95.4
Only checked back as far as 0.94.0 and it happens thru then also

**What is the current behavior?**

Create a spreadsheet in excel and add conditional formatting for a data bar.  Select SOLID (not gradient) for the color.

Save the spreadsheet.

Open and close using ClosedXML

Open in excel 

Data bars are now set to Gradient.

**What is the expected behavior or new feature?**

Data bars should remain solid

**Is this a regression from the previous version?**

If it is, its not recent... tested back to 0.94.0

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

The attached spreadsheet has solid data bars, use that with this code.  After, open it and they will be gradient.

**Code to reproduce problem:**
```c#
using System;
using ClosedXML.Excel;
namespace ClosedXMLTest
{
    class Program
    {
        static void Main(string[] args)
        {
            
            var wb = new XLWorkbook(@"../../../../data/DataBarBug.xlsx");
            wb.Save();

        }
    }
}```
[DataBarBug.xlsx](https://github.com/ClosedXML/ClosedXML/files/6729868/DataBarBug.xlsx)

- [X ] I attached a sample spreadsheet.  (You can drag files on to this issue)

bchmura:
[DataBarBug.xlsx](https://github.com/ClosedXML/ClosedXML/files/6729878/DataBarBug.xlsx)

PR #2296, opened 2024-02-09
Allow specifying gradients for DataBar conditional formatting

### LINKED ISSUE #386: Losing gradient filling format
- [x] Bug

I opened some sheets that have cells formated with gradient filling. Once I saved the workbook (even if I don't do anything else), the format is lost.

erenes:
Fixes #386, #1686

The dataBar element below the conditionalFormat supports a gradient property as of Excel 2010, which allows a solid fill on the conditional format. 

This PR allows specifying the gradient property when creating a dataBar programatically with ClosedXML, as well as support for loading and saving files that contain solid fill dataBars.

I've tried to adhere to the existing code style and ran CodeMaid on the files in the PR. I am happy to implement unit tests as well, but did not find any existing ones for the dataBar elements.

Example code: 
```csharp
var ws = workbook.AddWorksheet("test");
ws.FirstCell().SetValue(0.53).AddConditionalFormat()
    .DataBar(XLColor.Red, showBarOnly: false, gradient: false)
    .Minimum(XLCFContentType.Number, 0)
    .Maximum(XLCFContentType.Number, 1);
workbook.SaveAs("test.xlsx");
```


jahav (changes requested):
API has incorrect shape. It adds an optional parameter, which is not scalable for the rest of parameters  ([MS-XLSX]2.6.30 CT_DataBar has many other attributes).

The Gradient prop is on `IXLConditionalFormat`, where it doesn't belong. It is only for DataBar (possibly other, but only subset), not generic conditional formats.

ClosedXML uses fluent API for similar case (e.g. styles are fluent `something.Style.Font.SetSize(14).SetBold()`). 

There are two interfaces `IXLCFDataBarMax` and `IXLCFDataBarMin`. Since this thing hasn't gone through API review yet, let's just add `IXLCFDataBar` and chain it behind  `IXLCFDataBarMax` (currently returns void, to ensure that there are exactly two CFVO though Min and Max interface). 

```
interface IXLCFDataBar {
  bool Gradient {get;}
  IXLCFDataBar SetGradient(bool value = true);
}
```
Tests should be ClosedXml.Tests.ConditionalFormats directory. In this case, I think jut make a new ConditionalFormatDataBarTests. Test names should be `Explain_what_test_does()`. One test with TestHelper.CreateSaveLoadAssert should sufficie.  Test file should go to ClosedXML.Tests\Resource\Other\ConditionalFormats\. 

Add good descriptive XML documentation for added/modified methods/props.  Add user documentation for DataBar (see /doc directory). Any modification/new feature must include documentation.

Any breaking changes (in this case unification of interfaces) should be added to docs\migrations\migrate-to-0.104.rst.

erenes:
Thanks for the feedback, the direction was very helpful!

I have now implemented it as a Fluent API and added a unit test.
Addtionally I've also added the breaking changes to the migrate doc, and did my best to add useful documentation (I struggled a lot to come up with anything useful for such a trivial change)
 
Now that I've gone this far I think it's a logical next step to get rid of the `XLCFDataBarMin` and `XLCFDataBarMax` and unify all of those calls in the new `XLCFDataBar` object (maybe Min can still return the Max interface, so you are nudged towards not just setting one of the two). 
Let me know if you would want me to pick that up in this PR as well or to leave it as-is until there is an API review.

PR #1364, opened 2020-03-10
Fix issues occurred when running tests on .Net Core 3.0

Pankraty:
Fixes #1318

Fixes #1273

Fixes #1274 

One test is still failing due to andersnm/ExcelNumberFormat#23.

@igitur should we switch to `netcoreapp3.0` in tests or add it as a new target (the 4th one)?

PR #2318, opened 2024-03-25
Fix TimeSpan conversion

### LINKED ISSUE #1993: Regression bug: TimeSpan read/write inconsistency
- [x] Bug
- [ ] Feature
- [ ] Question

- [x] Yes
- [ ] No

**Version of ClosedXML**

0.100.3

Some TimeSpan values will have a slightly different value after reading them from the excel file.

The read TimeSpan value should have the same value as it was written in the excel file.

Yes, this is a regression as it works with version 0.97.0. The issue seems to be related to `ToSerialDateTime` and `ToSerialTimeSpan` methods which were added in version 0.100.0.

## Reproducibility

```c#
[Test]
public void TestTimeSpan()
{
    var expectedTime = new TimeSpan(11, 0, 0);
    var wb = new XLWorkbook();
    ws.Cell(1, 1).Value = expectedTime;
    wb.SaveAs("test.xlsx");

    using (var s = File.OpenRead("test.xlsx"))
    {
        var sWb = new XLWorkbook(s);
        sWb.TryGetWorksheet("Sheet1", out var sWs);
        var time = sWs.Cell(1, 1).GetTimeSpan();
        Assert.AreEqual(expectedTime, time);
    }
}
```

The above test fails with message:
```
Message: 
  Expected: 11:00:00
  But was:  10:59:59.9999999
```

OldBuddy:
closes #1993

reintroduce changes done by #1364 regarding TimeSpan conversion (broken by #1922).

### DIFF
--- a/ClosedXML.Tests/Excel/Cells/XLCellTests.cs
+++ b/ClosedXML.Tests/Excel/Cells/XLCellTests.cs
@@ -444,6 +444,15 @@ namespace ClosedXML.Tests
         }
 
         [Test]
+        public void TryGetValue_TimeSpan_Good2()
+        {
+            IXLWorksheet ws = new XLWorkbook().Worksheets.Add("Sheet1");
+            bool success = ws.Cell("A1").SetValue(0.0034722222222222199).TryGetValue(out TimeSpan outValue);
+            Assert.IsTrue(success);
+            Assert.AreEqual(TimeSpan.FromMinutes(5), outValue);
+        }
+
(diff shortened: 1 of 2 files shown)


jahav (approved):
LGTM. Thanks for the fix.

jahav:
The CI check failed on "502 unable to upload artifacts", so probably some problem with AppVeyor storage. Test were described as passing and checked locally. Github refused to rerun checks.

PR #2325, opened 2024-04-12
Avoid problems with System.InvalidCastException when opening file

tupubozu:
After building building a project with 0.102.2 on .NET 8, I tried opening an XLSX file (freshly converted form XLS) and I encountered an exception while trying to open this file:

System.InvalidCastException: Unable to cast object of type 'DocumentFormat.OpenXml.Packaging.DialogsheetPart' to type 'DocumentFormat.OpenXml.Packaging.WorksheetPart'.
at ClosedXML.Excel.XLWorkbook.Load(Stream stream)
at ClosedXML.Excel.XLWorkbook..ctor(Stream stream, LoadOptions loadOptions)
/* trace of my code from this point */

The 'as' operator shouldn't behave like this. However checking with 'is' before converting with 'as' should, in theory, ensure that 'as' will not(never?) fail as everything past the check 'is' a WorkbookPart. Nothing else in the code (LoadSpreadsheetDocument method), as far as I can find it, could potentially throw an InvalidCastException.

Should someone know something more about this, then I would like to know...

### DIFF
--- a/ClosedXML/Excel/XLWorkbook_Load.cs
+++ b/ClosedXML/Excel/XLWorkbook_Load.cs
@@ -224,12 +224,13 @@ namespace ClosedXML.Excel
                 // Although relationship to worksheet is most common, there can be other types
                 // than worksheet, e.g. chartSheet. Since we can't load them, add them to list
                 // of unsupported sheets and copy them when saving. See Codeplex #6932.
-                if (!(workbookPart.GetPartById(dSheet.Id) is WorksheetPart))
+                var part = workbookPart.GetPartById(dSheet.Id);
+                if (!(part is WorksheetPart))
                 {
                     UnsupportedSheets.Add(new UnsupportedSheet { SheetId = sheetId, Position = position });
                     continue;
                 }
-                var worksheetPart = workbookPart.GetPartById(dSheet.Id) as WorksheetPart;
+                var worksheetPart = part as WorksheetPart;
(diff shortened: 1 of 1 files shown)


jahav:
The changed place isn't where exception was thrown (at least in my test file [dialog.xlsx](https://github.com/ClosedXML/ClosedXML/files/15138478/dialog.xlsx)). This [XLWorkbook_Load.cs#L629](https://github.com/ClosedXML/ClosedXML/blob/0.102.2/ClosedXML/Excel/XLWorkbook_Load.cs#L629) just before loading of pivot tables.

```csharp
foreach (var dSheet in sheets.OfType<Sheet>())
{
  var worksheetPart = (WorksheetPart)dSpreadsheet.WorkbookPart.GetPartById(dSheet.Id);
```

This has been changed in 92f9fe2ab109f964e90420e8ca7377b4468ef78d when I split reading of pivot tables to separate class (it also threw on ChartsheetPart).

jahav:
Since the problem has been fixed in develop, closing. I have added test to ensure it doesn't happen in the future: #2334.

tupubozu:
I see. Thank you for telling me. It would seem like I still need to improve my sense for debugging/diagnostics. 

Am I correct in assuming the fix will be shipped in a future "major" release (0.103+), or will it be shipped as a patch/mainainence release?

PR #2329, opened 2024-04-20
Add EasyExport function for exporting object data to Excel

zanyar3:
Introduces EasyExport functionality tailored for seamless integration with ClosedXML. 
The new feature enables users to export object data to Excel files effortlessly using ClosedXML, enhancing the versatility and usability of the ClosedXML library.

### DIFF
--- a/ClosedXML.Tests/Extensions/ExportExtensionsTests.cs
+++ b/ClosedXML.Tests/Extensions/ExportExtensionsTests.cs
@@ -1,0 +1,81 @@
+﻿using System.Collections.Generic;
+using ClosedXML.Attributes;
+using ClosedXML.Excel;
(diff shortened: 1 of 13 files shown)


jahav:
Closing this without merging, because specific importing mechanisms are out of scope of this project. ClosedXML provides access to manipulate a workbook. Specific usages (in this case import with few settings) are out of scope.

I would suggest to make a nuget package that depends on ClosedXML. I don't see anything that needs to access to internals.

I would also recommend to use InsertData functions instead of setting cells individually, for anything larger, performance is a problem (setting each cell basically requires stuff like am I part of a table, am I part of merged range instead of just few checks to bulk data).

igitur:
Agree. Definitely out of scope.

Issue #1983, opened 2023-01-21
XLWorkbook cannot load the xlsx file which a sheet name does not have a whitespace and PageSetup.PrintAreas is added.


norainu98:
Do not randomly delete sections. They are here for a reason.

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

**Version of ClosedXML**

0.100.3
0.100.2
0.100.1
0.100.0
0.97.0
0.96.0

**What is the current behavior?**

XLWorkbook cannot load the xlsx file which a sheet name does not have a whitespace and **PageSetup.PrintAreas** is added.
An eception(There isn't a worksheet named '{SheetName}!{PrintRange},{SheetName}') occurs.

**What is the expected behavior or new feature?**

Even If a sheet name does not have a whitespace, it's possible to load a xlsx file which **PageSetup.PrintAreas** is added.

**Is this a regression from the previous version?**

0.95.4 does not have a problem. 
From 0.96.0 is occured a proble. 

## Reproducibility

```csheap
public static int Main(string[] args)
{
	// There is not a spaces in sheet name.-> An exception occur: System.ArgumentException: There isn't a worksheet named 'A!$A$1:$B$2,A'
	TestPrintArea("A");
	// There is some spaces in sheet name. -> An exception does not occur.
	TestPrintArea("A ");
	return 0;
}

private static void TestPrintArea(string sheetName)
{
	const string FileName = "printarea.xlsx";
	if (File.Exists(FileName))
	{
		File.Delete(FileName);
	}
	using (var workbook = new XLWorkbook())
	{
		var ws1 = workbook.Worksheets.Add(sheetName);
		ws1.PageSetup.PrintAreas.Add("A1:B2");
		ws1.PageSetup.PrintAreas.Add("D3:D5");
		workbook.SaveAs(FileName);
	}
	try
	{
		using (var workbook = new XLWorkbook(FileName))
		{
			Console.WriteLine($"'${sheetName}'\r\n -> SUCCESS");
		}
	}
	catch(Exception ex)
	{
		Console.WriteLine($"'${sheetName}'\r\n -> Exception occur. (${ex.Message})");
	}
}
```

:resut(v0.100.3)

```console
'$A'
 -> Exception occur. ($There isn't a worksheet named 'A!$A$1:$B$2,A'.)
'$A '
 -> SUCCESS
```

[...]

PR #2342, opened 2024-05-20
Added an option to not load defined names

adamscarlat:
In certain cases, the defined names section in an Excel workbook may contain broken links to ranges that no longer exist. While this doesn't prevent users from working with the Excel file directly in Excel, it poses a significant issue when loading these files with the ClosedXML library.

Currently, ClosedXML automatically attempts to load all defined names in the workbook. If it encounters any corrupt defined names or broken links, it throws an `ArgumentOutOfRangeException` exception (see `XLTable.Table`). This behavior prevents users from loading any Excel files with broken defined names, regardless of whether they need to use these for their processing or not.

This update addresses this issue by introducing a new option that allows users to skip the loading of defined names if they are not needed for their processing tasks. This enhancement ensures that users can still work with Excel files that have broken defined names without encountering exceptions.

Key points of this change include:

* Optional Defined Names Loading: Users can now choose to bypass the loading of defined names, which is particularly useful for files with broken links.
* Backward Compatibility: The change is fully backward compatible. By default, the option is set to true, meaning defined names will be loaded unless specified otherwise.
* Improved Flexibility: This new feature provides greater flexibility and robustness in handling a wider variety of Excel files, enhancing the overall user experience with ClosedXML.
This improvement ensures that users can process Excel files more reliably, even when encountering issues with defined names.

### DIFF
--- a/ClosedXML.Tests/Excel/Loading/LoadingTests.cs
+++ b/ClosedXML.Tests/Excel/Loading/LoadingTests.cs
@@ -410,9 +410,10 @@ namespace ClosedXML.Tests.Excel
             {
                 Assert.DoesNotThrow(() =>
                 {
-                    // The value in the file is blank and kept.
+                    // The value in the file is blank and kept. Defined names are loaded (defaulting to load defined names)
                     using var wb = new XLWorkbook(stream, new LoadOptions { RecalculateAllFormulas = false });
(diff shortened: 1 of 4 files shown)


fopsdev:
thanks!
this resolves:
https://github.com/ClosedXML/ClosedXML/issues/1983
for me

PR #2363, opened 2024-07-04
Eliminate a couple of performance killers

Pankraty:
Hi @jahav!

The guys from my team faced an issue with a form generated by ClosedXml.Report: it took **up to 20 minutes** to generate the report with ~64 columns and 500-600 rows.

In that report there were up to 45 columns being deleted after populating the data, and I found that this operation was the longest. This was caused by around 30K ranges in the range repository, most of them just one cell size (e.g. `AB11:AB11`). Further investigation showed that these ranges occur during copying the data from one worksheet to another.

In this PR I removed `AsRange()` from `HasDataValidation` getter, and excluded sparklines from cell-by-cell copying as they are copied in bulks. 

No behavior has changed; performance has increased drastically - up to 3-5 seconds for report generation.

Can we release a hotfix version with this change?


jahav:
@Pankraty Hotfix for which version?

This behavior of ranges (always kept in repo) is one of reasons why I want to slowly move away from this concept as a backend for objects to `ISheetListener` (to be improved in future). If every component can update itself independently, I could do several in parallel.

There are several other reasons, e.g. when range is changed to invalid (e.g. collapsed), the object it represents should be removed (in most cases). I am a proponent of always-valid model.

Pankraty:
Version 0.102.3, I think.

As for the ranges repository, it was designed this way to reduce the number of objects (ranges) kept in memory by removing duplicates, and also to get rid of events to which one had to subscribe/unsubscribe. Events made diagnostics way harder as it was nearly impossible to figure which of thousands of subscriptions lacks unsubscription thus making the range not GCed. With the old approach we would probably suffer not from 30K single-cell ranges but more likely 300K subscriptions to a single event 😆  

By using WeakReferences in the repository we can handle the case when the user code keeps references to ranges that also need to be updated; at the same time, we don't process ranges no one is referencing to. At least, that was the idea, and I was quite surprised that these `cell.AsRange()` caused so many troubles: in all synthetic tests they were almost immediately GCed and removed from the repository, while in real-life example they remain and needed attention.

Don't get me wrong, I am not defending this approach, it can be improved, for sure. Just trying to give some additional context on why it's been done this way.

jahav:
@Pankraty I will look at it tomorrow or day after.

jahav (approved):

PR #2380, opened 2024-08-16
Correctly load quoted and unquoted sheet names in print area definitions

### LINKED ISSUE #1983: XLWorkbook cannot load the xlsx file which a sheet name does not have a whitespace and PageSetup.PrintAreas is added.

igitur:
Fixes #1983

### DIFF
--- a/ClosedXML.Tests/Excel/PageSetup/PrintAreaTests.cs
+++ b/ClosedXML.Tests/Excel/PageSetup/PrintAreaTests.cs
@@ -1,0 +1,27 @@
+﻿using NUnit.Framework;
+using System.Linq;
+
+namespace ClosedXML.Tests.Excel
+{
+    [TestFixture]
+    public class PrintAreaTests
+    {
+        [Test]
+        [TestCase("A1:B2")]
+        [TestCase("A1:B2", "D3:D5")]
+        public void CanLoadWorksheetWithMultiplePrintAreas(params string[] printAreaRangeAddresses)
+        {
+            TestHelper.CreateSaveLoadAssert(
+                (_, ws) =>
+                {
+                    foreach (var printAreaRangeAddress in printAreaRangeAddresses)
+                        ws.PageSetup.PrintAreas.Add(printAreaRangeAddress);
+                },
+                (_, ws) =>
+                {
+                    var actualPrintAddresses = ws.PageSetup.PrintAreas.Select(pa => pa.RangeAddress.ToStringRelative());
+                    CollectionAssert.AreEqual(printAreaRangeAddresses, actualPrintAddresses);
+                });
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


jahav (approved):
The correct fix would be to use parser to parse defined name per grammar ([OI29500] 3.2.3.1.4 Name Formulas), but that would be really time-intensive.

This will work for now and has test to cover us in the future.

I modified the test, because I prefer to use helper methods for standard testing patterns instead of boilerplate code.

I should look into exposing API for parsing sheet references from ClosedXML.Parser. That would definitely bet better than regexp.

rominator1983:
We need that functionality (correctly parsing defined names) for a customer and have some code that needs to be prepared to be merged into develop.
The changes for some very old version can be found in this diff:
https://github.com/rubicon-oss/DocumentPartner.ClosedXML/pull/18/files

It will be some effort to prepare the code for the develop so I would like to have your thoughts on the changes before starting with that. @jahav @igitur

rominator1983:
see https://github.com/ClosedXML/ClosedXML/pull/2420

PR #2419, opened 2024-09-30
add nuget.config to not rely on settings of development machine

rominator1983:
When nuget is disabled on development machines then a restore/build will fail.
Settings for packages should be handled by the git repository and not development machine settings.


jhavlicek-profinit (approved):

jahav (approved):

rominator1983:
❤️

PR #2415, opened 2024-09-26
upgrading to supported dotnet frameworks 6 in unit tests, examples and sandbox

rominator1983:
Resolve warnings by upgrading to the supported dotnet framework 6 where appropriate.
Thus dropping tests with netcoreapp3.1.
This also ensures that developers don't have to install out of support software (SDK 3.1) to build/test ClosedXml.

Issue #2400, opened 2024-09-17
when inserting a new row above a cell with a hyperlink the hyperlink is left


BrianDanielsBH:
## Read and complete the full issue template

Do not randomly delete sections. They are here for a reason.

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**
0.102.3

**What is the current behavior?**
when inserting a row above a cell with a hyperlink the hyperlink is not also moved down.

**What is the expected behavior or new feature?**

the hyperlink is also moved down

**Is this a regression from the previous version?**

n/a. 

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

Without a code sample, it is unlikely that your issue will get attention. Don't be lazy. Do the effort and assist the developers to reproduce your problem. Code samples should be [minimal complete and verifiable](https://stackoverflow.com/help/mcve). Sample spreadsheets should be attached whenever applicable. Remove sensitive information.

**Code to reproduce problem:**
```c#
public void Main()
{
    var workbook = new XLWorkbook();
var worksheet = workbook.Worksheets.Add("Sample Sheet");
var worksheet2 = workbook.Worksheets.Add("Sample Sheet2");
worksheet.Cell("A1").Value = "Test";
var hyperLink = worksheet.Cell("A1").CreateHyperlink();
hyperLink.InternalAddress = "'Sample Sheet2'!A1";
worksheet.Cell("A1").WorksheetRow().InsertRowsAbove(1);
workbook.SaveAs("HelloWorld.xlsx");
}
```

- [x] I attached a sample spreadsheet.  (You can drag files on to
[HelloWorld.xlsx](https://github.com/user-attachments/files/17032667/HelloWorld.xlsx)
 this issue)

PR #2420, opened 2024-09-30
Support round tripping of formulas in print areas

rominator1983:
Print areas can contain formulas. The PR allows for formulas to be loaded and round tripped for saving.
Previously there were exceptions when loading documents with formulas in print areas.


rominator1983:
This PR includes commits of https://github.com/ClosedXML/ClosedXML/pull/2415 and https://github.com/ClosedXML/ClosedXML/pull/2419 so that I could build things on my development machine.

ClosedXML/Excel/Cells/XLCell.cs
```diff
@@ -1463,7 +1463,7 @@ internal static String ShiftFormulaRows(String formulaA1, XLWorksheet worksheetI
                     var useSheetName = false;
                     if (matchString.Contains('!'))
                     {
-                        sheetName = matchString.Substring(0, matchString.IndexOf('!'));
+                        sheetName = matchString.Substring(0, matchString.LastIndexOf('!'));
```
jahav: Whole ShiftFormulaRows/Columns should be deleted and replaced by functionality in a FormulaSlice. Making any changes here is pointless.

It also uses regex for parsing, which is inherently wrong, another reason to scuttle it.

ClosedXML/Excel/DefinedNames/XLDefinedName.cs
```diff
@@ -148,10 +149,11 @@ public override string ToString()
 
     internal void Add(String rangeAddress)
     {
-        var byExclamation = rangeAddress.Split('!');
-        var wsName = byExclamation[0].Replace("'", "");
-        var rng = byExclamation[1];
-        var rangeToAdd = _container.Workbook.WorksheetsInternal.Worksheet(wsName).Range(rng);
+        var indexOfExclamationMark = rangeAddress.LastIndexOf('!');
```
jahav: Custom parsing. Should be replaced by parsing function from ClosedParser.

ClosedXML/Excel/Hyperlinks/XLHyperlink_public.cs
```diff
@@ -92,10 +94,10 @@ public String InternalAddress
                     return _internalAddress[0] != '\''
                                ? String.Concat(
                                     _internalAddress
-                                        .Substring(0, _internalAddress.IndexOf('!'))
+                                        .Substring(0, _internalAddress.LastIndexOf('!'))
```
jahav: Same thing, sheet reference parsing should be done through ClosedParser.

ClosedXML/Excel/IO/WorkbookPartWriter.cs
```diff
@@ -229,23 +229,41 @@ orderby worksheet.Position
...
-                             printArea.RangeAddress.
-                                 FirstAddress.ToStringFixed(
-                                     XLReferenceStyle.A1) +
-                             ":" +
-                             printArea.RangeAddress.
-                                 LastAddress.ToStringFixed(
-                                     XLReferenceStyle.A1) +
-                             ","));
-                    definedName.Text = definedNameText.Substring(0, definedNameText.Length - 1);
+                    string definedNameText = null;
```
jahav: Thre should be no serialization here. Print area is a formula that satisfies name grammar [OI-29500] 2.2.2.5. The PrintAreas by itself should provide formula that is the source of the truth for areas. Code here should only take the formula and write it.

Trying to somehow extract references and serialize them back is not going to work. E.g. the PrintAreaTestCase5.xlsx  has 
```
<definedName name="_xlnm.Print_Area" localSheetId="0">INDIRECT('Tabelle1!'!$A$1)</definedName>
```

That is never going to be pass _I got something and spit it out back, mostly unmangled_ correctly.

ClosedXML/Excel/PageSetup/IXLPrintAreas.cs
```diff
@@ -30,5 +31,10 @@ public interface IXLPrintAreas: IEnumerable<IXLRange>
         /// <param name="firstCellAddress">The first cell address.</param>
         /// <param name="lastCellAddress"> The last cell address.</param>
         void Add(IXLAddress firstCellAddress, IXLAddress lastCellAddress);
+
+        internal void Add(Formula formula);
```
jahav: Public API interfaces should never contain internal members. Instead, the container should have props: one that is internal and contains actual member and other explicit implementation of the interface that returns the internal property (see `IXLWorksheet.Hyperlinks` and others). That causes the rest of the code to use the internal implementation and have access to internal methods without explicit casting.

The method themselves should be internal.

ClosedXML/Excel/PageSetup/XLPrintAreas.cs
```diff
@@ -1,14 +1,18 @@
...
 using System.Collections.Generic;
 using System.Linq;
 
 namespace ClosedXML.Excel
 {
     internal class XLPrintAreas : IXLPrintAreas
     {
-        List<IXLRange> ranges = new List<IXLRange>();
+        private List<IXLRange> ranges = new List<IXLRange>();
+        private List<Formula> formulaRanges = new List<Formula>();
```
jahav: Source of the truth should be a single formula, per XSD and OI-29500. This internal representation is not aligned with that. There should probably be a single string field and when users asks, the formula shoud modify the formula and provide access to `IEnumerable<IXLRange>` on demand (e.g. through `FormulaReferences` though I don't like that class very much).

Name formulas have their own pecularities, e.g. bang references. Therefore, it's best to keep it unparsed, until requested.

> 2.1.568 Part 1 Section 18.2.5, definedName (Defined Name)
> The standard states that the contents of this element are an extension on the ST_Formula simple
type. Excel requires the contents of this element to follow the grammar for a name formula, as defined
in Name Formulas (§3.2.3.1.4, Name Formulas).

ClosedXML/Excel/Ranges/XLRangeBase.cs
```diff
@@ -490,8 +492,8 @@ public void DeleteComments()
 
         public bool Contains(String rangeAddress)
         {
-            string addressToUse = rangeAddress.Contains("!")
-                                      ? rangeAddress.Substring(rangeAddress.IndexOf("!") + 1)
+            string addressToUse = rangeAddress.Contains('!')
```
jahav: Same thing, there should be `ReferenceArea.TryParseA1` or something like that instead of custom parsing.

ClosedXML/Excel/XLWorkbook.cs
```diff
@@ -302,11 +304,11 @@ private void InitializeTheme()
         {
             if (name.Contains("!"))
             {
-                var split = name.Split('!');
-                var first = split[0];
-                var wsName = first.StartsWith("'") ? first.Substring(1, first.Length - 2) : first;
-                var sheetlessName = split[1];
-                if (TryGetWorksheet(wsName, out XLWorksheet ws))
+                var indexOfExclamationMark = name.LastIndexOf('!');
```
jahav: Dtto, custom reference parsing + another two occurrences below. Should use ClosedParser.

ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -1004,22 +1004,73 @@ private void LoadDefinedNames(Workbook workbook)
 
                 if (name == "_xlnm.Print_Area")
                 {
-                    var fixedNames = validateDefinedNames(definedName.Text.Split(','));
-                    foreach (string area in fixedNames)
+                    try
```
jahav: This whole thing should be replaced by something like 
```csharp
if (name == "_xlnm.Print_Area" && !string.IsNullOrWhitespace(definedName.Text))
{
  if (definedName.LocalSheetId is null)
  {
    PageOptions.PrintAreas.Formula = definedName.Text;
  }
  else
  {
    var ws = WorksheetsInternal.FirstOrDefault<XLWorksheet>(w => w.SheetId == (localSheetId + 1));
    if (ws is not null)
    {
        ws.PageSetup.PrintAreas.Formula = definedName.Text;
    }
}
```

where Formula is the source of the truth. I am not confident enought that ClosedParser can parse everything and it is not needed here anyway. If formula is unparseable, throw in XLPrintAreas only, splash zone is smaller.

ClosedXML.Tests/Excel/Loading/LoadingTests.cs
```diff
@@ -142,6 +143,78 @@ public void CanLoadAndSaveFileWithMismatchingSheetIdAndRelId()
...
+
+                using var ms2 = new MemoryStream();
+                wb.SaveAs(ms2, true);
+            });
+        }
+
+        [Test]
+        public void CanLoadAndSavePrintArea_Details()
+        {
+            Assert.Multiple(() =>
```
jahav: Please don't use Assert.Multiple. It only clutters the code without any benefit. If it's necessary to know all fails at once to understand what went wrong, it's time to reevaluate the whole approach.

ClosedXML.Tests/Excel/Loading/LoadingTests.cs
```diff
@@ -142,6 +143,78 @@ public void CanLoadAndSaveFileWithMismatchingSheetIdAndRelId()
...
+        [TestCase("PrintAreaTestCase8.xlsx")]
+        [TestCase("PrintAreaTestCase9.xlsx")]
+        [TestCase("PrintAreaTestCase10.xlsx")]
+        [TestCase("PrintAreaTestCase11.xlsx")]
+        [TestCase("PrintAreaTestCase12.xlsx")]
+        [TestCase("PrintAreaTestCase13.xlsx")]
+        [TestCase("PrintAreaTestCase14.xlsx")]
+        public void CanLoadAndSavePrintArea(string file)
+        {
+            Assert.Multiple(() =>
```
jahav: I don't see what this test checks. We can load and save file, but checks that it actually saves anything correctly, just that it saves something, possibly mangled. 

Also, I generally prefer not to use test files for exhaustive checks. It makes sense if you have a complex parsing during the load, but that is IMO wrong approach. It would be better to have 2 tests that they are loaded/savbed correctly (check that print areas are associated with correct sheet/whole workbook) and unit tests for XLPrintAreas that it can modify a formula (represented by `_xlnm.Print_Area` ) using API (e.g. adding another range to print areas would append range with a range union operator).

ClosedXML/Excel/XLWorkbook_Load.cs
```diff
@@ -1004,22 +1004,73 @@ private void LoadDefinedNames(Workbook workbook)
...
+                            }
+                        }
+
+                        foreach (var printAreaToAdd in printAreasToAdd)
+                            printAreaToAdd.Item1.Add(printAreaToAdd.Item2);
+                    }
+                    catch (Exception)
                     {
-                        if (area.Contains("["))
+                        var formula = CalcEngine.Parse(definedName.Text);
```
jahav: Please don't use `Formula` if possible. That was a valid approach back when we used _XLParser_ to parse formulas, but back then it took ~150 μs to parse formula IIRC (i.e. 10k would take 1.5 second). Current approach is to do the work while parsing without building the tree in memory (`IAstVisitor` ect.). It;s simpler (one representation) and more memory efficient. Formula and materialized tree is going the way of the dodo.

jahav (changes requested):
The thing about fixing a bugs is that they rely on suboptimal code that generally needs to be fixed first. #2400 is an example of that.

Here, it depends on
* Move functionality of IXLCell.ShiftFormulaRows/Cols to Slice
* Expose function to parse sheet references in ClosedParser
* Replace custom parsing with parser function.

After that, generally internal structure is not greate (or in a direction I want to take ClosedXML). Here, the print area shoudl be represented as a string in XLPrintAreas instead of dealing with complicated parsing/serialization. We control API, received file is given.

Next is API, where it is always fun what to do. Here, it's rather simple. Only `Clear` (which is obvious) and `Add` methods that only need to append union operator and a reference to existing formula.

rominator1983:
Thanks for your detailed feedback. It is really appreciated as some of our customers need to handle such spreadsheets.

I will try my best to adapt the PR.
There might be some questions along the way though.

I do hope we can work things out with not too many round trips. 🙂

ClosedXML/Excel/PageSetup/IXLPrintAreas.cs
rominator1983: i see. since you stated that XLPrintAreas should hold a string I will make that string public gettable/settable and see where that leads me

ClosedXML/Excel/XLWorkbook_Load.cs
rominator1983: this seems quite radical to me but I guess you're the one to decide on the architecture.
I am not quite sure what you mean with the throwing in XLPrintAreas. I thought the Formula should not be parsed at all?

ClosedXML.Tests/Excel/Loading/LoadingTests.cs
rominator1983: you're right. Assert.Multiple only has benefits during development of that specific test. That was an oversight.

ClosedXML.Tests/Excel/Loading/LoadingTests.cs
rominator1983: The idea was to have a test that checks if the parsing and serialization of the ranges/formulas can be round-tripped without an exception. The other changes you proposed make that irrelevant so I will adapt that.

rominator1983: your proposed changes make several test cases irrelevant

ClosedXML/Excel/XLWorkbook_Load.cs
rominator1983: I see. I changed the PrintArea to hold a string as you suggested in another comment. I do not quite know if anything else has to be done to be honest.

ClosedXML/Excel/Cells/XLCell.cs
rominator1983: I will need to look at more code to see what you exactly mean by that. Maybe I will put those changes in another PR and remove all these changes from this PR.

rominator1983: I took out the handling of the exclamation mark in this PR in order to have a smaller manageable PR (for you and especially me ;-) )

ClosedXML/Excel/Hyperlinks/XLHyperlink_public.cs
rominator1983: outdated. taken out of the PR

ClosedXML/Excel/DefinedNames/XLDefinedName.cs
rominator1983: outdated. taken out of the PR

ClosedXML/Excel/IO/WorkbookPartWriter.cs
rominator1983: I am now using the string of the PrintArea for writing back.

ClosedXML.Tests/Excel/Loading/LoadingTests.cs
rominator1983: I removed several test cases except one for loading/saving of the PrintArea

ClosedXML/Excel/XLWorkbook.cs
rominator1983: outdated. taken out of the PR

ClosedXML/Excel/Ranges/XLRangeBase.cs
rominator1983: outdated. taken out of the PR

ClosedXML/Excel/PageSetup/XLPrintAreas.cs
rominator1983: i do not quite see, how we could give away an enumeration of IXLRange when at the same time there can be other things (references to variables, IF-functions etc.) in the Formula. I do think that the attempt to iterate over ranges when the print area really can be any formula is valid at all.

I now did drop the enumerator in the PR and requested a re-review with that state. Could you please elaborate on how you think this could/should be done (if at all).

rominator1983:
Ok. I do think that the enumerator on IXLPrintArea was not a good idea in the first place. I am thinking of some GetReferences or GetRanges method on IXLPrintArea where the existing Formula might get parsed and the references (IXLRange[] or something else) is returned. I still do not know if its best to just skip Functions like IF there or simply collect all references regardless of their use in the formula.

~~Anyhow I do also think the now chosen attempt of using a string for the print area has the downside of not supporting renamings of sheets (worksheet.name) in a clean fashion any more.~~
Edit: I made it possible to adapt the PrintArea to renamings and the copying of Worksheets again by using FormulaConverter and IWorkbookListener analogue to defined names

So I would need some guidance on the next steps in order to make this PR a success @jahav

PR #2428, opened 2024-10-03
Basic handling of threaded comments

### LINKED ISSUE #1575: Deleting all comments in a workbook ADDs new comments to the workbook

- [x] Bug

- [X] Yes

**Version of ClosedXML**
current develop (8dcc7c10f276c59baab8b2ecba5dfb515262d9cf)

When trying to delete ALL comments of a workbook, closedxml adds comments to all cells that have a value.
This happens with workbooks created by closedxml (as in the unit test below) as well as MS Excel documents that match the criteria.

This BUG probably causes https://github.com/ClosedXML/ClosedXML/issues/1570 too. So that one could be checked if it still occurs after the fix.

When deleting ALL comments of a workbook there should not be any more comments in the workbook.

Don't know. I don't think so.

## Reproducibility

```c#
[Test]
public void CanRemoveCommentsWithoutAddingOthers()
{
    using (var stream = new MemoryStream())
    {
        // arange
        using (var wb = new XLWorkbook ())
        {
            var sheet = wb.AddWorksheet("sheet1");

            var a1 = sheet.Cell ("A1");
            var b5 = sheet.Cell ("B5");

            // enabling this will break the test
            //a1.SetValue ("test a1");

            // enabling this will break the test
            //b5.SetValue ("test b5");

            a1.Comment.AddText ("no comment");

            var cellsWithComments3 = wb.Worksheets.SelectMany (_ => _.CellsUsed (XLCellsUsedOptions.Comments)).ToArray ();

            // BUG? When adding the b5.SetValue ("test b5"); this will return 2
            Assert.That (cellsWithComments3.Length, Is.EqualTo (1));

            wb.SaveAs (stream, true);
        }

        stream.Position = 0;

        {
            var cellsWithComments = wb.Worksheets.SelectMany (_ => _.CellsUsed (XLCellsUsedOptions.Comments)).ToArray ();

            Assert.That (cellsWithComments.Length, Is.EqualTo (1));

            cellsWithComments.ForEach (_ => _.Clear (XLClearOptions.Comments));

            wb.Save ();
        }

        // assert
        stream.Position = 0;

        {
            var cellsWithComments = wb.Worksheets.SelectMany (_ => _.CellsUsed (XLCellsUsedOptions.Comments)).ToArray ();

            // BUG? when adding a1.SetValue ("test a1"); this will return at least one cell instead of none
            Assert.That (cellsWithComments, Is.Empty);
        }
    }
}
```

rominator1983:
Handle threaded comments with a new OPT-IN feature.
This allows developers that desperately need the functionality of reading threaded comments to use that while at the same time not breaking functionality for existing use cases.
There is a straight forward way to handle threaded comments in another way in the future and thus deprecating this behavior.

This PR also fixes https://github.com/ClosedXML/ClosedXML/issues/1575
This PR also includes commits from https://github.com/ClosedXML/ClosedXML/pull/2415


jahav:
I am adding a one PR per person policy. When #2420  is resolved, you can open this one.

rominator1983:
Ok

Issue #2455, opened 2024-10-10
StyleFont unit test fails


Igor-Zlatomrezhev:
Unit test `StyleFont` is failing.

**Test detail summary**:
```
 StyleFont
   Source: StylesTests.cs line 40
   Duration: 2.1 sec

  Message: 
  Actual file '...\ClosedXML.Tests\bin\Debug\net462\Generated\Examples\Styles\StyleFont.xlsx' is different than the expected file 'Examples.Styles.StyleFont.xlsx'. The difference is: '/docProps/app.xml :NonEqual
/docProps/core.xml :OnlyOnRight
/xl/sharedStrings.xml :NonEqual
/xl/styles.xml :NonEqual
/xl/workbook.xml :NonEqual
/xl/worksheets/sheet1.xml :NonEqual
'
  Expected: True
  But was:  False


  Stack Trace: 
TestHelper.RunTestExample[T](String filePartName, Boolean evaluateFormulae) line 99
StylesTests.StyleFont() line 42
```

Unit test generates xlsx file with width of column B adjusted to its contents. At the same time, width of that column in the target xlsx file is not adjusted to its contents.

jahav:
Column B in Style Font.xlsx utilizes a lot of different fonts to check that their width is calculated correctly for automatic width calculation.

You are likely missing some of them. Because it works locally and on CI, closing as non-reproducible.

PR #2456, opened 2024-10-10
#2455: generator updated to match target xlsx file

Igor-Zlatomrezhev:
Column B in `StyleFont.xlsx` file is not adjusted to its contents. So I had to update generator to use fixed column width instead.


jahav:
There is no root cause, which is likely a computer without Calibri font. 

Closing, because I don't want a fixed width. Its supposed to adjust width, because it's an example how most users want to use it.

Issue #2425, opened 2024-10-02
0.104.1 FormulaParser ParsingException in XLTable.ReplaceData


RemcoBlok:
**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [ ] Yes
- [x] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.104.1

**What is the current behavior?**

I have an IXLTable with a single row. One field contains a formula based on the other fields. When I call `ReplaceData()` on the table passing in an array of business objects T with more than one item, such that the IXLTable needs to insert rows, I get the following exception in 0.104.1. 

```
ClosedXML.Parser.ParsingException: Error at char 0 of '': Unexpected token EofSymbolId.
   at ClosedXML.Parser.FormulaParser`3.RefAtomExpression(Boolean replaceFirstAtom, TNode refAtom)
   at ClosedXML.Parser.FormulaParser`3.RefSpillExpression(Boolean replaceFirstAtom, TNode refAtom)
   at ClosedXML.Parser.FormulaParser`3.RefRangeExpression(Boolean replaceFirstAtom, TNode refAtom)
   at ClosedXML.Parser.FormulaParser`3.RefIntersectionExpression(Boolean replaceFirstAtom, TNode refAtom)
   at ClosedXML.Parser.FormulaParser`3.RefImplicitExpression(Boolean replaceFirstAtom, TNode refAtom)
   at ClosedXML.Parser.FormulaParser`3.RefExpression(Boolean replaceFirstAtom, TNode refAtom)
   at ClosedXML.Parser.FormulaParser`3.AtomExpression(Boolean skipRangeUnion, Boolean& isPureRef)
   at ClosedXML.Parser.FormulaParser`3.PrefixAtomExpression(Boolean skipRangeUnion, Boolean& isPureRef)
   at ClosedXML.Parser.FormulaParser`3.PercentExpression(Boolean skipRangeUnion, Boolean& isPureRef)
   at ClosedXML.Parser.FormulaParser`3.PowExpression(Boolean skipRangeUnion, Boolean& isPureRef)
   at ClosedXML.Parser.FormulaParser`3.MultiplyingExpression(Boolean skipRangeUnion, Boolean& isPureRef)
[...]

PR #2462, opened 2024-10-13
#2425: A fix for ParsingException: Unexpected token EofSymbolId

Igor-Zlatomrezhev:
`ReplaceData` method used to raise following exception when the worksheet had defined names:
>ClosedXML.Parser.ParsingException: 'Error at char 0 of '': Unexpected token EofSymbolId.'

### DIFF
--- a/ClosedXML.Tests/Excel/Tables/AddingAndReplacingTableDataTests.cs
+++ b/ClosedXML.Tests/Excel/Tables/AddingAndReplacingTableDataTests.cs
@@ -77,27 +77,6 @@ namespace ClosedXML.Tests.Excel.Tables
             return wb;
         }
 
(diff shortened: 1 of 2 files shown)


hevilhuy:
Someone please review this because it fixes an important problem with table of the latest version

heyixin79:
@Igor-Zlatomrezhev @hevilhuy 
The same exception was raised when I use sheet.Row().InsertRowsBelow() and sheet.Column().InsertColumnsAfter().
When I tested f000ad4,
The ReplaceData() method and sheet.Row().InsertRowsBelow() method was OK, 
But sheet.Column().InsertColumnsAfter() still raised the same exception.
It seems that there is a same issue in MoveDefinedNamesColumns  method.

PR #2466, opened 2024-10-14
Update core runtime version to net8.0

jahav:
Generally speaking, we should test on minimum supported framework, which is net462 for .NET Framework and net8.0 for Core (in less than a month). net6 is going to stop being supported in oct 2024, so before next release.


rominator1983:
there is one test that fails with net8.0. Something with garbage collection. Did not check but cannot be that difficult.

Issue #2346, opened 2024-05-23
Tests failing on Mac


stefannikolei:
I opened the Solution on my Mac (Intel) and have some tests (20) that fail. Is this a known issue? Perhaps it would be good to test against Windows, Linux and Mac. Should be fairly easy with GitHub Actions. 

Are you interested that I look into it?

```


```

stefannikolei:
Hmm seems like there is no interest in contributions?

Issue #2171, opened 2023-09-20
SixLabors.Fonts v2.0.0 changed to Commercial License


dbuhl:
## Read and complete the full issue template

Do not randomly delete sections. They are here for a reason.

**Do you want to request a *feature* or report a *bug*?**
- [ ] Bug
- [ ] Feature
- [x] Question

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

1.0.0

**What is the current behavior?**

Works with SixLabors.Fonts 1.0.0 >

**What is the expected behavior or new feature?**

Will ClosedXML eventually require the SixLabors.Fonts v2.0.0 or greater as this changes to a commercial license from the Apache License? This is an important question for future planning for out application.  I would assume this will become an issue for most users of the library if this should occur.

**Is this a regression from the previous version?**

N/A

## Reproducibility

N/A

Without a code sample, it is unlikely that your issue will get attention. Don't be lazy. Do the effort and assist the developers to reproduce your problem. Code samples should be [minimal complete and verifiable](https://stackoverflow.com/help/mcve). Sample spreadsheets should be attached whenever applicable. Remove sensitive information.

**Code to reproduce problem:**
```c#
// N/A
```

- [ ] I attached a sample spreadsheet.  (You can drag files on to this issue)

jahav:
I do not plan on using `SixLabors.Fonts 2.0.0`. The primary reason is lack of support for `netstandard 2.0`.

For now, I plan to keep dependency on `SixLabors.Fonts 1.0.0 `and investigate `SkiaSharp` as a likely replacement. No concrete timeline, I would prefer to spend time on other things, but if a security bug or some other high priority issue appears, it will get priority. 

`SkiaSharp` also means the `Blazor` applications with `WebAssembly` wouldn't be able to use it (though there is an extension point), because it's a wrapper over native `Skia` library. I think `SkiaShap` is basically the only library left, based on past research into alternatives (https://github.com/ClosedXML/ClosedXML/issues/1805).

As for the `SixLabors.Fonts` split license, users of `ClosedXML` would likely be covered by Transitive Package Dependency clause, though lack of `netstandard 2.0`  makes this a moot point to investigate further. I also understand the reluctance of that license in a chain of dependencies, so switch would likely came up anyway.

[...]

PR #2470, opened 2024-10-16
Try to implement DefaultGraphicEngine with SkiaSharp

stefannikolei:


stefannikolei:
@jahav I tried to port the GraphicEngine to SkiaSharp. I think it will be really hard to get all tests green. The results of width and height are just of by a small number.. 

But my main problem is, that I am on a Mac and the tests from dev are not passing there. so its hard to know what's the real problem..

stefannikolei:
@jahav I would still be willing to get this PR into an mergeable state. But I am having problems with the development environment. I am using an MAC and on that some tests fail. This makes it hard to get this one right.

How can we go forward on that? I am stuck on older ImageSharp versions because of #2171

I would also contribute some time into setting up Github actions build for multiple platforms (#2346
). But I just need an ok of you before I invest the time for nothing.

jahav:
@stefannikolei If you need it, set the static property `LoadOptions.DefaultGraphicEngine` to the engine (see [docs](https://docs.closedxml.io/en/latest/features/graphic-engine.html)).

You can also still use SixLabors.Fonts version 1.0.1 ([nuget](https://www.nuget.org/packages/SixLabors.Fonts#versions-body-tab)), it is still available under Apache license. Just explicitly set dependency in the project.

As for the PR, probably don't invest time. `AdjustToContent` doesn't work properly for several things and if I have to invest time into checking all test files because of minor font size changes, I want to be benefit of some fixes so I don't need to do check files over again and again.

Situation is going to require change sooner or later, but Skia doesn't work on WebAssembly (=client Blazor) so I am postponing it.

`AdjustToContent` is a terrible mess (to be more clear Excel functionality is a mess) and each library calculates slightly differently, see [Cell-Dimensions](https://github.com/ClosedXML/ClosedXML/wiki/Cell-Dimensions#skiasharp-code) for some stuff of how I think it works.

stefannikolei:
Hm have not tested blazor with skis yet. But from what I can see e.g this PR https://github.com/mono/SkiaSharp/pull/1811

It should work.

terenzif (commented):
ok

PR #2520, opened 2024-11-20
Add non generic GetValue and TryGetValue overloads in IXLCell interface

michaeltg17:


jahav:
I don't want to introduce another API at this time. If you want something like that, use an extension method to wrap it in your project.
```csharp
internal static class IXLCellExtensions
{
	public static Boolean TryGetValue(this IXLCell cell, Type type, out object value) {
		throw new NotImplementedException();
	}
	public static object GetValue(this IXLCell cell, Type type)
	{
		throw new NotImplementedException();
	}
}
```

Closing.

michaeltg17:
Yes, just wanted to avoid the use of Reflection. In any case, great package!

image

PR #2532, opened 2024-11-25
Changed int to uint since Excel files can have pagebreaks with a uint id

### LINKED ISSUE #1732: Parsing row break data while reading excel file throws an OverflowException
- [x] Bug
- [ ] Feature
- [ ] Question

- [ ] Yes
- [x] No

Related source code not changed for quite a while.

**Version of ClosedXML**

0.95.4

System.OverflowException (Value was either too large or too small for a UInt32.) when parsing row break from the element's id attribute @ XLWorkbook_Load.LoadRowBreaks().

Parsing should be able to handle uint.MaxValue as unsignedInt seems to be the type according to the OpenXML specification.

No.

## Reproducibility

Can be reproduced by loading the attached file.
[MyExcelFile.xlsx](https://github.com/ClosedXML/ClosedXML/files/7386243/MyExcelFile.xlsx)

```c#
public void Main()
{
    // Code standards:
   using (var wb = new XLWorkBook("MyExcelFile.xlsx"))
   {
   }
}
```

Soreil:
fixes https://github.com/ClosedXML/ClosedXML/issues/1732 
This change affects the API of IXLPageSetup and XLPageSetup since currently int is used where uint should be.

### DIFF
--- a/ClosedXML/Excel/IO/WorksheetPartWriter.cs
+++ b/ClosedXML/Excel/IO/WorksheetPartWriter.cs
@@ -1228,7 +1228,7 @@ namespace ClosedXML.Excel.IO
                 var existingBreaks = rowBreaks.ChildElements.OfType<Break>().ToArray();
                 var rowBreaksToDelete = existingBreaks
                     .Where(rb => !rb.Id.HasValue ||
-                                 !xlWorksheet.PageSetup.RowBreaks.Contains((int)rb.Id.Value))
+                                 !xlWorksheet.PageSetup.RowBreaks.Contains(rb.Id.Value))
                     .ToList();
 
                 foreach (var rb in rowBreaksToDelete)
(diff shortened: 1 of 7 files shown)


jahav:
We use int unless it is really needed to use uint. In this case, there is no need to use uint. Only useable values are 1..1048576 row row breaks and 1..16384 for col breaks.

-1 (=4294966306 when converted to uint) is outside of meaningful values. It says break at the top of the first column, before the are any rows to break.

[ISO-29500]: > id (Id): Zero-based row or column Id of the page break. Breaks occur above the specified row and left of the specified column. The possible values for this attribute are defined by the W3C XML Schema `unsignedInt` datatype

[OI-29500]
> The standard states that the values of the id attribute are defined by the XML Schema unsignedInt
datatype. Excel restricts the value of this attribute to be at least 1 and at most 1048576.

PR #2546, opened 2024-12-01
Reimplement DOLLAR

jahav:
Original implementation didn't work with negative digits (e.g. `DOLLAR(12345.12, -2)` = `$12'300`). Also convert from `Expression` based to `AnyValue` base. 36 functions to go.


jahav:
Yay, Windows changed French thousands separator between Win11? and previous versions. Now I see why Excel just keeps all culture formatting information inside. It it didn't do that, spreadsheets might break between different versions of Windows and would break some stuff.

PR #2593, opened 2024-12-21
Xml tree reader

jahav:
@igitur @Pankraty Could you please take a look?

One of long term goals is to remove dependency on OpenXMl SDK. The first step is to be able to read parts without it. This is a PoC of how do I plan on dealing with that. It is for 0.106

## Rationale

OpenXML SDK is used during loading/writing as an intermediate representation. It shields us from need to deal with schema (everything is strongly typed) and allows us to patch the XML.

* OpenXML SDK loads parts to memory. That is not great, both from memory standpoint as well as performance.
* Updating existing workbook has a problem. It 2.x, it required a lot of memory (basically keep whole uncompressed XML in memory), in 3.x it writes it to a disk as a workaround (the root cause is in `System.IO.Packaging`).
* It encourages ClosedXML to use patching method of XML.

What I want is to represent workbook fully in memory and read and write it ourselves. OpenXML SDK is great and served us well. But it's time to remove training wheels and stand on our own. In the end, it's only a middleware between ClosedXML and xml representation.

In my opinion, it's one of four main architectural problems with ClosedXML:
* Full representation + read/write (todo)
* Formulas and CalcEngine (mostly done)
* Better internal structure (mostly done)
* Deal with structural changes (e.g. insert area and everything is adjusted, including formulas), copy and so on (in progress)

Sidenote: Mostly done means that fundamentals are there. Something that can be improved without ripping everything and replacing it.

## Requirements for XML parser

* Must use streaming for small memory footprint
* Easy to write parser that matches XML schema
* High performance
* Must have a way to deal with MCE preprocessor, at least `AlternateContent` or `Ignorable ` namespaces (ISO-29500 Part 3).
* Must be able to read from `XmlReader` (I can get access to that from OpenXML SDK), either directly or through adapter.
* Must allow gradual transition (=part by part) from OpenXMl SDK.

## Research
In past few months, I took a look at XML parsing options and none of the "higher" level ones are suitable. Most of them deal with object-XML translation and many of them have serious shortcomings when parsing a specified XML (e.g. [ExtendedXmlSerializer](https://github.com/ExtendedXmlSerializer/home/issues/636)) or no documentation/bugs (e.g. [NetBike.Xml](https://github.com/netbike/netbike.xml)). The `XmlSerializer` serializer requires public contract classes and is too rather limited.

LINQ to XML is great for selecting a specific nodes, not for parsing a fixed schema.

Also, most of the project are in maintenance mode, because XML has largely been supplanted by JSON, plus performance is unknown (or poor).

None of these options could even deal with stuff like MCE preprocessor.

The only option seems to be to use `XmlReader`.

## XmlReader

I took a look at the, the canonical way of parsing XML is terrible. It's too low level. There is no way I could use the API that depends on reader.NodeType + match name in a switch to parse existing schema and maintain sanity.

I need something that is really simple to write, similar to how writers are written (e.g. [PivotTableDefinitionPartWriter2.cs](https://github.com/ClosedXML/ClosedXML/blob/develop/ClosedXML/Excel/IO/PivotTableDefinitionPartWriter2.cs)).

I don't really need all that complex stuff:

* DTD is disallowed (not by standard, but Excel refuses to load it, so it's same thing).
* Mixed content is not allowed, other than `extlist` (that allows Any)
* Everything is basically a tree structure of elements. Values are either in attributes or leaf nodes have text content. 
* Schema is simple, using mostly `choice`  or `sequence` with `minOccurence`/`maxOccurence`.

I can strip unused stuff and make something simpler to use. So I made a wrapper.

## XmlTreeReader

It is basically a parser that moves along open and close tags. It always is on of these elements, so no need to deal with `Xmlreader.IsEmptyElement` duality.

The current element is either interpreted as processed element (i.e. it has a matching parser logic) or as a lookahead of size one. I think LA(1) si enough so I can parse whole schema.

It has very simple semantic. Basically three methods:
* `TryOpen(name)` - take a lookahead element and check it is opening element `elementName`. If it is, open the element and mark it as processed. Use `Get*` methods to get 
* `TryClose(elementName)` - take a lookahead element and check it is closing element `elementName`. If it is, close the element and mark it as processed (i.e. lookahead will be next element).
* `Get*` - get attribute value from from currently processed element.

An illustration of how to parse stuff. Generally a method per type.
```csharp
LoadStyleSheet() 
{
  // Check we are at this element and open it. throw if not correct one.
  xml.Open("styleSheet", _mainNs);

  // Is lookahead (i.e. element after styleSheet) numFmts? If it is, open it and parse the content in a method
  if (xml.TryOpen("numFmts", _mainNs))
    ParseNumFmts(xml);

  if (xml.TryOpen("fonts", _mainNs))
      ParseFonts(xml);

   // ...
  xml.Close("styleSheet", _mainNs);
}

// Parse CT_Fonts
private void ParseFonts(XmlTreeReader xml)
{
  // If we are here, we are processing `fonts` element. Because we are processing element, we can get its  attributes
  var fontCount = xml.GetOptionalUint("count");

  // while the next element is a `font` (for the first one, it is first element after <fonts>,
  // for subsequent it is next element after previous </font>). Parse the font
  while (xml.TryOpen("font", _mainNs))
    ParseFont(xml);

  // Close the fonts collection. There should be closing element </fonts>.
  // Possibly there is only <fonts count="0"/>, but XmlTreeReader parses it same way as <fonts count="0"></fonts>
  // i.e. as two elements
  xml.Close("fonts", _mainNs);
}

// Parse CT_Font
private void ParseFont(XmlTreeReader xml) {
  // 
}
```

I think this is rather easy to write, so it should be fine.

There is currently no MCE preprocessor, but it shouldn't be a problem. I have read it and it should be doable for `AlternateContent`/`Choice`/`Fallback`+`Ignorable`. I am not sure I will bother with `MustUnderstand` or `ProcessContent`. I have never seen them use, not even in CommonCrawl dump.

## Risks

One of the biggest risks is that parser will throw on valid input. To mitigate it, I will use CommonCrawl dump (tens of thousands of xlsx files found across the internet) to check that parser doesn't throw. It might still interpret it incorrectly, but at least it doesn't throw.

The other problem will be pieces of XML schema that don't match the expected tree structure. For that I plan to just load them into memory (`XDocument.Load(xmlreader.ReadSubtree())`). That should be only `extLst`, but possibly others.

## PoC scope

The code is only PoC. I used it against CommonCrawl dump and to experiment. It doesn't actually fill ClosedXML structures, nor is it intended to.

The PoC StyleSheetreader parsed nearly every styles.xml from CommonCrawl dump of ~50k files (there were 4 `AlternateContent` from MCE (=TODO), 1 incorrect namespace declaration and one that didn't match schema).

## Plan of works

This is for 0.106. The plan is to clean up the PoC, add basic stuff for MCE precprocessor and use it for a simple part I can represent fully. Currently the best candidate is pivot records part. It's simple and can be very large.

From there, possibly stylesheet, though that remains to be seen.

I would love to do it for worksheet part, but that `XLWorksheet` can't represent the sheet part fully. It will probably be the last part to be reworked.

Unimplemented/unknown parts (charts, forms, connections...) will be just loaded from source workbook as `XDocument` and later written as to part as-is (possibly with some relId changes). I am also exploring a possibility of using source generators to generate DTO and parsers from XML schema. From practical PoV, I don't want to write this code unless necessary. It's a manual labor that is error prone. I would prefer to use source of truth (XML Schema) do generate parsing (and possibly mapping where not specified) for me.


igitur:
Hey, currently on vacation, but will give this my full attention when I'm back.

jahav:
@igitur Ping? did you find some time to look it over?

PR #2596, opened 2024-12-26
Update ClosedParser to 2.0 (bang reference)

jahav:
Parser version 2.0 has two new methods in the `IAstFactory` interface for bang references (used only in defined names). Adding methods to interface is a breaking change, so major version was increased to 2.

Bang reference (e.g. `!$B$7:$D$8`) are 
* now parseable and formulas with it won't throw during sheet rename and delete or formula conversion from A1 to R1C1 and back.
* calc engine doesn't evaluate them, they are represented as `NotSupportedNode` for now.

There is also a method for bang name (`!Data`), but it's not yet parseable, but will be.

ClosedXML.Parser 2.0.0-preview1 package is unlisted on nuget.org.

### DIFF
--- a/ClosedXML/ClosedXML.csproj
+++ b/ClosedXML/ClosedXML.csproj
@@ -46,7 +46,7 @@
     </PackageReference>
     <PackageReference Include="RBush" Version="4.0.0" />
     <PackageReference Include="SixLabors.Fonts" Version="1.0.0" />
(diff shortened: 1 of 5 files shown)


rajtilakvyas:
@jahav I have been using ClosedXML 0.105.0-rc, which i downloaded from https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts

which needs ClosedXML.Parser2.0.0-preview1, from where i can get that, i could not find on nuget.

PR #2601, opened 2025-01-15
Small increase to unit test coverage

sbeca:
Many months ago I was working on a local fork of ClosedXML to improve the conformance of some functions, but never got it to a place where I could upstream the changes.

Now that you have done a big port of all the functions and improved their conformance in the process, most of my changes don't need completing and upstreaming, which is awesome!

My local changes did include a small increase to unit test coverage, so I could see if my changes matched Excel's output. I think those are still worth upstreaming so that ClosedXML can have the best test suite possible, so here is a small PR with some more test cases


jahav:
LGTM. Thanks.

Issue #1772, opened 2022-03-07
Loading an excel file with a comment fails with 'Sequence contains no matching element'


d3xter:
## Read and complete the full issue template

Do not randomly delete sections. They are here for a reason.

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.95.4

**What is the current behavior?**

Loading fails with the following error:

```
System.InvalidOperationException: Sequence contains no matching element
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
   at ClosedXML.Excel.XLWorkbook.LoadSpreadsheetDocument(SpreadsheetDocument dSpreadsheet)
   at ClosedXML.Excel.XLWorkbook.LoadSheets(Stream stream)
   at ClosedXML.Excel.XLWorkbook.Load(Stream stream)
   at ClosedXML.Excel.XLWorkbook..ctor(Stream stream, XLEventTracking eventTracking)
   at ClosedXML.Excel.XLWorkbook..ctor(Stream stream)
```

**What is the expected behavior or new feature?**

Excel file is loaded

**Is this a regression from the previous version?**

Regressions get higher priority. Test against the latest build of the previous minor version. For example, if you experience a problem on v0.95.3, check whether it the problem occurred in v0.94.2 too. 

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

Without a code sample, it is unlikely that your issue will get attention. Don't be lazy. Do the effort and assist the developers to reproduce your problem. Code samples should be [minimal complete and verifiable](https://stackoverflow.com/help/mcve). Sample spreadsheets should be attached whenever applicable. Remove sensitive information.

**Code to reproduce problem:**
```c#
public void Main()
{
   var excelDocument = new XLWorkbook("C:\path\to\Test.xlsx");
}
```

- [x] I attached a sample spreadsheet.  (You can drag files on to this issue)

[Test.xlsx](https://github.com/ClosedXML/ClosedXML/files/8196313/Test.xlsx)

rominator1983:
[...]

PR #2606, opened 2025-01-28
Fix load comments

sebgerber:


Arttuv:
Was there any particular reason this did not go forward? I just run into the issue of ClosedXML crashing when loading Excel files with certain kinds of comments (https://github.com/ClosedXML/ClosedXML/issues/1772), and I'm now trying to figure out a way forward. Just skipping the comments would be better than crashing.

PR #2611, opened 2025-02-07
Ignore `_X00000000_` (capital X followed by 8 hex chars) as XML escaping char for decoding

### LINKED ISSUE #2610: Incorrect Column Name Read from Excel File  "DE_XAB500161_seo_title"
Description:
I am trying to read an Excel file using C# and ClosedXML. One of the columns in my file has the name:

DE_XAB500161_seo_title

However, when I retrieve the column name using:

column.FirstCell().Value.ToString()

The result is incorrect:

DEšseo_title

It seems that part of the column name (_XAB500161_) is missing or replaced by unexpected characters (š).

Hier is full method

public static List<string> GetWorksheetColumns( IXLWorksheet worksheet)
{
     List<string> columnExcel = new List<string>();

    if (worksheet != null && dataTable == null)
    {
        foreach (IXLColumn column in worksheet.Columns())
        {
            if (!string.IsNullOrEmpty(column.FirstCell().Value.ToString().Trim())))
            {
                columnExcel.Add(column.FirstCell().Value.ToString());
            }
        }
    } 

    return columnExcel;
}

Best Regards
Alex

Hasan-75:
#### Ignore `_X00000000_` (capital X followed by 8 hex chars) as XML escaping char for decoding. Fixes #2610
 
 ## Investigation
When saving string values, unicode characters that are unsupported by XML are encoded as one of the following formats *(Check [References](#references)  section)*:
1. `_x0000_`
2. `_X0000_`
3. `_x00000000_`
4. `_X00000000_`

**Any hexadecimal characters can be in place of the 0s. And those characters after the `_x` or `_X` represent the unicode hexadecimal code.**

That implies that `_XAB500161_` is a valid escaped unicode character and it's being converted into `š`.

A similar issue was reported before: **Issue #1154**

That issue was fixed in this PR:
https://github.com/ClosedXML/ClosedXML/pull/1155

But this PR only contained a solution that could handle strings with format 1 and format 2 containing 4 hexadecimal characters after `_x` or `_X`.

## Solution
Solution will be to update the regular expressions so that it can handle other two formats of encoded strings containing 8 hexadecimal characters after `_x` or `_X`.

## References
https://github.com/dotnet/corefx/blob/b8b81a66738bb10ef0790023598396861d92b2c4/src/System.Private.Xml/src/System/Xml/XmlConvert.cs#L205

https://github.com/dotnet/corefx/blob/b8b81a66738bb10ef0790023598396861d92b2c4/src/System.Private.Xml/src/System/Xml/XmlConvert.cs#L96

### DIFF
--- a/ClosedXML.Tests/Excel/Misc/XmlEncoderTests.cs
+++ b/ClosedXML.Tests/Excel/Misc/XmlEncoderTests.cs
@@ -20,6 +20,9 @@ namespace ClosedXML.Tests.Excel
 
             // https://github.com/ClosedXML/ClosedXML/issues/1154
             Assert.AreEqual("_Xceed_Something", XmlEncoder.DecodeString("_Xceed_Something"));
+
+            // https://github.com/ClosedXML/ClosedXML/issues/2610
+            Assert.AreEqual("DE_XAB500161_seo_title", XmlEncoder.DecodeString("DE_XAB500161_seo_title"));
         }
 
         [Test]
--- a/ClosedXML/Utils/XmlEncoder.cs
+++ b/ClosedXML/Utils/XmlEncoder.cs
@@ -6,8 +6,8 @@ namespace ClosedXML.Utils
 {
     internal static class XmlEncoder
     {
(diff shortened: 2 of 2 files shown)


jahav:
Superseeded by #2661.

Issue #2629, opened 2025-03-14
XLFilterType.None is NotSupported


djseng:
Attempting to make modifications to a large spreadsheet. When I `SaveAs` the spreadsheet, an exception occurs b/c filter type is none in one of the unrelated sheets involved in the operation.

```csharp
wb.SaveAs(outputStream,
            new SaveOptions
            {
                GenerateCalculationChain = false,
                ConsolidateConditionalFormatRanges = false,
                ConsolidateDataValidationRanges = false
            })
```

https://github.com/ClosedXML/ClosedXML/blob/develop/ClosedXML/Excel/IO/WorksheetPartWriter.cs#L1578

I looked at the xml and see xml like

```xml
<autoFilter ref="A11:AM388" xr:uid="{C5B29462-0CA2-4446-9D5B-55AA8BB5D614}" />
```

Which I suspect is causing to get this `None` filterType

catman8s:
I may I might a solution to your issue. Check it out here [click](https://javierlopez.eu/captha/)

JarRami:
Just ran into the same thing. The only autofilter in my file is

```
<autoFilter ref="B2:B3" xr:uid="{F4790E31-B825-49E9-A955-AA2A07A2E705}">
	<filterColumn colId="0" hiddenButton="1"/>
</autoFilter>
```

so slightly different than the OPs example, but I'm still hitting the default

```
default:
    throw new NotSupportedException();
```
at https://github.com/ClosedXML/ClosedXML/blob/develop/ClosedXML/Excel/IO/WorksheetPartWriter.cs#L1569

I've attached below the offending tiny example file. The exception is hit with just a 

```
 using (XLWorkbook wb = new XLWorkbook("Autofilter exception.xlsx")) {
     wb.Save();
 }
```

[Autofilter exception.xlsx](https://github.com/user-attachments/files/22327596/Autofilter.exception.xlsx)

Edit. Interestingly, if I open the file in Excel, and check the Filter Button checkbox in the Table Design tab for the offending table and save the file, it works. And the autofilter is now shown below which closely resembles OPs one. I guess showing the Filters can be used as a workaround in the meantime for anyone stumbling onto this.

`<autoFilter ref="B2:B3" xr:uid="{F4790E31-B825-49E9-A955-AA2A07A2E705}"/>`

PR #2631, opened 2025-03-17
Fix: try parse formula, autofilter none

djseng:
Had two issues when working with a fairly complex spreadsheet.

1. `XLFilterType.None` is NotSupported #2629
2. Janky formulas won't allow save. (did not create an issue for this yet.)

### DIFF
--- a/ClosedXML/Excel/CalcEngine/Visitors/FormulaTransformation.cs
+++ b/ClosedXML/Excel/CalcEngine/Visitors/FormulaTransformation.cs
@@ -24,7 +24,14 @@ internal static class FormulaTransformation
         if (!MightContainFutureFunction(formula.AsSpan()))
             return formula;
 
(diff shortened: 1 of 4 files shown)


fushman:
Just a note that the try catch may have the side effect of also resolving the problem I was running into and attempted to fix with #2630

djseng:
43e4379cc37f981d0b50223d315b51ce1adcd996 resolves another problem that came up with deleting rows. The formula was already a bad formula. Essentially these are going to allow modifications to a spreadsheet that has already been generated by something that isn't double checking formula validity.

matt-henderson:
For references to this issue, setting ConsolidateConditionalFormatRanges to false in the SaveOptions works as a workaround to this issue.

Bykiev (approved):

PR #2664, opened 2025-04-16
Fix issue loading .xlsx files with "Notes" missing textBox

kaleem213:
When opening .xlsx files with Notes, an exception occurred if the textBox was missing. 

To fix this:

- Added a check to see if the textBox exists.


kaleem213:
Hi! If you could please review and approve this PR at your earliest convenience and update the NuGet package, I’d really appreciate it. I need this fix for a production issue, and some client requests are currently pending due to this. Thanks so much!

Issue #2694, opened 2025-05-18
Invalid HorizontalAlignment Parsing Throws Exception


jafin:
## Read and complete the full issue template

Do not randomly delete sections. They are here for a reason.

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

`v0.105.1`

**What is the current behavior?**

I have 3rd party source file which generates a invalid HorizontalAlignment attribute value.
They are setting the value of horizontal to `Center` whereas I believe it has to be `center`  according to 

https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheet.horizontalalignmentvalues?view=openxml-3.0.1

```xml
<xf numFmtId="0" fontId="0" fillId="0" borderId="0" xfId="0" applyAlignment="true">
<alignment horizontal="Center" vertical="Center"></alignment>
</xf>
```

Excel will load the file, and discard the invalid formatting.  
ClosedXML 0.105.1 with throw an exception when attempting to convert the bad value.

**What is the expected behavior or new feature?**

Change code to be more lenient of bad source values, and guard by checking alignment.Horizontal.IsValid first before attempting to convert ToClosedXML otherwise defer to default value.  Avoiding the exception.

```c#
Horizontal = alignment.Horizontal?.Value.ToClosedXml() ?? defaultAlignment.Horizontal,
Vertical = alignment.Vertical?.Value.ToClosedXml() ?? defaultAlignment.Vertical,
```

**Is this a regression from the previous version?**

No

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

Without a code sample, it is unlikely that your issue will get attention. Don't be lazy. Do the effort and assist the developers to reproduce your problem. Code samples should be [minimal complete and verifiable](https://stackoverflow.com/help/mcve). Sample spreadsheets should be attached whenever applicable. Remove sensitive information.

**Code to reproduce problem:**
No Code Sample

- [ ] I attached a sample spreadsheet.  (You can drag files on to this issue)

jahav:
I have seen it many times, but I never asked. What is the mental process where standardized issue template (the one that says don't delete it and fill it) is deleted and replaced with one-of-a-kind text?

jafin:
[...]

PR #2696, opened 2025-05-18
Update github issue template format.

jafin:
Relates to #2694


jahav:
I modified the template so it works like it used to.

PR #2700, opened 2025-05-19
Population of auto filters in the WorksheetPartWriter class support tables with only some colums filtered

leovel:
Fix: Correct The population of auto filters in the WorksheetPartWriter class
Previously, tha case when a Table Object don't have auto filters in all columns was not taking into account, which could lead to a NotSupportedException being thrown to the user. This change ensures that only the columns with filters are processed in the switch preventing potential runtime errors when saving.


jahav:
Closing per first sentence in [CONTRIBUTING.md](https://github.com/ClosedXML/ClosedXML/blob/develop/CONTRIBUTING.md).

Issue #1755, opened 2022-01-20
Error in /xl/styles.xml-Part 


kuch3n:
## Read and complete the full issue template

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [x] Yes
- [ ] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

.0.95.999.2326

**What is the current behavior?**

Copying a worksheet will give an error as above

**What is the expected behavior or new feature?**

No error

**Is this a regression from the previous version?**

v0.95.4 bug
v0.95.3 bug
v0.94.2 bug
v0.93.1 bug
v0.92.1 working

## Reproducibility

**Code to reproduce problem:**
```c#
static void Main(string[] args) {
            File.Copy("test.xlsx", "copy.xlsx", true);
            using XLWorkbook wb = new("copy.xlsx");

            bool res = wb.TryGetWorksheet("Original", out IXLWorksheet sheet);

            sheet = sheet.CopyTo(wb, "copy");

            wb.Save();
            
        }
```
I don't know how to remove information from the excel like company, name, etc. and my company doesn't allow me to publish such information. 
But you only need to fill some cells with a colour and name the sheet "Original"
- [ ] I attached a sample spreadsheet.  (You can drag files on to this issue)

ggrote:
Any news about it? Is there some sort of Workaround?

kuch3n:
> Any news about it? Is there some sort of Workaround?

Either use the versions posted or remove the styles

bogdandoi:
is there any plan to fix this?

jahav:
Not any concrete plans. Priority ATM is to deal with PRs.

akade:
Just encountered a similar (if not the same) bug on the latest version (0.101.0) Just having any custom table style "corrupts" the worksheet when saving again:

```csharp
using ClosedXML.Excel;

using XLWorkbook workbook = new(@"Test.xlsx");
workbook.SaveAs(@"Outputxlsx");
```

The excel file [Test.xlsx](https://github.com/ClosedXML/ClosedXML/files/11496672/Test.xlsx) is completly empty except a single custom table style. Maybe this helps whenever somebody finds some time to look into this issue.

pavel-belenko:
[...]

PR #2701, opened 2025-05-19
Add StrictAttributeParsing flag to LoadOptions 

### LINKED ISSUE #2694: Invalid HorizontalAlignment Parsing Throws Exception

jahav:
Expose a mitigation option against malformed attribute values at XML parser level. The added `LoadOptions.StrictAttributeParsing` option will tell the XML parser that it should treat attributes with values outside of their simple type permitted values as missing.

This should help in these specific cases:
* attribute is optional (malformed attribute will be considered not present)
* attribute is optional/required, but has a default value (malformed attribute will have default value) - e.g. border line style `triangle` is malformed and thus will be treated as missing. The attribute has default value `none` and thus the final border line style will be `none`.

That was relatively common occurrence (well, more common than markup compatibility and extensions) when I was designing designed `XmlTreeParser`. It's a XML parser level mitigation. It checks against xsd simple type, and the schema simple types have rather limited subset of constraints (e.g. font size has a simple type `xsd:double`, but things like negative font size don't make sense plus Excel requires size to be 1-409.55 points). Still, it helped a lot when I tested demo parser against common crawl dump (that that was obviously syntax only check).

This option is obviously set to strict parsing by default. Fix the damn OOXML producers instead. Hint: `XmlReaderSettings.Schemas` + `XmlReaderSettings.ValidationType` in every test.

Resolves #2694


egiorgioPatagonian:
Hi @jahav 
Is this fix could solve the problem of [https://github.com/ClosedXML/ClosedXML/issues/1755](https://github.com/ClosedXML/ClosedXML/issues/1755)?

PR #2709, opened 2025-05-27
Search for worksheets in without unescaping

### LINKED ISSUE #2704: System.ArgumentException: There isn't a worksheet named 'XXX'XXX'
There is a bug when I call:
```
            using var wb = new XLWorkbook(ms);
```

The name of the sheet is` XXX''XXX`, but there's this error on load:

System.ArgumentException: There isn't a worksheet named 'XXX'XXX'.
   at ClosedXML.Excel.XLWorksheets.Worksheet(String sheetName)
   at ClosedXML.Excel.XLWorkbook.Load(Stream stream)
   at ClosedXML.Excel.XLWorkbook..ctor(Stream stream, LoadOptions loadOptions)

here is an added minimal example to reproduce:
[error.xlsx](https://github.com/user-attachments/files/20401494/error.xlsx)

```
using var stream = File.OpenRead("error.xlsx");
using var wb = new XLWorkbook(stream);
```

I investigated a little, and this is caused because this method in `XLWorksheets.cs`
```
        public IXLWorksheet Worksheet(String sheetName)
        {
            sheetName = sheetName.UnescapeSheetName();

            if (_worksheets.TryGetValue(sheetName, out XLWorksheet w))
                return w;

            throw new ArgumentException("There isn't a worksheet named '" + sheetName + "'.");
        }
```

More specifically, becuase
```
        internal static string UnescapeSheetName(this String sheetName)
        {
            return sheetName
                .Trim('\'')
                .Replace("''", "'"); // <------
        }

```

I have no idea why it does this, but this seems like a very simple fix.. just remove this line:
```
            sheetName = sheetName.UnescapeSheetName();
```

I am not sure what other things it affects, so tell me what you think... 

EDIT:

Another option is to add the same line to:
```
        private void Add(String sheetName, XLWorksheet sheet)
        {
            // Add it here:
           // sheetName = sheetName.UnescapeSheetName();

            if (_worksheets.ContainsKey(sheetName))
                throw new ArgumentException(String.Format("A worksheet with the same name ({0}) has already been added.", sheetName), nameof(sheetName));

            _worksheets.Add(sheetName, sheet);

            _workbook.NotifyWorksheetAdded(sheet);
        }
```

jahav:
When searching for a sheet in a `IXLWorksheets`, accept the argument as-is, without any unescaping. The original logic not only cause problematic behavior when using the API, it also crashed during workbook load.

Resolves #2704

### DIFF
--- a/docs/migrations/migrate-to-0.106.rst
+++ b/docs/migrations/migrate-to-0.106.rst
@@ -20,3 +20,14 @@ The number format setters (```IXLNumberFormatBase.NumberFormatId```,
 ```IXLNumberFormat.SetNumberFormatId(int)``` and ```IXLPivotValueFormat.SetNumberFormatId(int)```)
 now throw an ```ArgumentOutOfRangeException``` when supplied number format id is
 not a predefined format id from ```XLPredefinedFormat```.
+
+*************
+IXLWorksheets
+*************
+
+The method ``IXLWorksheets.Worksheet(string sheetName)`` now throws ``KeyNotFoundException`` when
+sheet is not found, instead of original ``ArgumentException``.
+
+Generally speaking, get-by-name methods of all unique-name collections (e.g., worksheets, styles,
+table fields ect) should throw ``KeyNotFoundException`` when item is not found. This change aligns
+the behavior of the ``IXLWorksheets`` with the rest of API.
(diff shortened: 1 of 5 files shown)


docs/migrations/migrate-to-0.106.rst
```diff
@@ -20,3 +20,14 @@ The number format setters (```IXLNumberFormatBase.NumberFormatId```,
...
+
+*************
+IXLWorksheets
+*************
+
+The method ``IXLWorksheets.Worksheet(string sheetName)`` now throws ``KeyNotFoundException`` when
+sheet is not found, instead of original ``ArgumentException``.
+
+Generally speaking, get-by-name methods of all unique-name collections (e.g., worksheets, styles,
+table fields ect) should throw ``KeyNotFoundException`` when item is not found. This change aligns
```
idan-h, line 32: Typo here

Issue #1671, opened 2021-05-26
Error Occurs with a photo whose name is 32 characters or more


HRK-OSN:
## Read and complete the full issue template

Do not randomly delete sections. They are here for a reason.

**Do you want to request a *feature* or report a *bug*?**
- [x] Bug
- [ ] Feature
- [ ] Question

**Did you test against the latest CI build?**
- [ ] Yes
- [x] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.95.4

**What is the current behavior?**

An error occurs when reading an Excel file with a photo whose name is 32 characters or more with ClosedXML.
There are no such restrictions on the ECMA or Excel specifications.

**What is the expected behavior or new feature?**

Please do no verify the attribute value.

**Is this a regression from the previous version?**

Sorry I don't know

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

Without a code sample, it is unlikely that your issue will get attention. Don't be lazy. Do the effort and assist the developers to reproduce your problem. Code samples should be [minimal complete and verifiable](https://stackoverflow.com/help/mcve). Sample spreadsheets should be attached whenever applicable. Remove sensitive information.

**Code to reproduce problem:**
```c#
public void Main()
{
   using (var inputStream = new FileStream("filePath", FileMode.Open)
   using (var workbook = new XLWorkbook(inputStream, XLEventTracking.Disabled)) // Error Occurs
   {
     // .... Some process
   }
}
```

- [x] I attached a sample spreadsheet.  (You can drag files on to this issue)
[Test.xlsx](https://github.com/ClosedXML/ClosedXML/files/6547005/Test.xlsx)

Pankraty:
We are validating picture name with these rules:

1. Picture names cannot be empty
2. Picture names cannot contain any of the following characters: :\/?*[]
3. Picture names cannot be more than 31 characters

They all do not seem to be valid, at least for opening a file.
1. Excel opens such files with no issues and assigns a default name (e.g. Picture 1). Libre Office is fine with having it empty.
2. Neither Excel nor LibreOffice object against naming a picture `:\/?*[]` via UI, and they load the file with such picture with no issues too.
[...]

Issue #1564, opened 2020-11-13
Picture names cannot be more than 31 characters


JuliaTitova77:
## Read and complete the full issue template
I have WPF app Core.
When try to load excel file like workbook = XLWorkBook(filename) appears error = Picture names cannot be more than 31 characters, but filename.Length <=31. What's happend?
System.ArgumentException
  HResult=0x80070057
  Сообщение = Picture names cannot be more than 31 characters
  Источник = ClosedXML
  Трассировка стека:
   в ClosedXML.Excel.Drawings.XLPicture.SetName(String value)
   в ClosedXML.Excel.XLWorkbook.LoadDrawings(WorksheetPart wsPart, IXLWorksheet ws)
   в ClosedXML.Excel.XLWorkbook.LoadSpreadsheetDocument(SpreadsheetDocument dSpreadsheet)
   в ClosedXML.Excel.XLWorkbook.LoadSheets(String fileName)
   в ClosedXML.Excel.XLWorkbook..ctor(String file)
  
   var wb = new XLWorkbook(inFilePath);<--- here get exception
Do not randomly delete sections. They are here for a reason.

**Do you want to request a *feature* or report a *bug*?**
- [] Bug
- [ ] Feature
- [x] Question

**Did you test against the latest CI build?**
- [ ] Yes
- [x ] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

e.g. 0.95.3

**What is the current behavior?**

Complete this.

**What is the expected behavior or new feature?**

Complete this.

**Is this a regression from the previous version?**

Regressions get higher priority. Test against the latest build of the previous minor version. For example, if you experience a problem on v0.95.3, check whether it the problem occurred in v0.94.2 too. 
With v0.94.2 the same error. 

## Reproducibility
**This is an important section. Read it carefully. Failure to do so will cause a 'RTFM' comment.**

Without a code sample, it is unlikely that your issue will get attention. Don't be lazy. Do the effort and assist the developers to reproduce your problem. Code samples should be [minimal complete and verifiable](https://stackoverflow.com/help/mcve). Sample spreadsheets should be attached whenever applicable. Remove sensitive information.

**Code to reproduce problem:**
```c#
public void Main()
{
    string XLSXfile = "";
    if (_filename.Length >= 31)
            {                              
                string sourceDir = @"c:\test";

[...]

PR #2715, opened 2025-06-04
Remove restriction for loading a long named picture.

Li-Major:
Remove restriction for loading a sheet with any picture has a name longer than 31 characters.

This seems has been a problem for some years. #1564 #1671 
If this exception is be thrown for compatibility consideration, would it be better to validate when saving the file, not when loading one?


Haris-AK:
Is there any update as to when this could be merged upstream?

oreadvanthink:
Thanks for this PR.

A previous comment https://github.com/ClosedXML/ClosedXML/issues/1671#issuecomment-850457561 says that none of the 3 checks (empty name, name length > 31, and invalid characters) are necessary. In particular, the check for an empty name is a blocker for us, because the XlsGen library we use to generate XSLX files generates empty image names in some cases. XLSX files that contain images with empty names can be opened with no issue with XlsGen, EPPlus, and MS Excel, but fail to load with ClosedXML.

Please consider removing the two other checks, too.

jahav:
Superseded by #2836 with a test.

PR #2716, opened 2025-06-05
Fix loading Excel files with pictures having forbidden characters in names

laszlozold:
Allow loading Excel files that contain pictures with names having characters that ClosedXML normally forbids (like backslash, colon, etc). This commonly occurs when loading files created by other applications.

- Add optional validateName parameter to XLPicture.SetName method
- Skip validation when loading pictures from existing Excel files
- Maintain validation for user-created pictures through public API
- Add comprehensive unit tests for the new functionality

Fixes issue where files with picture names like "Picture@01\QPosted@" couldn't be loaded.

### DIFF
--- a/ClosedXML.Tests/Excel/ImageHandling/PictureTests.cs
+++ b/ClosedXML.Tests/Excel/ImageHandling/PictureTests.cs
@@ -480,5 +480,151 @@ namespace ClosedXML.Tests
                 () => new XLWorkbook(stream),
                 @"Other\Pictures\ImageShapeZOrder-Output.xlsx");
         }
+
+        [Test]
+        public void CannotCreatePictureWithForbiddenCharacters()
+        {
+            using (var wb = new XLWorkbook())
+            {
+                var ws = wb.AddWorksheet("Sheet1");
+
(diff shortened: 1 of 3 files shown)


laszlozold:
Hi @jahav, @igitur, @Pankraty,

I've created issue #2727 with a reproducible test case for this bug. This PR fixes 
a problem where valid Excel files (that open fine in Excel/LibreOffice) cannot be 
loaded by ClosedXML due to picture name validation.

The fix is minimal - it just removes the overly restrictive validation during loading 
while keeping it for creation (to maintain Excel compatibility).

Would appreciate a review when you have time. Happy to make any adjustments needed.

Pankraty:
@jahav , may I go ahead and merge this? It seems, this would also fix #2715

jahav:
@Pankraty Replaced by #2836. Generally speaking, if a workbook can be loaded, it should also be possible to create it through API (always valid model). It makes everything simpler and more consistent. The PR allowed loading unusual names, but the name couldn't be set through API. Also, `CanLoadPictureWithForbiddenCharactersFromFile` didn't load anything from a file.

PR #2802, opened 2025-11-23
Added .NET8.0 as explicit target and Update SixLabors.Fonts

stefannikolei:
Added .net8 as an explicit target and updated SixLabors.Fonts for .net8

With this we allow Applications on newer TargetFrameworks to consume the latest SixLabors.ImageSharp versions. 
The library still targets .netstandard2.0 and .netstandard2.1 for the rest.

### DIFF
--- a/ClosedXML/ClosedXML.csproj
+++ b/ClosedXML/ClosedXML.csproj
@@ -2,12 +2,12 @@
 
   <PropertyGroup>
     <LangVersion>11.0</LangVersion>
-    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1;net8.0</TargetFrameworks>
     <!-- netstandard 2.0 doesn't contain nullable annotations for BCL, so we want to export
          annotations for our consumers, but no warnings for our code. Warnings will be
          emitted by netstandard2.1 target -->
     <Nullable Condition="'$(TargetFramework)' == 'netstandard2.0'">annotations</Nullable>
-    <Nullable Condition="'$(TargetFramework)' == 'netstandard2.1'">enable</Nullable>
+    <Nullable Condition="'$(TargetFramework)' != 'netstandard2.0'">enable</Nullable>
   </PropertyGroup>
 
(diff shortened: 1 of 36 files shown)


stefannikolei:
Adding .NET8.0 as a Target Framework Manager (TFM) is one way to address this issue with pinning of SixLabors.Fonts to v1.0.0.

Another option is to revert to System.Drawing. There’s a package called Managed.System.Drawing that’s intended to be a drop-in replacement. However, it doesn’t support .NET Standard; it targets .NET 4.8 and .NET 8.0. Therefore, you’d need to drop .NET Standard and target the frameworks directly.

In my opinion, the simplest solution is to explicitly add .NET 8.0 as a TFM.

I believe it’s time for this change because the usage of SixLabors.Fonts in version 1.0 prevents users of newer frameworks from upgrading to the latest version of the SixLabors Libraries, particularly ImageSharp.

jahav:
Closing. 
1. `SixLabors.ImageSharp` doesn't depend on `SixLabors.Fonts` and thus is doesn't constrain its installation
2. `SixLabors.Font` version 2 has different license from 1 and thus `ClosedXML` for net8 would have different license requirements than one for net framework. The correct version should by default MIT one, i.e. `SixLabors.Fonts` 1.0
3. `ClosedXML` declares `SixLabors.Fonts` dependency as version 1+. That includes version 2. Nuget by default resolves version 1, but nothing is stopping an application from just adding `<PackageReference Include="SixLabors.Fonts" Version="2.1.3" />` to their project and override default nuget resolution logic. There was no API change between 1 and 2.
4. When application uses a package that requires higher version (e.g. `SixLabors.ImageSharp.Drawing`), all this resolution happens automatically

`Managed.System.Drawing` is only wrapper over `SixLabors.Fonts`. 


That was my mistake,  Wisej employed SixLabors to create the library. <del>Also, `Managed.System.Drawing` is irrelevant and likely infringes SixLabors copyright. It doesn't even declare license, doesn't have a source repo and official nuget [description ](https://www.nuget.org/packages/Managed.System.Drawing) is _Fully managed replacement for System.Drawing.Common built using SixLabor's ImageSharp._</del>

PR #2830, opened 2026-02-15
Styles infrastructure switch

jahav:
This is a PR to change style infrastructure from old code (`*Key`, `*Value`, `*Repository`) to `*FormatValues stored in a `XLWorkbookStyles`. It will retain external interfaces (`IXLStyle`, `IXLFont` and the rest). It won't compile nor pass any tests for a while. I expect this will be a 5k+ diff in the end.

There are several goals, the main one is to allow pass-through load and save. That should solve a lot of issues related to styles. The internal structure should be far more aligned (though not identical) with the actual structure of styles part. Of course it should be faster and few other things.

This is the key thing for release of 0.106.


jahav:
Yay, finally switched style infrastructure. Not the best code I have ever written, but after all the time IDGAF. When I look at it in a hindsight, it is sooo not worth it. No release for a year, bug fixing non-existent in that time, looking at slight changes to styles part in every test file, dealting with my favorite `XLRichText` and its inbred cousin `XLComment` along with phonetics, and still 30 `TODO Styles`...  Hopefully it will do what I wanted:
* massively faster for large sheets/fills
* full representation (e.g. differential formats or table styles)
* provides ability to lot of bugs because structure is just better
* removed most static repositories

igitur:
Good work @jahav

jahav:
@igitur Thanks.

Issue #2168, opened 2023-09-18
SixLabors.Fonts v2.0.0


stefannikolei:
## Read and complete the full issue template

Do not randomly delete sections. They are here for a reason.

**Do you want to request a *feature* or report a *bug*?**
- [ ] Bug
- [ ] Feature
- [x] Question

**Did you test against the latest CI build?**
- [ ] Yes
- [x] No

If you answered `No`, please test with [the latest development build](https://ci.appveyor.com/project/ClosedXML/ClosedXML/branch/develop/artifacts) first.

**Version of ClosedXML**

0.102.1

**What is the current behavior?**

Currently ClosedXml targets SixLabors.Fonts v1.0.0 and is build against .NETStandard 2.0 and 2.1

**What is the expected behavior or new feature?**

What is the way forward?  Sixlabors.Fonts 2.0.0 targets .NET6 only. So ClosedXml can not use this version. This will prevent users to update to any Sixlabors package to the newest version (ImageSharp v3, ImageSharp.Drawing v2, Fonts v2)

Are there any plans to just target .NET6 or soon .NET8?

jahav:
I want to keep targeting netstandard2.0. Don't know yet what I am gonna do (possibly depend on SkiaSharp).

 I knew fonts were eventually follow ImageSharp with split license and net6, but I thought I would have more time. It was nearly a year from beta19 to 1.0.

hagaygo:
> I want to keep targeting netstandard2.0. Don't know yet what I am gonna do (possibly depend on SkiaSharp).
> 
> I knew fonts were eventually follow ImageSharp with split license and net6, but I thought I would have more time. It was nearly a year from beta19 to 1.0.

@jahav 

Hi

We have a project that uses .net 4.8 and we can not update the dependency of Sixlabors.Fonts to 2.X because it does not support it.
For now we keep on keeping Sixlabors.Fonts on version 1.

Is there any chance the dependcy could change to a package that keeps targeting .net standard ?

Thanks.

PR #2870, opened 2026-07-14
fix: bound SixLabors.Fonts to version range [1.0.0,3.0.0)

### LINKED ISSUE #2844: TypeLoadException when loading workbook with SixLabors.Fonts 3.0.0 due to GlyphMetrics breaking change
- [x] Bug
- [ ] Feature
- [ ] Question

- [x] Yes
- [ ] No

**Version of ClosedXML**

0.105.0

Opening a valid `.xlsx` file throws a `System.TypeLoadException` when `DefaultGraphicEngine` accesses `SixLabors.Fonts.GlyphMetrics` because `SixLabors.Fonts.GlyphMetrics` is binary-incompatible between the version `ClosedXML` was built against and `SixLabors.Fonts 3.0.0`.

Exception:

```text
System.TypeLoadException : Could not load type 'SixLabors.Fonts.GlyphMetrics' from assembly 'ClosedXML, Version=0.105.0.0, Culture=neutral, PublicKeyToken=fd1eb21b62ae805b' due to value type mismatch.
```

```text
Stack Trace: 
DefaultGraphicEngine.CalculateMaxDigitWidth(MetricId metricId)
DefaultGraphicEngine.CalculateMaxDigitWidth(MetricId metricId)
ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
DefaultGraphicEngine.GetMaxDigitWidth(IXLFontBase fontBase, Double dpiX)
XLWorkbook.CalculateColumnWidth(Double charWidth, IXLFont font, XLWorkbook workbook)
XLWorkbook.LoadSpreadsheetDocument(SpreadsheetDocument dSpreadsheet)
XLWorkbook.LoadSheets(Stream stream)
XLWorkbook.Load(Stream stream)
XLWorkbook.ctor(Stream stream, LoadOptions loadOptions)
XLWorkbook.ctor(Stream stream)
```

This appears to be caused by a transitive dependency/version conflict involving `SixLabors.Fonts`.

`ClosedXML 0.105.0` depends on `SixLabors.Fonts 1.0.0`, where `SixLabors.Fonts.GlyphMetrics` is an abstract class. In our solution, Central Package Management forces `SixLabors.Fonts to 3.0.0`. In `SixLabors.Fonts 3.0.0`, `GlyphMetrics` was changed to a readonly struct, which causes the runtime type mismatch.

Relevant upstream breaking change:
https://github.com/SixLabors/Fonts/compare/v2.1.3...v3.0.0#diff-8bbd68bd7967f4b40c17ea6f4de2c092859c9cd2162625eee2b5662037ad9669

Opening a valid `.xlsx` file should succeed without throwing an exception.

`ClosedXML 0.105.0` currently declares `SixLabors.Fonts (>= 1.0.0)` without an upper bound. Because `SixLabors.Fonts 3.0.0` contains a binary-incompatible change to `GlyphMetrics`, it may be safer for `ClosedXML `to declare a bounded dependency range, for example `< 3.0.0`, unless v3 is officially supported.

Unknown. I did not test against previous versions of `ClosedXML`.
This is reproducible with:

`ClosedXML 0.105.0`
`SixLabors.Fonts 3.0.0`, forced through Central Package Management

The root cause appears to be a binary-incompatible change in `SixLabors.Fonts`: `GlyphMetrics `changed from an abstract class in earlier versions to a readonly struct in v3.0.0, resulting in a `value type mismatch` at runtime.

## Reproducibility

```c#
using ClosedXML.Excel;

public void Main()
{
    // ClosedXML 0.105.0
    // SixLabors.Fonts 3.0.0 is forced via Central Package Management override.
    // Opening any .xlsx file, even an empty one, triggers the exception.

    using var inputStream = File.OpenRead("Empty_file.xlsx");
    var workbook = new XLWorkbook(inputStream); // throws System.TypeLoadException
}

```
Example Central Package Management override:
```xml
<ItemGroup>
  <PackageVersion Include="ClosedXML" Version="0.105.0" />
  <PackageVersion Include="SixLabors.Fonts" Version="3.0.0" />
</ItemGroup>
```
[Empty_file.xlsx](https://github.com/user-attachments/files/27728746/Empty_file.xlsx)

igitur:
## Summary

Prevents NuGet from resolving SixLabors.Fonts ≥ 2.0.0 as a dependency of ClosedXML, which protects .NET Framework consumers from runtime crashes.

## Problem

SixLabors.Fonts 2.0.0+ introduced commercial license requirements, and 3.0.0 targets `net8.0` only — dropping `netstandard` entirely. Because ClosedXML specifies `Version="1.0.0"` (≥ 1.0.0, no ceiling), NuGet can resolve 3.0.0 on any project that updates its packages broadly, causing:

```
System.TypeLoadException: Could not load type 'SixLabors.Fonts.GlyphMetrics'
from assembly 'ClosedXML' due to value type mismatch.
```

This is tracked in:
- #2844 — TypeLoadException with SixLabors.Fonts 3.0.0
- #2168 — SixLabors.Fonts v2.0.0 incompatible with netstandard
- #2171 — SixLabors.Fonts v2.0.0 changed to Commercial License

## Fix

Changes `Version="1.0.0"` to `Version="[1.0.0,2.0.0)"` — matching the established pattern already used for `DocumentFormat.OpenXml` in the same file.

## Compatibility

| Target | Before | After |
|---|---|---|
| .NET Framework 4.6.1+ | Could resolve 3.x → crash | Resolves ≤ 1.x → works |
| .NET Core / .NET 5+ | Could resolve 3.x → crash | Resolves ≤ 1.x → works |
| License enforcement | Hits v2/v3 license wall | Stays on Apache-licensed v1 |

## Notes

This is a minimal fix targeting the `release-candidate` branch (0.105.0). The maintainer has indicated (#2844 comment) that SixLabors.Fonts will eventually be replaced by SkiaSharp — this bound serves as a safety net until that migration is complete.

Closes #2844

### DIFF
--- a/ClosedXML/ClosedXML.csproj
+++ b/ClosedXML/ClosedXML.csproj
@@ -45,7 +45,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
     <PackageReference Include="RBush.Signed" Version="4.0.0" />
-    <PackageReference Include="SixLabors.Fonts" Version="1.0.0" />
+    <PackageReference Include="SixLabors.Fonts" Version="[1.0.0,2.0.0)" />
     <PackageReference Include="ClosedXML.Parser" Version="2.0.0" />
   </ItemGroup>
 


jahav:
I changed it to `[1.0.0,3.0.0)` (2795a634887e72d9aaf52c2c91640dba4a497e62), because `[1.0.0,2.0.0)` would cause warnings in projects with SixLabors.Fonts 2.x. Installing 2.x (e.g. 2.1.3) should be supported scenario. There was no API change between SixLabors.Fonts 1.0.0 and 2.0.0, only license and some cleanup like file scoped namespaces.

Here are the scenarios that should work:
# Only ClosedXML package
```xml
<ItemGroup>
  <PackageReference Include="ClosedXML" Version="0.105.1" />
</ItemGroup>
```

The nuget will resolve SixLabors.Fonts 1.0.0 and it will use that as dependency. 1.0.0 was released under Apache license.

# SixLabors.Fonts 2.x
```xml
<ItemGroup>
  <PackageReference Include="ClosedXML" Version="0.105.1" />
  <PackageReference Include="SixLabors.Fonts" Version="2.1.3" />
</ItemGroup>
```

This can happen if user has license from SixLabors or if license department said it's ok due to transitive exception in the license. Developer is always responsible for satisfying license plus Visual Studio even shows extra dialogue that package contains a license that is not open source
image

# SixLabors.Fonts 3.x
```xml
<ItemGroup>
  <PackageReference Include="ClosedXML" Version="0.105.1" />
  <PackageReference Include="SixLabors.Fonts" Version="3.0.0" />
</ItemGroup>
```
User will see warning about versions and if there is no license, there is also info about missing license:
image

jahav:
I wanted to make 0.105.1 release, but I removed AppVeyor integration and replaced it with GH actions (AppVeyor didn't support VS 2026, dotnet10), except 0.105 still has uses AppVeyor, so 0.105.1 would use it too, if it worked. Will look into it later.

igitur:
Awesome, thanks!

PR #2882, opened 2026-09-04
Fix GetHashCode collisions in area/point hashing

### LINKED ISSUE #2881: Poor GetHashCode() implementations cause severe performance degradation in formula dependency tracking

- [x] Bug
- [ ] Feature
- [ ] Question

- [x] Yes
- [ ] No

*(Tested against `develop` at commit `6762b849`, built and run from source rather than the NuGet package.)*

**Version of ClosedXML**

Built from source, `develop` branch, commit `6762b849` (2026-07-26)

`GetHashCode()` on two internal types used by the formula dependency tree produces badly clustered hash values, degrading dictionary lookups from close to O(1) to close to O(n) for common cases.

1. Area hashing (`FirstPoint`/`LastPoint`-based area type):

```csharp
public override int GetHashCode()
{
    return FirstPoint.GetHashCode() ^ LastPoint.GetHashCode();
}
```

For a single-cell area, `FirstPoint == LastPoint`, so `h ^ h == 0` for *every* single-cell area. Single-cell precedents (e.g. `=A1+B1`) are by far the most common shape of formula precedent, so nearly all of them collide into the same hash bucket.

2. Point hashing (`Row`/`Column`-based point/address type):

```csharp
public override int GetHashCode()
{
    return (Row * -1) ^ Column;
}
```

Negating `Row` only flips its sign — it doesn't scramble bits — so it still occupies a contiguous, narrow range of values. XOR-ing in `Column` only perturbs whichever low bits `Column` happens to set. The number of distinct hash values ends up capped at roughly `max(rows, columns)` rather than `rows * columns`, so any `Point`-keyed dictionary on a wide or tall sheet gets long collision chains.

Measured with 50,000 formulas, each with single-cell precedents, calling `Workbook.RecalculateAllFormulas()`:

| | Before | After |
|---|---|---|
| Recalculate | 30,397 ms | 1,167 ms |

That is roughly **26x faster**.

Allocations were unaffected — 307 MiB before and after, with near-identical GC collection counts (gen0 54-58, gen1 12-14, gen2 1-2) — confirming this is purely a hash collision issue, not an allocation issue.

Measured on an Intel i7-8665U (15 W, `powersave` governor) running Linux, .NET 8, Release build, one process per run. Absolute timings on this part vary by about 20% run to run once the package is thermally saturated, which is why medians and ranges are given; the ~26x ratio is the stable figure.

A secondary load/save benchmark (200,000 rows × 10 text + 5 number columns, 1,000 distinct texts) showed a smaller improvement, since that path isn't dominated by the dependency tree (single run each, so treat these as indicative only):

| | Before | After |
|---|---|---|
| Load | 16,400 ms | 14,866 ms |
| Save | 23,499 ms | 21,699 ms |

Not checked against a previous minor version specifically, but this looks like a long-standing issue rather than a recent regression — the flawed logic is basic XOR-hashing, not tied to any specific recent change.

## Reproducibility

```csharp
using System;
using System.Diagnostics;
using ClosedXML.Excel;

public class Program
{
    public static void Main()
    {
        const int rows = 50_000;

        var ws = wb.Worksheets.Add("Data");

        for (var r = 1; r <= rows; r++)
        {
            ws.Cell(r, 1).SetValue(r);
            ws.Cell(r, 2).SetValue(r * 2);
            ws.Cell(r, 3).FormulaA1 = $"=A{r}+B{r}";
        }

        var sw = Stopwatch.StartNew();
        wb.RecalculateAllFormulas();
        sw.Stop();

        Console.WriteLine($"Recalculate: {sw.ElapsedMilliseconds:N0} ms");
        Console.WriteLine($"Allocated: {GC.GetTotalAllocatedBytes() / (1024.0 * 1024.0):N0} MiB");
        // ~30,400 ms before the GetHashCode fix, ~1,200 ms after, on my machine.
    }
}
```

AkshayDhola:
Area.GetHashCode() XORed FirstPoint/LastPoint, which is always 0 for single-cell areas (the most common precedent shape). Point.GetHashCode() used (Row * -1) ^ Column, spanning only ~max(rows, columns) values. Both caused severe collisions in the formula dependency tree's dictionaries. Replaced with HashCode.Combine(...).

Recalculate time for 50,000 single-cell-precedent formulas dropped from 30,397 ms to 1,167 ms (~26x faster), same allocations.

Fixes #2881

### DIFF
--- a/ClosedXML.Tests/Excel/Coordinates/AreaTests.cs
+++ b/ClosedXML.Tests/Excel/Coordinates/AreaTests.cs
@@ -352,4 +352,22 @@ public class AreaTests
         Assert.AreEqual(expectedRight, right?.ToString());
         Assert.AreEqual(expectedLeft, left?.ToString());
     }
+
+    [Test]
+    public void Single_cell_areas_hash_codes_have_few_collisions()
+    {
+        // Areas are used in many dictionaries, such as formula dependency tree. Make sure the hash
+        // function doesn't produce too many collision. The hash function originally always
+        // produced hash code 0 for all one-cell areas, which caused a terrible set/dictionary
+        // performance.
+        var hashes = new HashSet<int>();
+        for (var row = 1; row <= 100; ++row)
+        {
+            for (var column = 1; column <= 100; ++column)
+                hashes.Add(new Area(new Point(row, column)).GetHashCode());
+        }
+
+        // Some collisions are inevitable, but the vast majority of areas must be distinguishable.
+        Assert.That(hashes.Count, Is.GreaterThan(9900));
+    }
 }
--- a/ClosedXML.Tests/Excel/Coordinates/AreaTests.cs
+++ b/ClosedXML.Tests/Excel/Coordinates/AreaTests.cs
@@ -354,6 +354,7 @@ public class AreaTests
     }
 
     [Test]
+    [Issue("2881")]
     public void Single_cell_areas_hash_codes_have_few_collisions()
     {
         // Areas are used in many dictionaries, such as formula dependency tree. Make sure the hash
--- a/ClosedXML.Tests/Excel/Coordinates/PointTests.cs
+++ b/ClosedXML.Tests/Excel/Coordinates/PointTests.cs
@@ -1,6 +1,7 @@
-﻿using ClosedXML.Excel;
+using ClosedXML.Excel;
 using NUnit.Framework;
 using System;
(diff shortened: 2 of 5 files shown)


jahav:
Thanks. I modified some texts to describe the problem instead of a code that has been deleted.

AkshayDhola:
Sounds good! Thanks for the update.

